PR #4, opened 2018-10-15
- Improve the docstring for the prepare service.

rjschwei:


schaefi:
yeah much better :+1: thanks.

PR #6, opened 2018-11-12
Added zypper migration service

schaefi:
The service which actually runs the zypper migration plugin.
The service is called after the preparation step has completed
successfully

### DIFF
--- a/suse_migration_services/units/migrate.py
+++ b/suse_migration_services/units/migrate.py
@@ -38,9 +38,9 @@ def main():
         Command.run(
             [
                 'zypper', 'migration',
+                '--non-interactive',
                 '--gpg-auto-import-keys',
                 '--no-selfupdate',
-                '--migration', get_product_selection_index(),
                 '--auto-agree-with-linceses',
                 '--product', get_migration_product(),
                 '--root', root_path
@@ -52,21 +52,6 @@ def main():
                 issue
             )
         )
-
-
-def get_product_selection_index():
-    """
-    Returns migration index
-
-    The zypper migration module can provide several migration targets
-    which gets indexed by a number. In the scope of this service the
-    explicit selection of a product also fix the index number to '1'
-    However for the case the specified product is not listed as the
-    first entry in the list we allow to specify the product index
-    in the configuration file too
-    """
-    config = _read_migration_config()
-    return config['migration']['target'].get('product_index') or '1'
 
 
 def get_migration_product():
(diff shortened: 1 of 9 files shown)


schaefi:
@jesusbv I have a few open questions on the code

1. The option ```--migration 1``` what does that mean ? and can we make it more explicit ? I'd like to give '1' a better name if possible
2. I choose the one zypper call method we discussed but that means we need the migration ruby script with the patch in the image. Let me know if that applies before merge of this one

schaefi:
As a side note the code uses the following yaml format for the config

```yaml
migration:
  target:
    product: foo
```

jesusbv:
> @jesusbv I have a few open questions on the code
> 
>     1. The option `--migration 1` what does that mean ? and can we make it more explicit ? I'd like to give '1' a better name if possible
> 
>     2. I choose the one zypper call method we discussed but that means we need the migration ruby script with the patch in the image. Let me know if that applies before merge of this one
> 
>     3. The code read /etc/migration-config.yml which needs to be present in the image build as well. Let me know when this applies.
> 
> 
> Thanks

[--migration N](https://github.com/SUSE/zypper-migration/blob/master/zypper-migration.8#L29) means that the script should take the migration with that number automatically.
In the next example, there is one migration available and the script is waiting for the input.

example

With migration option, it is automatically selected.

suse_migration_services/units/migrate.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    root_path = Defaults.get_system_root_path()
+
+    try:
+        Command.run(
+            [
+                'zypper', 'migration',
+                '--gpg-auto-import-keys',
+                '--no-selfupdate',
+                '--migration', get_product_selection_index(),
```
rjschwei: I think we want to drop the use of the "--migration" option and add "--non-interactive".

Only when "--non-interactive" is set  for migration will the option be passed to zypper [1]. When "--non-interactive" is set and "--migration" is not set then the migration plugin handles the "number based selection" [2].

This approach will get us out of the business of picking a number. If the meaning of the numbered selection ever changes this implementation will not be broken as the internal implementation of what

`migration_num = options[:migration]
if options[:non_interactive] && migration_num == 0
  # select the first option
  migration_num = 1
end`

means would also have to change.

[1] https://github.com/SUSE/zypper-migration/blob/master/zypper-migration#L519
[2] https://github.com/SUSE/zypper-migration/blob/master/zypper-migration#L397

schaefi: I also think it would be good to change the zypper-migration code to reflect this. You might have noticed that @jesusbv created a pull request against the zypper-migration repo with some adaptions.

@jesusbv do you think you can extend your changes to let us call the migration independent of any selection index ?

jesusbv: I think code wise there would be no change but no include --migration option in the zypper migration call. I guess in this case, it is known that it is going to be the first option.

PR #8, opened 2018-11-14
Added umount service

schaefi:
The umount service cleans up the migration host such that
no active mount reference into the migrated systems exists
anymore and we are safe to reboot

### DIFF
--- a/suse_migration_services/units/umount_system.py
+++ b/suse_migration_services/units/umount_system.py
@@ -1,0 +1,58 @@
+# Copyright (c) 2018 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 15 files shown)


schaefi:
Please note:

The umount service also should depend on the bootloader config update service. But as this service does not yet exists the dependency change here should be done when the pull request for the bootloader update service come along

suse_migration_services/units/umount_system.py
```diff
@@ -0,0 +1,58 @@
...
+    migration host. If for whatever reason a filesystem is busy
+    and can't be umounted, this condition is not handled as an
+    error. The reason is that the cleanup should not prevent us
+    from continuing with the migration process. The risk on
+    reboot of the migration host with a potential active mount
+    is something we take into account intentionally
+    """
+    root_path = Defaults.get_system_root_path()
+
+    for shared_location in ['/etc/sysconfig/network', '/etc/zypp']:
```
rjschwei: Makes me wonder if we shouldn't put this information somewhere in a file, /etc/sysconfig/network gets mounted in setup_host_network.py and zypp is mounted in mount_system.py, so we are collecting distributed data here. I just think having multiple locations of this data is a risk.

schaefi: yes agreed, should be mentioned only once. Will adapt

schaefi: Ok that turned out into quite some refactoring but I think it's worth it, please have a look again

rjschwei:
LGTM, thanks

PR #9, opened 2018-11-14
Update grub

jesusbv:
Upgrade the grub config with the proper migrated version content


suse_migration_services/units/grub_setup.py
```diff
@@ -0,0 +1,63 @@
...
+    root_path = Defaults.get_system_root_path()
+    grub_path = Defaults.get_grub_path()
+
+    proc_mount_point = os.sep.join(
+        [root_path, 'proc']
+    )
+
+    dev_mount_point = os.sep.join(
+        [root_path, 'dev']
+    )
```
rjschwei: I am not certain if grub config needs /sys but it may be a good idea to bind mount it as well just in case.

suse_migration_services/defaults.py
```diff
@@ -30,3 +30,7 @@ def get_system_root_path(self):
     @classmethod
     def get_migration_config_file(self):
         return '/etc/migration-config.yml'
+
+    @classmethod
+    def get_grub_path(self):
```
schaefi: should imho be more specific, e.g get_grub_config_file

suse_migration_services/exceptions.py
```diff
@@ -86,3 +86,9 @@ class DistMigrationZypperException(DistMigrationException):
     """
     Exception raised if the zypper migration call has failed
     """
+
+
+class DistMigrationGrubException(DistMigrationException):
```
schaefi: It's all about the config, we are not installing grub. Thus I think we should be specific e.g
DistMigrationGrubConfigException

suse_migration_services/units/grub_setup.py
schaefi: we should only add what's really needed. Are you sure proc/ is needed ?

suse_migration_services/units/grub_setup.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+        Command.run(
+            ['mount', '--bind', '/proc', proc_mount_point]
+        )
+        Command.run(
+            ['chroot', root_path]
+        )
+        Command.run(
+            ['grub2-mkconfig', '-o', grub_path]
+        )
```
schaefi: This should be one call

```python
Command.run(
￼    ['chroot', root_path, 'grub2-mkconfig', '-o', grub_path]
)
```

systemd/suse-migration-grub-setup.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update grub to migrated version
+
```
schaefi: We are missing the dependencies here. Should add

```
After=suse-migration.service
Requires=suse-migration.service
```

schaefi: Also the comment I think could be more specific, saying that we recreate the grub configuration file

suse_migration_services/units/grub_setup.py
```diff
@@ -0,0 +1,63 @@
...
+    except Exception as issue:
+        raise DistMigrationGrubException(
+            'Update grub failed with {0}'.format(
+                issue
+            )
+        )
```
schaefi: The exception is missing a cleanup of the mounted paths. Please note in case of an exception the service pipeline ends. Thus any service has the responsibility to cleanup on exception

schaefi:
@jesusbv hope you don't mind, I have merged master into your branch

The reason is that the changes done in the #8 pull request impacts this one too

Any mount process a service calls needs to be stored in the newly introduced system mount info file.
As you also bind mount locations your code needs to add the following:

```
from suse_migration_services.fstab import Fstab

system_mount = Fstab()
system_mount.read(
    Defaults.get_system_mount_info_file()
)

# your mount call

system_mount.add_entry(
    ...mount_device and mount_point as used in your mount call
)

system_mount.export(
    Defaults.get_system_mount_info_file()
)
```

schaefi:
If you need an example take a look at e.g suse_migration_services/units/prepare.py

I'm sorry if this is causing any inconvenience on your side

schaefi:
In order to get your local branch updated you can do the following

```
branch --set-upstream-to origin/update_grub update_grub
git reset --hard b01cbb6fca834289059ac83e31874b97957712d2
git pull
```

suse_migration_services/units/grub_setup.py
rjschwei: That I can answer ;) yes it is the script looks in /proc/self/mountinfo

schaefi: ok, thanks

suse_migration_services/units/grub_setup.py
```diff
@@ -0,0 +1,89 @@
...
+            ]
+        )
+        system_mount.export(
+            Defaults.get_system_mount_info_file()
+        )
+    except Exception as issue:
+        for entry in system_mount.get_devices():
+            Command.run(
+                ['umount', entry.mountpoint], raise_on_exception=False
+            )
```
schaefi: This is not going to work well because the order matters. The mount stack needs to be resolved in reverse order to avoid any busy state. E.g if there is /boot and /boot/efi you need to resolve reversed or both umount calls will fail.
Thus:

```python
for mount in reversed(system_mount.get_devices()):
```

schaefi:
ok this looks good now.

@jesusbv could you now rebase the branch and squash the commits in one or a useful set including a useful comment. Once that is done feel free to merge :smiley:

suse_migration_services/units/grub_setup.py
jesusbv: Yes, agreed. It should be solved now

PR #10, opened 2018-11-21
Loopback boot activation

schaefi:
__Added grub config extension to activate migration__
    
Provide /etc/grub.d/99_migration plugin which causes the creation of a Migration grub menu entry. In addition the package build was cleaned up and extended by a new sub-package suse-migration-activation which provides  that grub config plugin.

During testing of the loopback mount I found a "busy lock" problem which only exists on that boot scenario. As consequence this pull request also contains a commit to fix this part

__Notes:__
On merge of this change follow up changes to the package build which contains the live ISO migration image are needed. Those changes will be:

* Requires: suse-migration-activation
* A %post script which calls ```grub2-mkconfig```

If this is done the package with the migration ISO image will activate itself in a grub menu on install of 
the package. In my testing I could not yet came across a solution to make this entry the default boot 
entry. So at the moment the customer would need to select it from the menu which if course is not 
acceptable if we do that in the cloud.

* Also found by testing is the fact that we need to uninstall suse-migration-activation and recall ```grub2- 
  mkconfig``` to rollback the grub menu changes done when the migration system was activated.
  I suggest to do this as part of the grub update service in the suse-migration-services project

Thus there will be follow up changes/discussion to complete the processing here

Please take that into consideration on review

Thanks

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,6 +3,7 @@
 graft suse_migration_services/units
 graft package
 graft systemd
+graft grub.d
 
 include tox.ini
 include README.md
--- a/Makefile
+++ b/Makefile
@@ -13,14 +13,16 @@ build: check test
 	mv setup.pye setup.py
 	# provide rpm source tarball
 	mv dist/suse_migration_services-${version}.tar.gz \
-		dist/suse_migration_services.tar.gz
+		dist/suse-migration-services.tar.gz
 	# provide rpm changelog from git changelog
 	git log | helper/changelog_generator |\
-		helper/changelog_descending > dist/suse_migration_services.changes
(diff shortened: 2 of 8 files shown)


schaefi:
I have added the described changes related to this PR to the image.spec.in file of the Devel:PubCloud/SLES15-Migration-Package package.

schaefi:
@jesusbv I've opened https://github.com/SUSE/suse-migration-services/issues/12 for your reference

rjschwei:
LGTM, thanks for ading the explicit comments on the services, should avoid hassle when we go to submit this.

jesusbv:
LGTM, @schaefi thank you for opening #12

PR #11, opened 2018-11-21
Add kexec reboot service

jesusbv:
The kexec reboot service loads the new migrated kernel 
and reboot the system.

### DIFF
--- a/suse_migration_services/units/kernel_reboot.py
+++ b/suse_migration_services/units/kernel_reboot.py
@@ -1,0 +1,82 @@
+# Copyright (c) 2018 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+
+# project
+import re
+import os
+from suse_migration_services.command import Command
+from suse_migration_services.defaults import Defaults
+
+from suse_migration_services.exceptions import (
+    DistMigrationKernelRebootException
+)
+
+
+def main():
+    """
+    DistMigration reboot with  new kernel
+
+    Reads the info of the new kernel and loads the latest (new migrated kernel)
+    version for rebooting.
+    """
+    target_kernel = Defaults.get_target_kernel()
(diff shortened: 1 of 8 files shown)


suse_migration_services/units/kernel_reboot.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+
+def _get_cmdline(vmlinuz):
+    grub_config_file = Defaults.get_grub_config_file()
+    pattern = '"linux\t\\K.+"'
+    try:
+        cmd_line = Command.run(
+            ['grep', '-oP', '-m', 1, pattern, grub_config_file]
+        )
```
rjschwei: IMHO we should use Python to get the information from the grub config file rather than using grep.

schaefi:
@jesusbv my pull request changed the spec file and caused a conflict with your request. I fixed that and pushed the merged sources

suse_migration_services/units/kernel_reboot.py
```diff
@@ -0,0 +1,82 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+
+# project
+import re
+import os
```
schaefi: re and os are non project imports. This is just nitpicking but they should not be below the #project comment ;)

jesusbv: good catch. I will change it

suse_migration_services/units/kernel_reboot.py
```diff
@@ -0,0 +1,82 @@
...
+            ['kexec', '--unload']
+        )
+        raise DistMigrationKernelRebootException(
+            'Kernel rebooting failed with {0}'.format(
+                issue
+            )
+        )
+
+
+def _get_cmdline(vmlinuz):
```
schaefi: I think it would be more generic if the parameter is named ```kernel_name``` instead of vmlinuz
If at any point this code should run on another architecture the name will be different and the target_kernel method will become more generic but if the code will still match a variable named vmlinuz it gets confusing

suse_migration_services/units/kernel_reboot.py
```diff
@@ -0,0 +1,82 @@
...
+    grub_config_file_path = Defaults.get_grub_config_file()
+    if not os.path.exists(grub_config_file_path):
+        raise DistMigrationKernelRebootException(
+            'Could not find {0} to load the kernel options'.format(
+                grub_config_file_path
+            )
+        )
+    pattern = r'(?<=linux\t).*'
```
schaefi: just asking but does ```\t``` identify a tab ? if so I would not rely on that format, it should at least match on a char class with tabs and spaces

schaefi: grub config does not enforce tabs like e.g make

jesusbv: Yes, it does identify tab.

jesusbv: I will refactor the method with the parameter to make it more generic.

schaefi:
Looks all good now :+1:

Issue #12, opened 2018-11-22
grub_setup service should uninstall suse-migration-activation


schaefi:
The package suse-migration-activation is pulled in if the customer installs the SLES15-Migration rpm providing the live migration image. That package installs the grub plugin to activate the migration and should be deleted once the migration is done.

Thus the grub_setup service should uninstall:

```
rpm -e suse-migration-activation
```

__before__ the grub2-mkconfig call such that this call removes the migration entry from the grub menu again

schaefi:
All done, thanks

PR #15, opened 2018-11-25
Uninstall suse-migration-activation

jesusbv:
This is done when the migration is completed. 
The new grub menu should not have this entry.
This solves issue #12


schaefi:
yep that's all we need to make the migration menu entry to disappear after migration. There are also no dependencies set on the activation plugin, so a simple ```rpm -e``` seems good to me

PR #16, opened 2018-11-25
Add kexec-tools

jesusbv:
This is nedded for rebooting the migrated system
with the new kernel.


zhang532682642:
In the system call of kexec, after the MMU is closed, the code will automatically restart after a period of execution.

PR #19, opened 2018-11-28
Fixed kernel reboot service

schaefi:
The unit requirements and dependencies were set wrong


schaefi:
Currently testing the migration and saw that the reboot unit is called even before the migration started :)

PR #22, opened 2018-11-28
Fixed removal of suse-migration-activation

schaefi:
A simple rpm -e call leads to an error because the package
depends on the master package providing the image. Thus it's
better to instruct zypper to remove the package including its
dependencies


schaefi:
With this last change the migration went through

PR #28, opened 2018-12-05
Ssh access

jesusbv:
Copy the authorized key files into the authorized key file
of migration user to be able to access through ssh

### DIFF
--- a/suse_migration_services/defaults.py
+++ b/suse_migration_services/defaults.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
+import os
 
 
 class Defaults(object):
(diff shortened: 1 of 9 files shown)


suse_migration_services/defaults.py
```diff
@@ -46,3 +46,11 @@ def get_target_kernel(self):
     @classmethod
     def get_target_initrd(self):
         return 'boot/initrd'
+
+    @classmethod
+    def get_ssh_keys_path(self):
+        return '/home/*/.ssh/authorized_keys'
```
rjschwei: This is not sufficient. While the images we publish by default do not have a configuration that would create "/root/.ssh/authorized_keys" we cannot guarantee that a sysadmin of an instance did not create this and set the VM up more like a "traditional" system. Most likely that user/sysadmin would also be the one migrating the system. Meaning we have to check/account for "/root/.ssh/authorized_keys"

suse_migration_services/units/ssh_keys.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def main():
+    """
+    DistMigration ssh access for migration user
+
+    Copy the authoritation key found to the migration user
+    directory in order to access through ssh
+    """
+    ssh_keys_glob_path = Defaults.get_ssh_keys_path()
```
rjschwei: See above, I think this needs to be a list

suse_migration_services/units/ssh_keys.py
```diff
@@ -0,0 +1,58 @@
...
+    migration_ssh_file = Defaults.get_migration_ssh_file()
+    try:
+        ssh_keys_paths = glob.glob(ssh_keys_glob_path)
```
rjschwei: and here we have to handle the list

suse_migration_services/defaults.py
jesusbv: Good point. That case should be solved now.

PR #33, opened 2019-01-09
Mount kernel fs

### LINKED ISSUE #32: Mount kernel file systems like proc inside /system-root
The migration process contains errors like

```
Executing %posttrans script 'coreutils-8.29-2.12.x86_64.rpm' [..
Output of coreutils-8.29-2.12.x86_64.rpm %posttrans script:
    Creating initrd: /boot/initrd-4.12.14-25.25-default

update-bootloader: 2019-01-07 11:55:04 <3> update-bootloader-5204 run_comman
d.294: '/usr/lib/bootloader/grub2/config' failed with exit code 1, output:
    <<<<<<<<<<<<<<<<
    + /usr/sbin/grub2-mkconfig -o /boot/grub2/grub.cfg
    awk: fatal: cannot open file `/proc/self/mountinfo' for reading (No such fil
e or directory)
    /usr/sbin/grub2-probe: error: cannot find a device for / (is /dev mounted?).
    >>>>>>>>>>>>>>>>
    Updating bootloader failed
```

This is because the prepare service did not mount proc and other filesystems below /system-root

jesusbv:
This Fixes #32

### DIFF
--- a/suse_migration_services/units/grub_setup.py
+++ b/suse_migration_services/units/grub_setup.py
@@ -19,7 +19,6 @@ import os
 
 # project
 from suse_migration_services.command import Command
-from suse_migration_services.fstab import Fstab
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import log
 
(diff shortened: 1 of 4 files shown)


schaefi:
It would be good to squash both commit into one and provide a more meaningful commit message. It should be clear "why" we are moving the mounts from the commit message

PR #38, opened 2019-01-14
Import certificates

### LINKED ISSUE #37: Migration in the cloud on custom smt infrastructure
In case a registration is made against a different SMT server it is required to fetch the server
certificate from the system to become migrated into the live migration system and update the certificate pool

schaefi:
Copy certificates from /usr/share/pki/trust/anchors of the
system to become migrated into the live migration system
and update the certificate pool. This Fixes #37

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -49,6 +49,9 @@ def main():
     hosts_setup = os.sep.join(
         [root_path, 'etc', 'hosts']
     )
+    trust_anchors = os.sep.join(
+        [root_path, 'usr', 'share', 'pki', 'trust', 'anchors']
+    )
     if os.path.exists(suse_connect_setup):
         shutil.copy(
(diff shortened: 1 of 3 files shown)


suse_migration_services/units/prepare.py
```diff
@@ -49,6 +49,9 @@ def main():
     hosts_setup = os.sep.join(
         [root_path, 'etc', 'hosts']
     )
+    trust_anchors = os.sep.join(
+        [root_path, 'usr', 'share', 'pki', 'trust', 'anchors']
```
jesusbv: The '/usr/.../anchors' could be the returning of a Default method but it's ok as it is.

rjschwei: Well, update-ca-certificates does not have any location option, i.e. it will only update the trust root if the certificate to be added is in "/usr/share/pki/trust/anchors/". Thus having this hard coded here is IMHO preferable of a "defaults" implementation as the latter would imply that it can be customized.

PR #39, opened 2019-01-15
Product setup

schaefi:
__Product Setup__

Long time ago zypper added a distro_target information into the repository file and matches those with the product information of the system. If the distro doesn't match  zypper refused to create the repository. In the migration process using SMT this caused the zypper migration plugin to fail because zypper never created repositories.

That's  because in a migration process the repository and the system to migrate never matches. Therefore we bind mount  the /etc/products.d information from the migration live system into the
/system-root of the system to become migrated.

At the end of a successful migration the new product information is copied to the migrated system.
This task is handled by the new product_setup service

### DIFF
--- a/suse_migration_services/units/product_setup.py
+++ b/suse_migration_services/units/product_setup.py
@@ -1,0 +1,56 @@
+# Copyright (c) 2018 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+import os
+
+# project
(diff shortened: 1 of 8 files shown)


schaefi:
@jesusbv ok here is my share of the implementation we discussed

In case the ```zypper migration``` fails and tries its rollback code we should have a patch in that plugin which umounts etc/products.d beforehand. I'll wait for that ruby patch before I merge this one such that we have all complete. Thanks

suse_migration_services/units/product_setup.py
```diff
@@ -0,0 +1,56 @@
...
+        Command.run(
+            ['umount', products_metadata]
+        )
+        log.info('Syncing product data to migrated system')
+        Command.run(
+            [
+                'rsync', '-zav', '--delete', '/etc/products.d/',
+                products_metadata + os.sep
+            ]
+        )
```
rjschwei: Please add a comment that this may only work as long as FATE#320882 is not implemented. Once we support multiple registration servers this may break depending on the implementation of migration.

Yes it makes no sense to try and upgrade multiple sources from different servers at the same time, but if that happens then this implementation would clobber important files for the next migration.

Consider the following use case once FATE#320882 is implemented:
An on demand instance is also registered against SCC for life patching

Then there would be a life-patching.prod file in the products.d directory. If the migration runs in phases per update server and this code runs first then the life-patching.prod file wold be removed and that upgrade against SCC would fail.

There are way around this, and I can help with the implementation if we want to do that now. If we want to wait to handle that use case later then we should at least leave ourselves a note about this problem.

schaefi: Thanks for the talk on mumble. I've added comment blocks and hope they make the point clear

schaefi:
@rjschwei ok tried to explain why we do things as we do it

PR #40, opened 2019-01-16
Revert/Refactor distro_target handling

schaefi:
zypper implements a handling for the distro_target attribute.
If present in the repository metadata, zypper compares the
value with the baseproduct <target> setup in /etc/products.d.
If they mismatch zypper refuses to create the repo. In the
process of a migration from distribution [A] to [B] this mismatch
always applies by design and prevents the zypper migration plugin
to work. In SCC or RMT the repositories doesn't contain the
distro_target attribute which is the reason why we don't see
this problem there. SMT however creates repos including
distro_target.

The refactored workaround solution is now to delete the target
specification in the baseproduct registration if present, because
the overlay mounting of etc/products.d did not work as it would
lead to a wrong upgrade path on response from SCC.

In addition a backup of the original products.d data is created
and used in the zypper migration plugin in case of an error

### DIFF
--- a/suse_migration_services/units/product_setup.py
+++ b/suse_migration_services/units/product_setup.py
@@ -16,6 +16,7 @@
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
 import os
+from xml.etree.ElementTree import ElementTree
 
 # project
 from suse_migration_services.command import Command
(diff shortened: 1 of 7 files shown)


suse_migration_services/units/product_setup.py
```diff
@@ -29,44 +30,57 @@
...
-            ['umount', products_metadata]
-        )
-        log.info('Syncing product data to migrated system')
-        Command.run(
-            [
-                'rsync', '-zav', '--delete', '/etc/products.d/',
-                products_metadata + os.sep
-            ]
+        baseproduct = os.sep.join(
+            [products_metadata, 'baseproduct']
```
rjschwei: We have SLES4SAP images where the baseproduct link is incorrect. INstaned of looking for "baseproduct' we should look for 'SLES' or 'SLES_SAP' files.

PR #44, opened 2019-01-17
Detection of baseproduct

### LINKED ISSUE #42: Stable detection of baseproduct
The product_setup service relies on a symlink ```/etc/products.d/baseproduct``` to exist and to be correct. However that is not always the case and a source for trouble. A better solution is to walk through all .prod files in /etc/products.d and find the product which does __NOT__ define a

```xml
<flavor>module</flavor>
```

jesusbv:
Find the .prod file inside /etc/products.d
which does not contain a flavor tag.

This Fixes #42


suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +58,61 @@ def main():
...
-            xml.parse(baseproduct)
-            register_sections = xml.findall('register')
-            for register in register_sections:
-                target_sections = register.findall('target')
-                for target in target_sections:
-                    register.remove(target)
-            xml.write(
-                baseproduct, encoding="UTF-8", xml_declaration=True
-            )
+        baseproduct_filename = get_baseproduct(products_metadata)
```
schaefi: I think this is a bit misleading. The return from get_baseproduct is not a filename it's a list of filenames. I would change this to return a static tuple like this:

```python
from collections import namedtuple

base_product_type = namedtuple(
    'base_product_type', ['count', 'prod_filename']
)

...

baseproduct = base_product_type(
    count=1, prod_filename="foo"
)

print(baseproduct.count)
print(baseproduct.prod_filename)
```

The advantage is this is more readable and the code checking for the exceptions is more clean.
In case you found multiple baseproducts count is set to a number > 1 and prod_filename is None.
In case you found no baseproduct count is set to 0 and prod_filename is None. In case one baseproduct 
is found count is set to 1 and prod_filename is the name.

You don't have to fiddle with a list here and it makes the code more explicit. Thoughts ?

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +58,61 @@ def main():
...
+        if len(baseproduct_filename) == 1:
+            baseproduct = baseproduct_filename.pop()
+        else:
+            if not baseproduct_filename:
+                log.error('There is no baseproduct')
+                raise DistMigrationProductSetupException(
+                    'There is no baseproduct'
+                )
+            else:
+                log.error('There are more than 1 product without flavor tag')
```
rjschwei: Please update the message, I suggest "Found multiple product definitions without <flavor> element'

Also assign the message to a variable so it doesn't need to be typed twice, once for the exception and once for the log

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +58,61 @@ def main():
...
+
+
+def get_baseproduct(products_metadata):
+    baseproduct_filename = []
+    xml = ElementTree()
+    for product_name in os.listdir(products_metadata):
+        product = os.sep.join(
+            [products_metadata, product_name]
+        )
+        xml.parse(product)
```
rjschwei: This does not account for a user having put crap into the products.d directory by accident. We should consider only .prod files, i.e. use glob.glob(products_metadata + '/*.prod') that gives us a higher chance of only getting the files we put there.

You also need to catch any exceptions the parser may throw.

suse_migration_services/units/product_setup.py
jesusbv: I like the approach. Although, prod_filename will be ignored or irrelevant if count is not 1 as that content is never checked in those cases.

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +60,74 @@ def main():
...
+        if baseproduct.count != 1:
+            if not baseproduct.count:
+                message = 'There is no baseproduct'
+                log.error(message)
+                raise DistMigrationProductSetupException(message)
+            else:
+                message = ('Found multiple product definitions'
+                           'without element <flavor>')
+                log.error(message)
+                raise DistMigrationProductSetupException(message)
```
schaefi: This could be shorten a bit

```python
if not baseproduct.count:
    message = 'a'
else:
    message = 'b'
log.error(message)
raise DistMigrationProductSetupException(message)
```

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +60,74 @@ def main():
...
+        return base_product_type(count, name)
     except Exception as issue:
-        message = 'Base Product update failed with {0}'.format(issue)
+        message = ('Getting base product for product '
+                   'update failed with {0}'.format(issue))
+        log.error(message)
+        raise DistMigrationProductSetupException(message)
+
+
+def update_baseproduct(baseproduct_prod_filename):
```
schaefi: This is a bit nitpick but when I wrote that code it was not part of a function and there was a long comment block at the beginning describing what we do here. You moved the code into a method which is good because more code has been added and that requires a bit more structure. But you named the method 'update_baseproduct' and I think that's misleading now, because what we actually do is we delete the registered target from the baseproduct. I think we should change the name to be more explicit e.g

```python
def delete_target_registration(products_metadata):
```

Thoughts ?

jesusbv: Agree, it reads better with that name

suse_migration_services/units/product_setup.py
```diff
@@ -66,53 +68,66 @@ def main():
...
+                flavor = register.findall('flavor')
+                if not flavor:
+                    count = count + 1
+                    name = prod_filename
 
-    return baseproduct_filename
+        return base_product_type(count, name)
+    except Exception as issue:
+        message = ('Getting base product for product '
+                   'update failed with {0}'.format(issue))
```
rjschwei: Tee error should just tell us that there is a file that is not correct. This in and off itself is not really a problem, we just move on.

message = 'Could not pars {0} as XML file'.format(prod_filename)

Also the try: execpt needs to be inside the for loop so we can follow a best effort approach, i.e. parse as many files as possible ignore any errors.

suse_migration_services/units/product_setup.py
```diff
@@ -66,53 +68,66 @@ def main():
...
+        log.error(message)
+        raise DistMigrationProductSetupException(message)
```
rjschwei: Log the error, do not raise, we want a "best effort" approach

s/raise DistMigrationProductSetupException(message)/pass/

suse_migration_services/units/product_setup.py
```diff
@@ -66,53 +68,66 @@ def main():
...
+            xml.parse(prod_filename)
+            register_sections = xml.findall('register')
+            for register in register_sections:
+                flavor = register.findall('flavor')
+                if not flavor:
+                    count = count + 1
+                    name = prod_filename
 
-    return baseproduct_filename
+        return base_product_type(count, name)
```
rjschwei: While we count the number of files that do not have the <flavor> element, this construct looses the information about which files have the missing element, i.e. you may end up with count having the value 3, and name has the value of the last file that did not have the <flavor> element. This is confusing. We either only count and then complain later if count > 1 as is currently implemented, or we count and keep the filename information, but then we need to keep track of all the filenames that have the problem. The latter approach makes for better error reporting and thus we should go with that.

suse_migration_services/units/product_setup.py
```diff
@@ -66,53 +68,66 @@ def main():
...
+            target_sections = register.findall('target')
+            for target in target_sections:
+                register.remove(target)
+        xml.write(
+            baseproduct_prod_filename, encoding="UTF-8", xml_declaration=True
+        )
+    except Exception as issue:
+        message = 'Base product update failed with {0}'.format(issue)
+        log.error(message)
+        raise DistMigrationProductSetupException(message)
```
rjschwei: See comments about "best effort"

suse_migration_services/units/product_setup.py
schaefi: I agree on that, we should have the proper filenames available with the counter

suse_migration_services/units/product_setup.py
schaefi: Regarding the objection on raise conditions I disagree with you here. I think they are properly set.
If any code in the try block failed this means there was an error on product data backup or on reading / handling product files. We should not continue the migration if there was an error. The best effort approach does not apply here because we just move the error to the subsequent zypper call.

The other raise condition is if the base product counter is != 1. In this case I also think we should not continue the migration. The result of zypper is basically undefined if there are more than one base products and we should raise before calling zypper.

All of that of course only if the way we detect the base product is correct. So far the lookup for the only one with no flavour set seems correct.

Thus I would not change the raise conditions. At this point we still can safely stop and haven't destroyed the system. The level of strictness is imho ok and preferable

suse_migration_services/units/product_setup.py
jesusbv: I agree that the more info we show, the better. I understood, in the previous comments, if count > 1 , prod_name would be None so I did not keep the names. I will adapt the code to reflect the changes and keeping the names.

suse_migration_services/units/product_setup.py
schaefi: sorry I think I misread it. I just saw there is a raise call in get_baseproduct(). I agree that method should not raise, just log and always return with a result. The result is evaluated and that could lead to raise

schaefi: Same for delete_target_registration. That method doesn't need to try / catch. It's result is covered by the overall try block where we call it.

schaefi: I think that was the part Robert was referring too and that would make sense to me too

schaefi: As delete_target_registration raises with the same exception than the overall try block it got called it doesn't make sense. So I suggest to not raise in delete_target_registration

less code is better

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +60,74 @@ def main():
...
-            )
+                flavor = register.findall('flavor')
+                if not flavor:
+                    count = count + 1
+                    names.append(prod_filename)
+
+        except Exception:
+            message = ('Could not parse {0} as XML file'.format(prod_filename))
+            log.error(message)
+            pass
```
schaefi: The pass statement in Python is used when a statement is required syntactically but you do not want any command or code to execute. The pass statement is a null operation; nothing happens when it executes. In the block you are using it there is code, so you don't need this statement here, it's superfluous. Same applies to the other code block where you used it.

In addition I don't see a reason for an extra message variable in this case. So the code can be shortened like this

```python
log.error(
    'Could not parse {0} as XML file'.format(prod_filename)
)
```

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +60,74 @@ def main():
...
+                register.remove(target)
+        xml.write(
+            baseproduct_prod_filename, encoding="UTF-8", xml_declaration=True
+        )
     except Exception as issue:
-        message = 'Base Product update failed with {0}'.format(issue)
+        message = 'Could not delete target registration with {0}'.format(issue)
         log.error(message)
-        raise DistMigrationProductSetupException(message)
+        pass
```
schaefi: see above, superfluous pass statement

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +60,74 @@ def main():
...
+                message = 'There is no baseproduct'
+            else:
+                message = ('Found multiple product definitions '
+                           'without element <flavor>: \n{0}'
+                           .format('\n'.join(baseproduct.prod_filenames)))
+            log.error(message)
+            raise DistMigrationProductSetupException(message)
+
+        log.info('Updating Base Product to be suitable for migration')
+        delete_target_registration(baseproduct.prod_filenames.pop())
```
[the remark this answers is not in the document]
schaefi: And sorry there is now another issue which came to my mind. Due to Roberts last objection the get_products method was changed to store all base product names as a list. Doing so invalidates the idea of the tuple type. In this case get_products can simply return the list of names. len(list) will be the same as count:

```python
def get_baseproduct(products_metadata):
    base_product_files = []
    prod_filenames = glob.glob(
        os.path.join(products_metadata, '*.prod')
    )
    xml = ElementTree()
    for prod_filename in prod_filenames:
        try:
            xml.parse(prod_filename)
            register_sections = xml.findall('register')
            for register in register_sections:
                flavor = register.findall('flavor')
                if not flavor:
                    base_product_files.append(prod_filename)
        except Exception as issue:
            log.error(
                XML Parsing {0} failed with: {1}'.format(prod_filename, issue)
            )
    return base_product_files
```

Later on this could be used like this

```python
baseproducts = get_baseproduct(products_metadata)
if baseproducts:
    count = len(baseproducts)
    ...
```

suse_migration_services/units/product_setup.py
jesusbv: I thought it was not needed (the pass statement) but I understood that from the comments Robert shared about it.

suse_migration_services/units/product_setup.py
```diff
@@ -58,29 +59,67 @@ def main():
...
+        try:
+            xml.parse(prod_filename)
+            register_sections = xml.findall('register')
+            for register in register_sections:
+                flavor = register.findall('flavor')
+                if not flavor:
+                    base_product_files.append(prod_filename)
+
+        except Exception:
+            log.error('Could not parse {0} as XML file'.format(prod_filename))
```
schaefi: minor issue, we should report the "why" for the exception. Could you change this to be:

```python
except Exception as issue:
    log.error(
        'Parsing XML file {0} failed with: {1}'.format(prod_filename, issue)
    )
```

schaefi:
@jesusbv please squash commits into one before merge. Thanks :+1:

PR #49, opened 2019-02-08
Rename migration log file

jesusbv:
As zypper is part of the process but not the only
component, 'zypper_migration.log' does not describe it properly.


schaefi:
@jesusbv also noted the update of the root.tar.gz tarball, looks good now too :+1:

PR #50, opened 2019-02-11
Report migration error in /etc/issue

### LINKED ISSUE #47: Report migration error in /etc/issue
In case of an error the migration system should write a file ```/system-root/etc/issue" which contains something like:

* Migration has failed, for further details see /var/log/zypper_migrate.log

This will give users a direct pointer to the state of the migration when they log in back to the system.
In case of success no further information is needed imho

jesusbv:
In order to be able to reboot and still report errors in case the migration fails,
the error will be reported in a log file (/var/log/distro_migration.log).
The message and log file name to check will be indicated in /etc/issue.

Fixes #47

### DIFF
--- a/suse_migration_services/units/migrate.py
+++ b/suse_migration_services/units/migrate.py
@@ -16,6 +16,7 @@
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
 import yaml
+import os
 
 # project
 from suse_migration_services.command import Command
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/migrate.py
```diff
@@ -54,6 +55,15 @@ def main():
             ['bash', '-c', bash_command]
         )
     except Exception as issue:
+        root_path = Defaults.get_system_root_path()
+        etc_issue_path = os.sep.join(
+            [root_path, 'etc/issue']
+        )
+        with open(etc_issue_path, 'a+') as issue_file:
```
schaefi: I think we should not append our information but overwrite with open mode ```w```. The reason is this. By default etc/issue has something like "Welcome ..." if we append an error message it sounds like "Welcome ... you are screwed" :-) So I think it's better we just write a clear message and not allow maybe misleading or funny combinations. Thoughts ?

schaefi:
Another details. This PR Fixes #47 so your commit message should include "... Fixes #47" This will then auto close the issue as soon as this gets merged

suse_migration_services/units/migrate.py
jesusbv: Yes, agree that would look weird as well but, I thought what if aother process is using /etc/issue for the same purpose, user could have more info besides "welcome" usual but, yes it should only report about the migration. I will change it.

schaefi: I'm not aware of another process setting /etc/issue. The file belongs to the XXX-release package and I think nothing else modifies it

schaefi:
Looks good now :+1: 

Please rebase into one commit before merge. Thanks

PR #53, opened 2019-02-20
Right path for debug file

jesusbv:


jesusbv:
Re open the conversation about the debug file.
I agree that the file should be searched on the [system we want to migrate](https://github.com/SUSE/suse-migration-services/pull/51#discussion_r256747445) but that has a problem:
- By the time the code reaches the point of checking that file, /system-root has been umounted.

So, there are workarounds for that:
 - Put the file on /tmp (not /system-root/tmp): This guarantees the file is there when reading happens.
 - Not umount /system-root **if migration failed**, read the file before reboot and then, umount /system-root. 

Thoughts ?

rjschwei:
The timing an handling is up to us. We can look for the file when system-root gets mounted and copy it to some location of our choosing inside the migration system.

Issue #41, opened 2019-01-16
Collection of side effects


schaefi:
__After the Migration__

* Configuration files marked as %config in spec files will have a .rpmnew version but won't be replaced. It's in the responsibility of the customer to adapt those. Example: custom /etc/motd file in pubcloud images

* Repositories without an upgrade path stay in the migrated system and are outdated. It's a user responsibility to update those to point to the new distribution if required at all. Example: static nvidia repos in pubcloud images

* Instance metadata in the cloud console will still point to the old system. If you run a sles12 instance in the cloud and migrate it, only the system itself knows it got migrated. But the cloud console will not be able to show you a change of the operating system. Example: See AMI ID in EC2 console of a sles12 on demand image after migration to sle15. There is nothing we can do against it

Please add your findings here

schaefi:
In reference to #55 we don't support the migration on systems that uses a root filesystem or setup which is not directly reachable from the grub bootloader. The loopback boot support used with the migration must be able to load the image from the root filesystem. If this is not possible for reasons shown in the following list, the migration system will not boot:

* unsupported fs from the grub space
* encrypted root filesystem

schaefi:
* No support for upgrading intermediate distribution states like Beta or RC.

schaefi:
* As stated previously I consider a 12 Sp1 -> SLE 15 upgrade out of scope. It also breaks our advertised "skip a service pack" capabilities, SP1 to 15 upgrade would skip 3 service packs.

PR #55, opened 2019-03-04
Boot space too small

### LINKED ISSUE #54: Migration of Systems with small boot partition
I was testing the migration process using a SLES12 SP4 instance in AWS. This image is designed by us using an XFS root partition and an extra boot partition of 300MB size. Our live migration image gets installed intentionally below /boot because of the reasons explained here #45 

Now the 300MB are not big enough to install the SLES15-Migration package:

```
zypper in SLES15-Migration
Refreshing service 'SMT-http_smt-ec2_susecloud_net'.
Refreshing service 'cloud_update'.
Loading repository data...
Reading installed packages...

The following NEW package is going to be installed:
  SLES15-Migration

The following package has no support information from it's vendor:
  SLES15-Migration

1 new package to install.
Overall download size: 0 B. Already cached: 189.1 MiB. After the operation,
additional 229.0 MiB will be used.
Continue? [y/n/...? shows all options] (y): y
In cache SLES15-Migration-1.15.0-26.x86_64.rpm
                                           (1/1), 189.1 MiB (229.0 MiB unpacked)
Checking for file conflicts: .............................................[done]
(1/1) Installing: SLES15-Migration-1.15.0-26.x86_64 .....................[error]
Installation of SLES15-Migration-1.15.0-26.x86_64 failed:
Error: Subprocess failed. Error: RPM failed: warning: /var/cache/zypp/packages/suse-migration/SLES15-Migration-1.15.0-26.x86_64.rpm: Header V3 DSA/SHA1 Signature, key ID 8a7c64f9: NOKEY
	installing package SLES15-Migration-1.15.0-26.x86_64 needs 15MB on the /boot filesystem
```

Older images are designed differently such that the size limit on /boot is of no relevance.
But now as SP4 is out the 300MB fixed boot partition is a big problem

schaefi:
This patch is two fold:

__Add rootpart detection to grub activation script__
    
The live migration image gets installed to the system again because of the space limitation on /boot. This affects the menuentry created on grub side in a way that we can't use the pre-allocated pointer to the boot device but have to search the root partition like in a real grub root entry. This patch adds the needed code changes to locate the root part, insert the needed filesystem module and initializes the root variable to allow the loopback loading of the image.

__Revert location change from /usr/share to /boot__
    
This reverts commit #66dd8d4943d38f121f4b16b70bf0ab8d0b2ec82d.  If there is an extra boot partition and it's too small we are not able to install the migration system and the customer will not be able to use the migration concept. The image is  usually around ~300MB of size and that can often be too much  for an extra boot partition which is often designed to be small

This Fixes #54

### DIFF
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -13,7 +13,7 @@ else
     ) ${CLASS}"
 fi
 
-migration_iso=$(echo /boot/*-Migration.*.iso)
+migration_iso=$(echo /usr/share/migration-image/*-Migration.*.iso)
 
 if grub_file_is_not_garbage "${migration_iso}"; then
     kernel="(loop)/boot/x86_64/loader/linux"
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -14,6 +14,9 @@ else
 fi
 
 migration_iso=$(echo /usr/share/migration-image/*-Migration.*.iso)
+root_device=$(lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | cut -f1 -d" ")
+root_uuid=$(blkid -s UUID -o value "${root_device}")
+root_type=$(blkid -s TYPE -o value "${root_device}")
 
 if grub_file_is_not_garbage "${migration_iso}"; then
     kernel="(loop)/boot/x86_64/loader/linux"
@@ -22,6 +25,8 @@ if grub_file_is_not_garbage "${migration_iso}"; then
     boot_options="rd.live.image root=live:CDLABEL=CDROM"
     printf "menuentry '%s' %s \${menuentry_id_option} '%s' {\n" \
         "${OS}" "${CLASS}" "Migration-${boot_device_id}"
+    printf "    insmod %s\n" "${root_type}"
+    printf "    search --no-floppy --fs-uuid --set=root %s\n" "${root_uuid}"
     printf "    set isofile='%s'\n" \
         "${migration_iso}"
     printf "    loopback loop (\$root)\$isofile\n"


schaefi:
This should work but hasn't been tested on all clouds. There is a case where it will not work and this is if the used filesystem is not supported by grub or cannot be read for other reasons e.g because root is encrypted. So we are limiting the concept here which should be clear to everybody.

However the limitation of the boot partition being too small will happen more often than the case that the rootfs can't be read by grub.

Thoughts

jesusbv:
> This should work but hasn't been tested on all clouds. There is a case where it will not work and this is if the used filesystem is not supported by grub or cannot be read for other reasons e.g because root is encrypted. So we are limiting the concept here which should be clear to everybody.

I agree with being clear about it, but do we want to support filesystems not grub supported?

> However the limitation of the boot partition being too small will happen more often than the case that the rootfs can't be read by grub.

It makes sense to me cover the more often scenario.

schaefi:
> I agree with being clear about it, but do we want to deal with filesystems that
> are not supported by grub ?

As of today there are none. But regarding xfs and also btrfs grub was late supporting them. So there might be the situation that a filesystem becomes popular and is supported on the kernel side but has no native support from the bootloader side. This scenario was often supported in the past by distributions. From a support point of view the answer is yes

I see the trouble more in the area of rootfs encryption. With the design before we would have kept the opportunity to upgrade such systems as well because /boot is never part of the encrypted area. There would still be the question how to ask for the password but at least a possible path to success exists. Once we merge this we are no longer able to support it. We have no customer for this scenario yet but at least Azure invests into encrypted instances

> It makes sense to me to cover the more common scenario.

Same here

schaefi:
I know I'm constructing corner cases but as we write documentation about the upgrade concept we should list what is possible and what not. The standard way "put the shiny CD in and run through the process" surely supports upgrade of encrypted systems. Our way would not support it and that is definitely worth a note in the documentation

schaefi:
If we agree to merge this I would add a short note in the #41 to remember

rjschwei:
FYI SLE no longer supports /boot on s separate FS. But we obviously still build images that way, so we need to support this.

PR #61, opened 2019-03-21
Initialize logging and debug as early as possible

schaefi:
The earliest opportunity to setup the log file and the
debug flag is directly after the mount-system code has
successfully mounted the target to migrate. At this point
we should init the logfile and also the debugging if
needed. The code before this commit has potential to
not even reach the log/debug setup. With this commit
the only requirement is a successful mount of the target
system.


rjschwei:
Great catch :)

PR #67, opened 2019-03-22
Copy kernel and initrd in-memory

### LINKED ISSUE #65: kexec based reboot broken
The current implementation runs kexec to load kernel and initrd from the root_path. The root_path however gets umounted before the kexec run is invoked as part of the reboot service. That reboot type will fail because the initrd file is no longer accessible. After that the force reboot kicks in.

We should fix this in a way that the kernel and the initrd needs to be copied from the target system after migration into the live migration memory. From there kexec should load the data.
After that umount is a safe operation

jesusbv:
It copies the files in-memory so kexec can access them
once root-path gets umounted.

Fixes #65

### DIFF
--- a/suse_migration_services/units/kernel_load.py
+++ b/suse_migration_services/units/kernel_load.py
@@ -17,11 +17,13 @@
 #
 import re
 import os
+import shutil
 
 # project
 from suse_migration_services.command import Command
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import log
+from suse_migration_services.path import Path
 
 from suse_migration_services.exceptions import (
     DistMigrationKernelRebootException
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/kernel_load.py
```diff
@@ -36,16 +37,23 @@ def main():
     """
     root_path = Defaults.get_system_root_path()
 
-    target_kernel = Defaults.get_target_kernel()
-    target_initrd = Defaults.get_target_initrd()
+    target_kernel = os.sep.join([root_path, Defaults.get_target_kernel()])
+    target_initrd = os.sep.join([root_path, Defaults.get_target_initrd()])
+    in_memory_path = os.sep
```
schaefi: I think this is hard to understand without knowing the context of the system underneath. The code doesn't know that it runs in a kiwi live system loaded into RAM. So I would name that differently and I think we should stick it to ```/var/tmp/kexec/``` , maybe something like:

```python
kexec_boot_data = '/var/tmp/kexec'
Path.create(kexec_boot)

....
```

test/unit/units/kernel_load_test.py
```diff
@@ -64,32 +67,40 @@ def test_main_raises_on_kernel_load(
...
     ):
+        mock_get_system_root_path.return_value = '../data'
         cmd_line = \
             'root=UUID=ec7aaf92-30ea-4c07-991a-4700177ce1b8' + \
             'splash root=UUID=ec7aaf92-30ea-4c07-991a-4700177ce1b8 rw'
         mock_get_cmdline.return_value = cmd_line
         main()
+        assert mock_shutil_copy.call_args_list == [
+            call('../data/boot/vmlinuz', '/'),
+            call('../data/boot/initrd', '/')
```
schaefi: why do we need test/data/boot/... files to really exist ? I've seen the two empty files coming along with the pull request. I think that's not needed because the kexec call itself is also mocked

schaefi:
Looks great now :+1:

PR #68, opened 2019-03-26
Fix pattern for initrd

jesusbv:
When getting initrd inside grub for cmdline,
the pattern should find '/initrd'.

### DIFF
--- a/suse_migration_services/units/kernel_load.py
+++ b/suse_migration_services/units/kernel_load.py
@@ -43,9 +43,6 @@ def main():
     kexec_boot_data = '/var/tmp/kexec'
     Path.create(kexec_boot_data)
     shutil.copy(
-        target_kernel, kexec_boot_data
-    )
-    shutil.copy(
         target_initrd, kexec_boot_data
     )
     try:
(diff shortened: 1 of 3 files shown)


schaefi:
Can we rename ```test/data/fake_grub_no_boot.cfg``` to something like ```fake_grub_with_bootpart.cfg```
I for sure will not understand what "no_boot" means ;) Thanks

suse_migration_services/units/kernel_load.py
```diff
@@ -87,9 +82,13 @@ def _get_cmdline(kernel_name):
...
             )
         )
-    pattern = r'(?<=linux)[ \t]+{0}{1}.*'.format(os.sep, kernel_name)
     with open(grub_config_file_path) as grub_config_file:
         grub_content = grub_config_file.read()
+    pattern = r'(?<=linux)[ \t]+{0}{1}.*'.format(os.sep, kernel_name)
+    if not re.findall(pattern, grub_content):
+        pattern = r'(?<=linux)[ \t]+{0}{1}.*'.format(
+            os.sep, Defaults.get_target_kernel()
+        )
```
schaefi: Huh this is hard to read. So it took me a while to understand how it works. So you called _get_cmdline() with the basename of target_kernel which is actually the basename of what Defaults.get_target_kernel() returns. So that is the basename of 'boot/vmlinz' which is 'vmlinuz', ok and if that doesn't match we recall the match with Defaults.get_target_kernel() which isi 'boot/vmlinuz' ok that works but this is pretty confusing

Wouldn't it also work with an adapted expression:

**untested**

```python
pattern = r'(?<=linux)[ \t]+{0}({1}.*|boot/{1})'.format(os.sep, os.path.basename(kernel_name))
```

jesusbv: Yes, argh, didnt think in changing the reg exp

PR #70, opened 2019-04-09
Support new cloud regionsrv client

### LINKED ISSUE #69: Support new cloud-regionsrv-client
New `cloud-regionsrv-client` service plugin outputs repositories with `plugin:/susecloud` URLs, which breaks migration, as the live migration system is missing URL resolver plugin.

In the future, the update server will also require instance metadata to allow access to the repositories, so the live system also needs to be able to collect it.

Potential solutions that we've briefly discussed with @schaefi and @jesusbv:

1. Copy/mount zypper plugins and metadata gathering utils from the host system
2. Bake `cloud-regionsrv-client-plugin-$csp` into the live system image (this, however, makes it CSP-specific).

schaefi:
The new cloud region service client comes with a plugin for
zypper that translates repo URIs into a specific plugin format
such that it's no longer possible to gain access to the SLES
repositories of our public cloud infrastructure by playing
tricks with snapshotting an on-demand instance without the
billingtag. However the new and fancy way makes migration of
such a system more difficult because the new repo URI format
must be known to zypper migration. This requires all software
components to be provided in the live migration image which
has been done in the image description but furthermore requires
that the cloud specific /etc/regionserverclnt.cfg and the
contents of /usr/lib/zypp/plugins from the system to migrate
are available in the migration live image. The later is
implemented in this commit and Fixes #69

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -45,6 +45,9 @@ def main():
     suse_connect_setup = os.sep.join(
         [root_path, 'etc', 'SUSEConnect']
     )
+    suse_cloud_regionsrv_setup = os.sep.join(
+        [root_path, 'etc', 'regionserverclnt.cfg']
+    )
     hosts_setup = os.sep.join(
         [root_path, 'etc', 'hosts']
     )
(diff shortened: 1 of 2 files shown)


schaefi:
@ikapelyukhin as Robert is still traveling would be great if you can have a look if this is correct from your point of view. Maybe reviewing by the test makes it pretty clear what happens. Thanks much

schaefi:
Per ssh debugging session one additional commit to make /var/lib/cloudregister available on the live migration system was still missing.

@ikapelyukhin could you have a look again. Should be good now I think

Thanks

rjschwei:
Yes of course the server object cache. LGTM

PR #71, opened 2019-04-11
Fix network setup service

schaefi:
The network setup service for the migration inherits the
network as it is configured on the system to become migrated.
As part of that process the network is restarted through
systemctl. However this is a problem because it interferes
with the network service dependency chain. This could result
in a failed network startup depending on how fast other
services e.g dbus are up and running. So the way we do it
causes a race condition. This commit deletes the network
startup from our network setup service. Our network setup
service runs by definition of the unit file before the
network.target. This means at the time the network service
activates the network our setup routine is already done and
a normal network startup should be guaranteed.

### DIFF
--- a/suse_migration_services/units/setup_host_network.py
+++ b/suse_migration_services/units/setup_host_network.py
@@ -67,12 +67,6 @@ def main():
         system_mount.add_entry(
             sysconfig_network, '/etc/sysconfig/network'
         )
-        Command.run(
-            ['systemctl', 'daemon-reload']
-        )
-        Command.run(
-            ['systemctl', 'restart', 'network']
-        )
         system_mount.export(
             Defaults.get_system_mount_info_file()
         )
--- a/systemd/suse-migration-setup-host-network.service
+++ b/systemd/suse-migration-setup-host-network.service
@@ -2,7 +2,7 @@
 Description=Setup Migration Host Network
 After=suse-migration-mount-system.service
 Requires=suse-migration-mount-system.service
-Before=network.target
+Before=network-pre.target
 
 [Service]
 Type=oneshot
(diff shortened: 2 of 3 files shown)


schaefi:
This change was inspired by a discussion with Tim Serong. I hope that Tim joins the discussion because we have to first test this change before we merge. Thanks

tserong:
Thanks @schaefi, I'll test this tomorrow (I'm in GMT+10:00) and report back

tserong:
In addition to removing the restart, I'd suggest also doing this:

```diff
--- suse_migration_services-0.5.24.orig/systemd/suse-migration-setup-host-network.service
+++ suse_migration_services-0.5.24/systemd/suse-migration-setup-host-network.service
@@ -2,7 +2,7 @@
 Description=Setup Migration Host Network
 After=suse-migration-mount-system.service
 Requires=suse-migration-mount-system.service
-Before=network.target
+Before=network-pre.target
 
 [Service]
 Type=oneshot
```

I've tested with @schaefi's original commit only, and with the above addition, and they both work for me, but IMO adding the above is still correct/useful, as `man systemd.special` says network-pre.target "may be pulled in by services that want to run before any network is set up, for example for the purpose of setting up a firewall. All network management software orders itself after this target, but does not pull it in."

(My original problem was that various wicked services weren't starting -- I've since discovered this may be some bizarre problem in my test environment, because if I hit the ESC key a few times during boot, wicked somehow suddenly starts correctly. This makes no sense to me at all. Still, I think this PR is worthwhile, it makes startup a bit cleaner.)

jesusbv:
LGTM

schaefi:
> In addition to removing the restart, I'd suggest also doing this:

Thanks fully agree and applied

PR #75, opened 2019-04-23
Update debug flag file to config migration file

jesusbv:
Modify the debug flag file to add the option to
decide what type of migration to run.

### DIFF
--- a/suse_migration_services/units/mount_system.py
+++ b/suse_migration_services/units/mount_system.py
@@ -110,7 +110,7 @@ def main():
     mount_system(root_path, fstab)
 
     initialize_logging()
-    initialize_system_config()
+    initialize_migration_config()
 
 
 def initialize_logging():
@@ -119,7 +119,7 @@ def initialize_logging():
         log.set_logfile(Defaults.get_migration_log_file())
 
 
-def initialize_system_config():
+def initialize_migration_config():
     migration_config_file = Defaults.get_system_migration_config_file()
     # delete potentially existing migration config file from migration live system
     Path.wipe(
@@ -135,10 +135,10 @@ def initialize_system_config():
         target_config_file_path = os.sep.join(
             ['/etc', os.path.basename(migration_config_file)]
         )
-        create_migration_file(target_config_file_path)
+        create_migration_config_file(target_config_file_path)
 
 
-def create_migration_file(target_config_file_path):
+def create_migration_config_file(target_config_file_path):
     data = Defaults.get_system_migration_config_values()
 
     with open(target_config_file_path, 'w') as config:
--- a/suse_migration_services/units/mount_system.py
+++ b/suse_migration_services/units/mount_system.py
@@ -16,6 +16,7 @@
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import yaml
 
 # project
 from suse_migration_services.command import Command
@@ -109,7 +110,7 @@ def main():
     mount_system(root_path, fstab)
 
     initialize_logging()
-    initialize_debugging()
+    initialize_system_config()
 
 
 def initialize_logging():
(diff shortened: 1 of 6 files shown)


schaefi:
I knew we will need a config file and should have insisted to do it from the beginning on instead of having that flag file. Anyway thanks for the implementation

schaefi:
Please give me some time to review, I'm currently adding the current state to the Stable project

jesusbv:
> I knew we will need a config file and should have insisted to do it from the beginning on instead of having that flag file. Anyway thanks for the implementation

You were absolutely right

suse_migration_services/units/mount_system.py
```diff
@@ -109,7 +110,7 @@ def main():
     mount_system(root_path, fstab)
 
     initialize_logging()
-    initialize_debugging()
+    initialize_system_config()
```
schaefi: Can we rename this to ```initialize_migration_config``` that imho is more explicit and clear

suse_migration_services/units/mount_system.py
```diff
@@ -118,17 +119,30 @@ def initialize_logging():
...
+            ['cp', migration_config_file, '/etc']
         )
+    else:
+        target_config_file_path = os.sep.join(
+            ['/etc', os.path.basename(migration_config_file)]
+        )
+        create_migration_file(target_config_file_path)
+
+
+def create_migration_file(target_config_file_path):
```
schaefi: Same here, can we change this to ```create_migration_config_file```

suse_migration_services/units/reboot.py
```diff
@@ -57,3 +54,19 @@ def main():
...
+    Returns True or False depending on debug mode
+    """
+    migration_config_file = os.sep.join(
+        ['/etc', os.path.basename(Defaults.get_system_migration_config_file())]
+    )
+    config = _read_system_migration_config(migration_config_file)
+    return config['debug']
+
+
+def _read_system_migration_config(config_file_path):
```
schaefi: I think the yaml.load() doesn't need to be put into an extra private method. I would just add that code in the is_debug_mode() code as it's used only once and basically fits into the code flow. I leave this part up to you though

jesusbv: I have removed the method and added that code inside ```is_debug_mode()```.
I was following the same approach as [get_migrate_product](https://github.com/SUSE/suse-migration-services/blob/fd890a2cf4d8440e98da92823b4e64ff2f51728b/suse_migration_services/units/migrate.py#L90).

schaefi:
Setting blocked for the moment because I'd like to finish the stable build before we merge further code changes. Thanks

schaefi:
Apart from the currently present merge conflict there is one part which I think should be changed.
We have decided to make use of a yaml config file to control certain aspects of the migration. So far
there is

* debug on|off
* setup of target product
* ...and I'm sure there will be more in the future

So in this pull request there is a function is_debug_mode() that reads the config file and reads the relevant entry from the dict. Pretty much the same would need to be done for the target setup and so on.

My suggestion is that we write a class

```python
class MigrationConfig(object):
    def __init__(self):
        migration_config_file = os.sep.join(
            ['/etc', os.path.basename(Defaults.get_system_migration_config_file())]
        )
        with open(migration_config_file, 'r') as config:
                self.config_data = yaml.safe_load(config)
```

and in that class we write methods to access relevant information from the config file, e.g

```python
def is_debug_requested(self):
    if self.config_data.get('debug') return True else return False
```

The actual service unit code can then simply do

```
from suse_migration_services.migration_config import MigrationConfig

migration_config = MigrationConfig()

if migration_config.is_debug_requested():
    ...
```

Imho this should be done looking at the open issues and what will be next we should have a clear
messaging model that puts the responsibility to read process configuration from a class instance

schaefi:
I will express my thoughts on the yaml in an example PR, stay tuned

jesusbv:
> I will express my thoughts on the yaml in an example PR, stay tuned

In [allow product](https://github.com/SUSE/suse-migration-services/pull/79) PR, I changed the yaml structure of the previously debug flag file. Now it is:
```yaml
debug: ['true' | 'false']
migration:
    type: ['migration' | 'dup']
    target:
        product: ['SLES/15/x86_64' | 'ses' | 'sles']
```
The debug and migration['type'] keys are coming from the user generated file.
If not present, ```debug: 'false'``` and ```migration['type']: 'migration'``` will apply

schaefi:
@jesusbv can the conflict be resolved please ? Thanks

PR #76, opened 2019-04-24
Fixed network requirement

schaefi:
To come up with a race free network start the setup-host-network
service has to be called after network.target. This ensures the
wicked service is already active. An active wicked service allows
for a reload of the network service. The setup-host-network
service overmounts /etc/sysconfig/network and reloads the network
service. That tells the running wicked service to elaborate
over the new ifcfg configurations and updates the network status.
In the prepare unit we require the network-online target which
makes sure it operates only on an online network.


jesusbv:
LGTM

PR #77, opened 2019-04-29
Split activation package from services

### LINKED ISSUE #74: Remove hard requirement on suse-migration-activation
Currently, the SLES15-Migration package requires suse-migration-activation, which adds the relevant grub config to reboot into the migration system. So, install the migration package, reboot, migrate, and you're done. This is ideal for single systems, but doesn't work for cases such as Ceph clusters where there are multiple hosts that need to be upgraded in a particular sequence.

To handle this case, can we remove the requirement for the activation package in SLES15-Migration? This would mean documenting that an activation step is required before reboot. For the simple case, that would mean documenting that you have to install suse-migration-activation. In the case of a Ceph cluster that would involve invoking some DeepSea/salt task to trigger the activation, which would be covered separately in the SUSE Enterprise Storage docs.

schaefi:
From a release process it doesn't make sense to keep the activation
script as a sub package of the services. The reason is because the
suse-migration-services package will be released to the migration
target distribution and the activation package will be released to
the migration start distribution. Thus the code is relevant on
different distributions and it is not useful to maintain that in
one package. This Fixes #74

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,9 +1,10 @@
 # Manifest describing source package contents
 
+prune grub.d
+
 graft suse_migration_services/units
 graft package
 graft systemd
-graft grub.d
 
 include tox.ini
 include README.md
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,15 @@ build: check test
 	# provide rpm rpmlintrc
 	cp package/suse-migration-services-rpmlintrc dist
 
+sle15_activation: check
+	rm -f dist/*
+	git log grub.d | helper/changelog_generator |\
+		helper/changelog_descending > \
+		dist/suse-migration-sle15-activation.changes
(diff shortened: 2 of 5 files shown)


schaefi:
The corresponding package build for the live migration image package build from here:

* https://build.suse.de/package/show/Devel:PubCloud:Stable:ImagesSLE124P/SLES12-SP4-SLES15-Migration

Also does not include a strict requirement to the activation package anymore

package/suse-migration-activation-spec-template
```diff
@@ -0,0 +1,46 @@
+#
+# spec file for package suse-migration-activation
+#
+# Copyright (c) 2018 SUSE LINUX GmbH, Nuernberg, Germany.
```
rjschwei: s/2018/2019/

It's almost May already time to catch up ;)

PR #78, opened 2019-05-01
Fix reboot race condition

schaefi:
The issue described here came up with the package split and
rename of the activation package. The grub setup service
deletes the activation package and calls grub2-mkconfig which
deletes the migration live boot entry as well as updates grub
to boot the migrated system. However if the activation package
could not be removed the process ends in a boot loop. Therefore
one part of this patch makes the removal of the migration
packages more robust. Another part provides detailed log
information about the removal and the grub setup in the log
file.

Last but not least the patch includes a refactoring of the
umount and reboot process which offers a race condition if
the grub-setup service has failed. The patch here also
Delete the umount service and put the code in the reboot
service. Only on reboot we umount the system. In case of
debugging we want to have full access to the root of the
system that should have become migrated. In addition the
systemd dependency chain for the kernel-load vs. reboot
service was racy and got fixed by a clear dependency
for grub-setup -> kernel-load -> reboot


schaefi:
I have done tests on EC2 to verify this is now race free and working also with regards to the package submissions that will follow to the SUSE namespace

suse_migration_services/units/grub_setup.py
```diff
@@ -32,28 +33,44 @@ def main():
...
-                '{0}{1}'.format(os.sep, grub_config_file)
-            ]
+        system_package_list = Command.run(
+            ['rpm', '--root', root_path, '-qa'], raise_on_error=False
+        ).output.split(os.linesep)
+        migration_packages = [
+            package for package in system_package_list if re.match(
+                r'suse-migration-.*-activation|SLE.*-Migration', package
+            )
+        ]
```
rjschwei: I think you can simplify the search for the migration packages. Rather then getting a list of all installed packages. We know that the migration packages leave certain files behind, thus we should be able to use

rpm --root .... -qf $A_FILE_FROM_ACTIVATION
rpm --root .... -qf $A_FILE_FROM_MIGRATION

and think of it that way we can simply check this without using rpm, I think. Getting the full list of all package of the system to be migrated could result in a list of thousands of packages. I thinks it's a heavy handed approach.

suse_migration_services/units/reboot.py
```diff
@@ -21,20 +21,30 @@
...
 from suse_migration_services.defaults import Defaults
+from suse_migration_services.fstab import Fstab
 
 
 def main():
     """
     DistMigration reboot with new kernel
+
+    After the migration process is finished, the system reboots
+    unless the debug file /etc/sle-migration-service is set.
```
rjschwei: Since the config file is now a real config file and no longer a flag, this wording needs to change.

`...unless debug: True is set in the ... configuration file.`

suse_migration_services/units/reboot.py
```diff
@@ -21,20 +21,30 @@
...
+
+    Before reboot a reverse umount the filesystems that got mounted
```
rjschwei: '...of the filesystems....' "of"' missing

suse_migration_services/units/reboot.py
```diff
@@ -21,20 +21,30 @@
...
+    by the mount_system service is performed and thus releases the
+    upgraded system from the migration host. If for whatever reason
+    a filesystem is busy and can't be umounted, this condition is not
+    handled as an error. The reason is that the cleanup should not
+    prevent us from continuing with the reboot process. The risk on
+    reboot of the migration host with a potential active mount
+    is something we take into account intentionally
```
rjschwei: s/we take into account intentionally/accept/

suse_migration_services/units/grub_setup.py
jesusbv: Agree with Robert, if we can avoid check in the configured repos, better as we know we dont have to check there.

PR #79, opened 2019-05-01
Allow specify target

### LINKED ISSUE #73: allow specifying product to migrate to
The current SLES15-Migration image is configured to try to migrate to SLES/15/x86_64 (in /etc/migration-config.yml). Can we make this configurable, via a file on the host system, which would specify the product to migrate to?

Rationale: we can't always just migrate to the latest released product. Here's two examples:

1) We might want to test migrations to an unreleased product.
2) We may be upgrading to a new product, which is older than the latest SLE release (think: some future SUSE Enterprise Storage user, who wants to upgrade from SLE 12 SP3 + SES5.5 to SLE 15 SP1 + SES 6, after SLE 15 SP2 has been released).

jesusbv:
Allow specify target product to migrate in optional
configuration file, following a defined structure.
It merges the new configuration into the provided file,
so the services handles only one file.
This Fixes #73

### DIFF
--- a/suse_migration_services/units/migrate.py
+++ b/suse_migration_services/units/migrate.py
@@ -84,9 +84,15 @@ def get_migration_product():
     zypper migration call
     """
     config = _read_migration_config()
-    return config['migration']['target']['product']
+    migration_product = config['migration']['target']['product']
+    target_products = dict(ses='SLES/15.1/x86_64', sles='SLES/15/x86_64')
+    if migration_product in target_products:
+        migration_product = target_products[migration_product]
+
+    log.info('The migration product is {0}'.format(migration_product))
+    return migration_product
 
 
 def _read_migration_config():
     with open(Defaults.get_migration_config_file(), 'r') as config:
-        return yaml.load(config)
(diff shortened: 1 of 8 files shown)


schaefi:
Can the conflict be resolved please ? Thanks

suse_migration_services/units/migrate.py
```diff
@@ -84,9 +84,15 @@ def get_migration_product():
     zypper migration call
     """
     config = _read_migration_config()
-    return config['migration']['target']['product']
+    migration_product = config['migration']['target']['product']
```
rjschwei: I thought we had agreed not to layer this and just use:

```
migration_product: $VALUE
```
rather than

```
migration:
  - target:
      - product: $VALUE
```

jesusbv: @rjschwei yes, this PR is old and [1](https://github.com/SUSE/suse-migration-services/pull/81) ahs the right structure.
I can update this code once [1] PR is merged to use that code and structure.

PR #81, opened 2019-05-02
Add target product keys

jesusbv:
Add different keys that represent different target products.

### DIFF
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -19,6 +19,10 @@ import yaml
 
 # project
 from suse_migration_services.defaults import Defaults
+from suse_migration_services.logger import log
+from suse_migration_services.exceptions import (
+    DistMigrationProductNotFoundException
+)
 
 
 class MigrationConfig(object):
@@ -42,11 +46,13 @@ class MigrationConfig(object):
 
         The value returned is passed to the --product option in the
         zypper migration call
-
-        migration:
-          target:
-            product: product_name
         """
-        migration = self.config_data.get('migration')
-        if migration and migration.get('target'):
-            return migration.get('target').get('product')
+        migration_product = self.config_data.get('migration_product')
+        if not migration_product:
+            message = (
+                'Migration product not found, aborting migration attempt.'
+            )
+            log.error(message)
+            raise DistMigrationProductNotFoundException(message)
+
+        return migration_product
--- a/test/data/migration-config.yml
+++ b/test/data/migration-config.yml
@@ -1,3 +1,1 @@
-migration:
-  target:
-    product: foo
+migration_product: 'SLES/15/x86_64'
--- a/suse_migration_services/exceptions.py
+++ b/suse_migration_services/exceptions.py
@@ -113,3 +113,9 @@ class DistMigrationProductSetupException(DistMigrationException):
     from the bind mounted etc/products.d location into the
     migrated system has failed
     """
+
+
+class DistMigrationProductNotFoundException(DistMigrationException):
+    """
+    Exception raised if the target product to migrate is not found
+    """
--- a/test/unit/migration_config_test.py
+++ b/test/unit/migration_config_test.py
@@ -1,7 +1,11 @@
 from unittest.mock import patch
+from pytest import raises
 
 from suse_migration_services.migration_config import MigrationConfig
 from suse_migration_services.defaults import Defaults
+from suse_migration_services.exceptions import (
+    DistMigrationProductNotFoundException
+)
 
 
 class TestMigrationConfig(object):
@@ -12,4 +16,12 @@ class TestMigrationConfig(object):
         self.config = MigrationConfig()
 
     def test_get_migration_product(self):
-        assert self.config.get_migration_product() == 'foo'
+        assert self.config.get_migration_product() == 'SLES/15/x86_64'
+
+    @patch('suse_migration_services.logger.log.error')
+    def test_get_migration_product_targets(self, mock_error):
+        self.config.config_data = {'not_migration_product': 'another_info'}
+
+        with raises(DistMigrationProductNotFoundException):
+            self.config.get_migration_product()
+            assert mock_error.called
--- a/test/unit/units/migrate_test.py
+++ b/test/unit/units/migrate_test.py
@@ -54,7 +54,7 @@ class TestMigration(object):
                 '--auto-agree-with-licenses '
                 '--strict-errors-dist-migration '
                 '--replacefiles '
-                '--product foo '
+                '--product SLES/15/x86_64 '
                 '--root /system-root '
                 '&>> /system-root/var/log/distro_migration.log'
             ]


suse_migration_services/migration_config.py
```diff
@@ -0,0 +1,63 @@
...
+        Returns the full qualified product
+
+        The value returned is passed to the --product option in the
+        zypper migration call
+
+        migration:
+          target:
+            product: product_name
+        """
+        migration = self.config_data.get('migration')
```
rjschwei: Let's call the config option "migration_product". I don't think we need sub-options for the migration as is currently implemented. Based on teh current implementation it would look as follows:

```
migration:
  - target: sles
```

I think

```
migration_product: sles
```

is sufficient. Then the code simplifies to:
```
migration_product = self.config_data.get('migration_product', 'sles')
```

schaefi: Fine by me

how does the red cape fit on :))

jesusbv: I have removed the sub-options, when/if approved, I will update ```/etc/migration-config.yml``` file we provide to match this structure.

suse_migration_services/migration_config.py
```diff
@@ -0,0 +1,57 @@
...
+
+        The value returned is passed to the --product option in the
+        zypper migration call
+
+        migration_product: product_name
+        """
+        migration_product = self.config_data.get('migration_product', 'sles')
+        if migration_product in self.target_products:
+            migration_product = self.target_products[migration_product]
+
```
rjschwei: This doesn't work this way.

Since we provide a default value to get() we know that is nothing is specified in the config we have an entry in target_products{}. If the user specifies a value in the config and specifies something we do not understand we have to exit and abort the migration.

As currently implemented if the user specifies `bar` as the migration product then `bar` would trickle onto the command line for `zypper migration` which would probably trigger an error and that OK. But hen we might as well dispense of the member test. This should simply be:

```
migration_target = self.config_data.get('migration_product', 'sles')
migration_product = self.target_products.get(migration_target, 'Fail')
```
"Fail" is now the default and will trigger a failure of migration. But we should probably do better:

```
migration_target = self.config_data.get('migration_product', 'sles')
migration_product = self.target_products.get(migration_target, 'Fail')
if migration_product == 'Fail':
    log.error('The configured migration product "%s" is not supported, aborting migration attempt' % migration_target)
```

schaefi:
@jesusbv I'm going to take a closer look if the conflict got resolved. That will reduce the number of lines to look at. Thanks

schaefi:
@jesusbv hope you don't mind I rebased to master to get this one conflict free

suse_migration_services/migration_config.py
```diff
@@ -31,6 +35,11 @@ class MigrationConfig(object):
     live image system. It is usually provided as a pre-setup
     version as part of the live migration image build.
     """
+    target_products = {
+        'ses': 'SLES/15.1/x86_64',
+        'sles': 'SLES/15/x86_64'
+    }
```
schaefi: I think we should not do that mapping here. The solution should be generic. This means in our image we pre-configure ```migration_product: _correct_value_```. The customer will have an opportunity to provide its own value by an overwrite and is allowed to shot himself. If we maintain a static list here the concept becomes inflexible. I suggest we take 1:1 what is written in the config file and do not map the value against a static mapping table

suse_migration_services/migration_config.py
```diff
@@ -43,10 +52,16 @@ def get_migration_product(self):
...
-        migration = self.config_data.get('migration')
-        if migration and migration.get('target'):
-            return migration.get('target').get('product')
+        migration_target = self.config_data.get('migration_product', 'sles')
+        migration_product = self.target_products.get(migration_target)
+
+        if not migration_product:
+            message = ('The configured migration product {0}'
+                       'is not supported, aborting migration attempt'
+                       .format(migration_target))
```
schaefi: Can we go and use the dedent module here: Example:

```python
message = dedent('''
    The configured migration product {0} is not supported,
    aborting migration attempt
''')

log.error(message.format(migration_target))
```

test/data/migration-config.yml
```diff
@@ -1,3 +1 @@
-migration:
-  target:
-    product: foo
+migration_product: sles
```
schaefi: again I don't think introducing our own names and maintain a mapping table is not a good idea

suse_migration_services/migration_config.py
jesusbv: I agree that the static list is less flexible and that customer should be able to set their own value for migration product. Following [this](https://github.com/SUSE/suse-migration-services/issues/73) conversation, one option is to check if the value provided matches any of the list else let the migration go with whatever value comes from the customer config (which would make the migration crash if the value is not valid).

rjschwei: The point of the mapping is that we get to control where it goes. If you open it up as "free form" we open a trap. So we end up with "I set SLES/15.4/x86_64 and started at SLES 12" and we'll end up telling the customer "sorry not supported". Fair enough but certainly not satisfactory from the customer point of view. Or we end up having a humongous test matrix to make sure things work. I strongly recommend against opening the trap and with it the stream of support requests. Support is expensive for us and we can avoid it by locking things in place.

We can put the triplet into the config, that would help when we have to support migration for other architectures. But we'd still have to check the triplet for "is this a supported target". A "free form do what you want" approach is IMHO a giant mistake.

schaefi: If an upgrade to the specified target is possible or not is defined by the response of the zypper migration plugin. It will respond to you with either repos that can upgrade to the requested target or it will fail.

There is no good reason why we should maintain that mapping table to provide an error message "sorry can't do this" which is going to be sent by the zypper migration plugin asking for an upgrade path.

For the upcoming implementation of a simple "zypper dup" there is even no target spec to be set because the origin repo setup that a customer has to do beforehand will be the driver of an update.

So I don't think my suggestion is a mistake. Handling the request for yes we can upgrade this or not is the responsibility of the zypper migration plugin. You told me support matters. Yes it does and I rather tell the customer look this is what SCC told me: what you want is not possible rather than look this is our stupid mapping table that told you we can't do it but maybe we could.

Sorry I don't want a mapping table it's just another source of trouble in my eyes

rjschwei: You're L3, we'll find out. Well, actually we probably will not find out as we usually don't hear back from support even if we put them into a tough position, but we'll probably hear from our partners.

Depending on SCC/SMT/RMT is a problem as there is no rule that enforces our support policies. For example, on a SLES 12 SP1 system I can run `zypper migration` and SLES 12 SP4 will be offered as a migration target. This is NOT supported. From SLES 12 SP1 the target that is supported, as in you will get help if things go wrong, is SLES 12 SP3.

Of course you will point out that this problem needs to be solve elsewhere, which is correct, and I agree. However, IMHO, as long as the problem exists letting the user walk into a trap when it can be prevented is not "the right thing to do".

I am not here to argue the point, I am just here to provide a different perspective and that maybe we should consider the potential user issues over "there is a workaround in the code to help the user avoid major issues". But if you think the work-around needs to go so we can be "technically correct" and let the user as you stated "shoot themselves in the foot" (paraphrased) then so be it.

rjschwei: https://bugzilla.suse.com/show_bug.cgi?id=1135243

jesusbv: > Handling the request for yes we can upgrade this or not is the responsibility of the zypper migration plugin

I agree that the responsibility whether a path is valid for migration or not belongs to zypper-migration plugin. My preference is 80% to let user provide whatever path and zypper-migration to decide and also to give the customer a nice experience but probably both of you have seen more similar cases and faced what can go wrong. 

So, as Robert did opening the bug, that invalid (available and not supported) paths should be fixed where appropriate. In the meantime, as I don't know how long it will take to have that fixed and I don't know what the consequences (test matrix, user experience, and so on Robert mentioned) of opening that door of let the customer migrate to whatever target they want are; we could compromise and make sure that the path is supported, maybe not with the mapping table but some kind of control until that bug is fixed, then update the code as removing that control and let the user provide any target they want to migrate and let zypper-migration plugin decides, as it is its responsibility.

One possible outcome of that approach is customers, who think a path is available because of a NOT supported path has been offered to them, end up thinking distro migration tool has limitations because it does not support that path. 

Thoughts?

rjschwei: It is a trade off, as pretty much everything we do is.

On the one hand we have the creation of potentially letting the user choose something that is not supported, on the other hand we have the creation of the impression that the migration tool has limitations. But if we are honest with ourselves the migration tool does have limitations and we are OK with that, I think. We did not want to implement a solution that "solves all migration problems in the world". We set out to solve one use case.

Also we have the trade off between being "technically correct" and "creating a safe environment for the user".

All things considered we are probably in a 50/50 situation and no matter which path we choose there will be second guessing. My personal background would favor the route that reduces the risk for the user. But expecting the user to "know what hey are doing" is equally valid.

If the translation table is not the preferred direction and we want to let the user specify the triplet then at the very least I would put a look up table of supported values in place and compare the triplet. Using the triplet now will have the advantage that the format will not have to change when we need to support more than x86_64.

Yes, implementing the lookup table with verification we take some responsibility as to what we allow the user to do and what not. I don't think this is a bad thing in this case. But again it is a 50/50 decision.

suse_migration_services/migration_config.py
jesusbv: Fixed.

suse_migration_services/migration_config.py
jesusbv: Not happy with the dictionary for targets but from user point of view, it makes sense to show them that it's not supported and avoid possible strange states left by a failed migration.

schaefi: From my point of view the mapping table just introduces product names that we have to own. It's a 1:1 mapping that does not more. The context is a translation from nameA to nameB and the benefit that we can react on an unknown name is something I don't see as a value. If a product name is set that is not known by the upgrade path in SCC the request response includes a well formulated error message saying that this product doesn't exist. From my perspective this is good enough and not considered a "strange state". The migration will fail without having any change or damage to the system. In default mode we will reboot and the user has a log file including an understandable error message.

I really don't see the need for the mapping table and it's not the job of the migration services to check on the integrity of the data provided for the migration. That's the job of the migration plugin. If you want to maintain a mapping table then do it there but it really makes no sense if the request answer does the validation on the server.

my 2cents. Less code to maintain is better

Martin-Weiss:
I might not understand the full picture of this discussion but I will try to give some feedback from a consultants point of view. 

When we upgrade SLES and there are multiple target versions available like for SLES12SP3 where we can go to SLES12SP4, SLES15GA or SLES15SP1 - IMO the default should be "to the latest available version provided via the connected SMT/RMT/SCC".
 
This should not require a configuration option. (Keep in mind that some customers might want to do a "down server upgrade" using this solution even in case it is supported using direct zypper dup/migrate "online" like from SLES 12 SP3 to SLES 12 SP4).

In case the customer does not want to use the latest version and i.e. wants to go from SLES12SP3 to SLES15GA a configuration option should be there that allows to specify the target to allow "overriding the default "to the latest".

I am not sure why the target example above is "ses" vs "sles" - as I assume the "zypper dup or migrate" should not be "add-on" dependent.

Btw. I can also foresee that we might want to use this technology for upgrades in CaaSP or when using SUSE Manager in the future ..

schaefi:
> This should not require a configuration option.

sure and this is also not the issue here. The image we deliver will have a product setup as you said. We just want to allow customers to overwrite it.

> In case the customer does not want to use the latest version and i.e. wants to go from SLES12SP3 to > SLES15GA a configuration option should be there that allows to specify the target to allow "overriding > the default "to the latest".

correct. Whether or not this is possible will be defined by the upgrade path that is returned by the request made to SCC. It's not in the responsibility of this services code to check or map or in any way massage this information. Either SCC tells me yes you can do that or no you can't do that. I also expect a good message back from the request. If there are issues they need to be solved in that area SMT/RMT/SCC ... whatever is responsible but not here

> I am not sure why the target example above is "ses" vs "sles"

and here we go and the first misunderstanding with that mapping table starts. As I said I don't want it, it will just cause us problems. Please let us just go with a free form name and don't try to be clever here. The name is send in an request to service (SCC) and it will respond back if it can do something useful with it or not. Which product names exists and whatever people invent in that area in the future is outside of our control. We can only loose with mapping tables. If a customer wants to know which product names and upgrade paths exists they go to the documentation of the service that provides this functionality but surely not to this project code here. With the mapping table we introduce we know it, but we don't know it either and the table will be probably mistaken from the beginning on.

@jesusbv Please go ahead without the mapping table and let's keep things damn simple. I take over responsibility of the decision was wrong and you guys can blame me and I owe you a beer next time we meet.

Let us please continue

Thanks

jesusbv:
> @jesusbv Please go ahead without the mapping table and let's keep things damn simple. I take over responsibility if that decision was wrong and you guys can blame me and I owe you a beer next time we meet.

Yes, based on your answer yesterday plus the  fact I was not too keen on the table (as mentioned before, I agree it's not this code's responsibility to decide that) I decided to remove it yesterday, Martin's feedback reinforces that and it is clear now.

Or next time we meet, we could do the same but to celebrate it was not wrong!

suse_migration_services/migration_config.py
```diff
@@ -31,6 +31,7 @@ class MigrationConfig(object):
     live image system. It is usually provided as a pre-setup
     version as part of the live migration image build.
     """
+
```
schaefi: superfluous empty line

suse_migration_services/migration_config.py
```diff
@@ -42,11 +43,5 @@ def get_migration_product(self):
...
         zypper migration call
-
-        migration:
-          target:
-            product: product_name
         """
-        migration = self.config_data.get('migration')
-        if migration and migration.get('target'):
-            return migration.get('target').get('product')
+        return self.config_data.get('migration_product', 'SLES/15/x86_64')
```
schaefi: There is no default that we should hard code. The default product setup happens in the image description. Let's keep this code free from any static product string:

```python
return self.config_data.get('migration_product')
```

test/data/migration-config.yml
```diff
@@ -1,3 +1 @@
-migration:
-  target:
-    product: foo
+migration_product: 'foo'
```
schaefi: If you like you can use a really existing product name in the test here

```yaml
migration_product: 'SLES/15/x86_64'
```

suse_migration_services/migration_config.py
jesusbv: Yes, I know. I plan to unify both files in our config file as agreed. I thought to do it in a different PR but more than happy to unify them here as well.

jesusbv: Also, if no default here, dict.get defaults to ```None``` which would end up in the command ```zypper migration --flags....--product None```

schaefi: yes that's why the class MigrationConfig() exists in migration_config.py and in there the method get_migration_product(). I would adapt that method to the changed layout and raise an exception if no product is specified

jesusbv: I'm not sure we need an exception, the setup of that product is optional and if no product is provided by the user, it should migrate to the default one in the image description.

jesusbv: Got it. Ignored what I said. I will implement that.

suse_migration_services/migration_config.py
```diff
@@ -42,11 +46,13 @@ def get_migration_product(self):
...
         """
-        migration = self.config_data.get('migration')
-        if migration and migration.get('target'):
-            return migration.get('target').get('product')
+        migration_product = self.config_data.get('migration_product')
+        if not migration_product:
+            message = (
+                'Migration product not found, aborting migration attempt.'
+            )
+            log.error(message)
```
schaefi: why use a message variable for static text ?

```python
log.error(
    'Migration product not found, aborting migration attempt.'
)
```

jesusbv: To re use it in the exception

rjschwei:
> I might not understand the full picture of this discussion but I will try to give some feedback from a consultants point of view.
> 
> When we upgrade SLES and there are multiple target versions available like for SLES12SP3 where we can go to SLES12SP4, SLES15GA or SLES15SP1 - IMO the default should be "to the latest available version provided via the connected SMT/RMT/SCC".

The point is that "to the latest available version provided" is likely not a supported migration, your example of SLES 12 SP3 to SLES 15 SP1 demonstrates that. And as pointed out in the bug I filed it appears that ot is currently not possible to move from SLES 12 SP3 to even SLES 15 GA without having access to the LTSS channel. Thus allowing such a migration in the Public Cloud means we will leave our users dead in the water for on-demand instances.

However, there is a very valid argument to be made, and it is being made here, that it should not be our responsibility to enforce the products support policy.

test/data/migration-config.yml
rjschwei: I agree we should use an actual value such that it functions as an example for the value we expect.

suse_migration_services/migration_config.py
jesusbv: @schaefi if you are ok with the message variable to be re-used in the exception, I will rebase the 7 commits (2b11417 to fb21d4d) into 1 and merge.

schaefi: Oh I see, sorry Jesus, overlooked that part. Sure that's all fine then

schaefi:
@jesusbv ok this looks all good now. Thanks for being so patient. Could you squash all into one please ?

Thanks much

Martin-Weiss:
> 
> 
> > I might not understand the full picture of this discussion but I will try to give some feedback from a consultants point of view.
> > When we upgrade SLES and there are multiple target versions available like for SLES12SP3 where we can go to SLES12SP4, SLES15GA or SLES15SP1 - IMO the default should be "to the latest available version provided via the connected SMT/RMT/SCC".
> 
> The point is that "to the latest available version provided" is likely not a supported migration, your example of SLES 12 SP3 to SLES 15 SP1 demonstrates that. 

Puh - this message was a shock for a moment ;-) - but https://susedoc.github.io/doc-sle/develop/SLES-upgrade/single-html/#fig.update.sle.paths-overview has a picture that shows it is supported to upgrade from 12SP3 to 15SP1 and with a SES 5 subscription SLES 12 SP3 seems to be supported for these customers until end Sep. 2020 based on https://www.suse.com/lifecycle/.

>And as pointed out in the bug I filed it appears that ot is currently not possible to move from SLES 12 SP3 to even SLES 15 GA without having access to the LTSS channel. Thus allowing such a migration in the Public Cloud means we will leave our users dead in the water for on-demand instances.
> 
> However, there is a very valid argument to be made, and it is being made here, that it should not be our responsibility to enforce the products support policy.

I believe from a suse-migration-services we should not care what is supported and what not as this might also change over time..

PR #82, opened 2019-05-17
User documentation

jesusbv:
Update the format to asciidoc

Update the installation process as it requires two packages.

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -1,0 +1,102 @@
+= SUSE Distribution Migration System
+Marcus Schäfer; Jesús Velázquez
+
+:toc:
+:icons: font
+:numbered:
+
+:Authors: Marcus Schäfer and Jesús Bermúdez Velázquez
+:Publication_Date: TBD
+:Latest_Version: 0.5.15
+:Contributors: 
+:Repo: https://github.com/SUSE/suse-migration-services[suse-migration-services]
+
+ifdef::env-github[]
+//Admonitions
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+[%hardbreaks]
+Authors: {Authors}
+Contributors: {Contributors}
+Publication Date: {Publication_Date}
+Latest Version: {Latest_Version}
+Code available: {Repo}
+
+== Concept
+The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration tool is the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
+
+The upgrade to a new major versions requires the system to be offline during the upgrade to avoid system inconsistencies that may leave the system in a state that does not allow recovery. This behavior is implemented using a Live Migration Image.
+
+The distribution migration system provides the live image and a component that modifies the bootloader configuration such that on the next boot the system will boot into the upgrade image. Once booted into the upgrade live image the following chain of services will be executed:
+
+1. detect the system to be upgraded
+2. mount the necessary file systems
+3. setup the network to match the network configuration of the system to be upgraded
+4. prepare the package manager for the upgrade task
+5. upgrade the system
+6. update the boot loader configuration
+7. unmount all mounted file systems
+8. reboot
+
+In case an error occurs prior to the start of the upgrade the system will be reverted to its original state.
+
+== Requirements
+Systems that are to be upgraded need to be registered. "Pay as you go"-instances in the Public Cloud are automatically registered to the SUSE operated update infrastructure. All other systems must be connected to the SUSE Customer Center (SCC), a Subscription Management Tool (SMT), or a Repository Management tool (RMT) server. For systems managed via SUSE Manager use the upgrade path provided by SUSE Manager. 
+
+== Recommendation
+During the upgrade it is only possible to log in via SSH key-based login. If your system is not configured for it, it is recommended that at least one of the users on the system has a `~/.ssh/authorized_keys` file with a private key accessible by the person executing the system upgrade.
+
+== Installation
+
+The distribution migration system is available from the Public Cloud module. Therefore this module has to be enabled on the system prior to upgrade. For running on-demand instances this module is already enabled.
+
+For data center customers it is recommended to continue to use the documented offline distribution upgrade using the next distribution version installation media.
+
+To install the distribution migration system run:
+
+[listing]
+tux > sudo zypper in SLE15-Migration suse-migration-sle15-activation
+
+== Run the Migration
+
+The `suse_migration_activation package` configures the boot loader to boot into the live image which will start the upgrade automatically. After the installation of the packages is complete, reboot the system to start the upgrade process.
+
+After the upgrade has started, the only way to access the system during the
+upgrade process is via ssh with a user called `migration`:
+
+[listing]
+tux > sudo ssh migration@IP_OF_INSTANCE
+
+[NOTE]
+There is no need to provide any other information or key. The known SSH keys on the system to be upgraded have been imported into the upgrade system. Password-based login is not possible.
+
+== Debugging the Migration
+In case of a upgrade failure, consult the migration log. The file `/etc/issue` will contain a pointer to the respective log file.
+
+In order to debug a failed upgrade, the file `/etc/sle-migration-service`
+must exist prior to the start of the upgrade process. This file will prevent the upgrade system from rewinding the setup steps and rebooting due to a failed upgrade, allowing the issue to be debugged.
+
+[listing]
+tux > ssh INSTANCE_USER@IP_OF_INSTANCE 'touch /etc/sle-migration-service'
+
+If the upgrade succeeded and that file is present, it will be removed and
+the system will reboot into the upgraded version.
+
+== After the Migration
+Whether the upgrade succeeded or not, a log file is available in
+`/var/log/distro_migration.log` and it will contain information
+about the upgrade process. If the upgrade failed and the issues have been successfully debugged, the file `/etc/issue` should be cleared.
+
+== Caveats and Unsupported Conditions
+* Configuration files that have been modified in the original system will not be overwritten by the upgrade process. The new version of the respective configuration file will be copied into the same directory with the file name extension `.rpmnew`. It is recommended to compare the existing and the new configuration files and make manual adjustments when needed.
+* Public Cloud instances from SUSE images have a custom `/etc/motd` file that makes a reference the distribution version. This needs to be updated manually after the upgrade.
+* Repositories not registered via SUSEConnect and added to the system manually will remain untouched.
+* For Public Cloud instances the metadata will not change. As far as the cloud framework is concerned, you will still be running a "SLES 12 SP4" instance even if you upgraded to SLES 15. This cannot be changed.
+* Upgrade is only possible for systems that have direct access to the root file system from the boot loader.
+* Upgrade is only possible for systems that use unencrypted root file systems, at the OS level. Encrypting the root device using a cloud framework encryption mechanism happens at a different level.
+* Upgrade has been tested for SLES 12 SP4 to SLES 15
--- a/doc/DC-distribution-migration-system
+++ b/doc/DC-distribution-migration-system
@@ -1,0 +1,15 @@
+## ----------------------------
+## Doc Config File for DAPS
+## DAPS User Guide
+## ----------------------------
+##
+## Basics
+
+MAIN=user_guide.adoc
+ADOC_POST=yes
+ADOC_TYPE=article
+#ADOC_IMG_DIR=adoc/images
+
+STYLEROOT="/usr/share/xml/docbook/stylesheet/suse2013"
+
+
--- a/doc/adoc/Building_Documentation.adoc
+++ b/doc/adoc/Building_Documentation.adoc
@@ -1,0 +1,50 @@
+= Building the Documentation
+Frank Sundermeyer <fs@suse.com>
+
+:toc:
+:icons: font
+:numbered:
+
+== Prerequisites
+
+The following packages from the repository
+https://download.opensuse.org/repositories/Documentation:/Tools/ are needed:
+
+* daps
+* suse-xsl-stylesheets
+
+You also need
+
+* optipng
+* perl-checkbot
+* rubygem-asciidoctor
+* w3m
+
+Install the packages without recommended pckages:
+
+[source]
+----
+zypper in --no-recommends daps suse-xsl-stylesheets optipng \
+perl-checkbot rubygem-asciidoctor w3m
+----
+
+== Building the Document
+
+To build the document, run one of the following commands:
+
+[source]
+----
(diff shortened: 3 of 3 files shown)


doc/user_quide.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+$ zypper in SLE15-Migration suse_migration_activation
+
+=== Run migration
+
+When the package has been installed, run:
+
+[listing]
+$ reboot
+
+When the migration has started, the only way to access the system during the
```
Martin-Weiss: Hm - that means that we can not login and debug in case the network setup does not work or in case there is a problem with the SSH keys or SSH configuration (what ever that might be).

Could we also login on the console with the migration or any other user? 
Or could we use the passwd of the system to be upgraded to allow user-login to the console while the upgrade is running? (i.e. in case the system we upgrade does not have any authorized keys)

doc/user_quide.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+
+=== Concept
+Upgrading from one release to a newer one having to skip a range of multiple
+releases introduces an upgrade problem.
+
+The Distribution Migration System provides an upgrade path from A to B,
+where A and B are to major distributions versions (for example, from SLES 12 SP4 to SLES 15 GA).
+
+=== Requirements
+The system to be migrated must be registered.
```
Martin-Weiss: must be registered to SCC, SMT or RMT.

(SUSE Manager and custom repo configs are not "yet" supported)

doc/user_quide.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+
+[listing]
+$ SUSEConnect --regcod [SLES_REGCODE]
+
+=== How does distribution migration system do it ?
+To achieve the migration between this two releases, the entire migration process
+is called from within a Live Migration Image.
```
Martin-Weiss: Which kernel / OS is used while in this Live Migration Image? (what to do in case the NIC or HDD needs a special or 3rd party driver?

doc/user_quide.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+[listing]
+$ ssh migration@IP_OF_INSTANCE
+
+NOTE: There is no need to provide any other information or key.
+
+=== Debugging the migration
+Once the migration has finished and the system has rebooted, the file
+'/etc/issue' points to the log of migration, in case it has failed.
+
+In order to debug a failed migration, a file '/system-root/etc/sle-migration-service'
```
Martin-Weiss: s/system-root/<root>

--> system-root only exists in the Live system..

doc/user_quide.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+must exists before the migration process starts. This file will prevent the
+system to reboot in case of a failed migration, allowing to debug the system.
+For example, as soon as the system started rebooting, user must create the file:
+
+[source,bash]
+----
+$ ssh migration@IP_OF_INSTANCE
```
Martin-Weiss: ssh root@ -> the user "migration" does not exist on the instance we want to upgrade.. that user only exists in the Live system AFAIK.

doc/user_quide.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+----
+
+If the migration succeeded and that file is present,
+it will be ignored and removed, the system will reboot normally
+into the newer version.
+
+=== After the migration
+Whether the migration succeeded or not, a log file is available in
+`/var/log/distro_migration.log` and it will contain information
+about the migration process.
```
Martin-Weiss: We also should document what it is required to cleanup /etc/issue in case of a failure and we should link to the official "support pack rollback" documentation for rollback in case the new system does not boot after the upgrade.

doc/user_quide.asciidoc
schaefi: > Hm - that means that we can not login and debug in case the network setup does not work or in 
> case there is a problem with the SSH keys or SSH configuration (what ever that might be).

correct. So this process is made for "no physical access to the machine possible". This means to install the migration packages some access method to the system which is not "a human being sits on the computer" must exist. Access to a machine with a password is considered insecure and so we don't allow that and require ssh.

* if the network could not be setup, login is the smallest problem ;) you will be returned to the former state without any changes. Given we have sorted out all the network dependency issues there is not much that should prevent a former working network to run again in the live migration system.

* if the access method to the instance was password based you will be on a black period until the migration is done.

schaefi: > Could we also login on the console with the migration or any other user?

No during migration only the migration user exists. This is on purpose to make clear the system is just there for running the migration not for working with it

schaefi: > Or could we use the passwd of the system...

Nope sorry. The image as we deliver does not contain any credentials. The migration user is restricted to ssh based auth_key login. All for security reasons

schaefi: We also don't take over any credentials from the system that gets migrated into the live migration system. Only public keys from the authorized_keys files from the existing users are taken into account. Again all for security reasons

doc/user_quide.asciidoc
schaefi: The live migration image is a build of the migration target. This means the SLE15-Migration package comes with a live migration system that is build against SLE15. This is also one of the requirements of the zypper migration plugin. You can migrate from e.g SLE12 to SLE15 only with the SLE15 version of zypper. So you can expect the live migration system to be always running in the target distro

> what to do in case the NIC or HDD needs a special or 3rd party driver

We can include whatever 3rd party driver is needed into the image build. But please note that the live migration image build is standard SLES and everything that we include must be available in some repo and SLE compliant

doc/user_quide.asciidoc
schaefi: good point :+1:

doc/user_quide.asciidoc
Martin-Weiss: Ok - then I think we really should advice the customer to add at least one SSH key to the authorized_keys file. 

And maybe we also should add an "unsecure" option that allows debugging of network problems on the servers console.

doc/user_quide.asciidoc
Martin-Weiss: > We can include whatever 3rd party driver is needed into the image build. But please note that the live migration image build is standard SLES and everything that we include must be available in some repo and SLE compliant

Do we have a documentation on how to rebuild and customize the ISO so that a customer or partner could add third party drivers on his own?
I do not believe that we really have this problem today - as the servers customers are using are SLES 12 SP* and YES certified hardware should be used - so expectation is that the same hardware works for SLES 15 SP* - but you never know for sure ;-).

doc/user_quide.asciidoc
schaefi: > Ok - then I think we really should advice the customer to add at least one SSH key

yes as a preparation task if not present I agree that would be good

> And maybe we also should add an "unsecure" option that allows debugging 

You mean access by a simple password that a user could set in the customer config file ? might be possible. If you like you can add that as an enhancement issue here

doc/user_quide.asciidoc
schaefi: There is no docs on how to rebuild the live iso. The live iso build is packaged into an rpm. So the customer has access to the rpm and binary blob inside. The building of that image is in the internal buildservice in a project. If you are a developer at suse you can simply run ```osc branch _project_``` and just add the needed modifications. In terms of 3rd party stuff that would mean adding an enterprise buildservice repo and some packages. The problem I see here is that we do not populate that build to the outside. Imho this is a complete new topic in which we could discuss that we also package the live image build descriptions and put some documentation into that package such that a customer can build them on their own. To be honest I don't want customers to build that live image to drive an upgrade process. The reason is that you can do all kinds of nasty things in that image and if we open this gate we also should tell customers that we don't take over any warranty. If we do that it means we don't care for the upgrade process and that in the end will not be helpful to customers. I see this part more as a service. A customer with special needs gets in contact with the suse consulting and then we build based on what we have. I see this more as a business opportunity

Martin-Weiss: Yeah - agreed. Might be Pandoras box.. lets go with L3 for the moment and see what comes in over time.

doc/user_quide.asciidoc
rjschwei: The doc would be https://www.suse.com/documentation/sles-12/book_sle_deployment/data/sec_update_migr_rollback.html which addresses service pack roll back. At this point we are on a new distribution. I have my doubts that "SUSEConnect --rollback" works backwards from SLES 15 to SLES 12 SP4. To have certainty this would have to be tested. Therefore from my perspective pointing the user at the "rollback" documentation would be misleading.

doc/user_quide.asciidoc
rjschwei: Customers that have a physical machine should continue to use the "boot from install ISO" method. I don't see us covering the general case of "use this anywhere SLES may run and with whatever configuration you might have". The install ISO handles this we have no intention of duplicating that functionality.

doc/user_quide.asciidoc
schaefi: Robert added a recommendation about ssh usage

doc/user_quide.asciidoc
schaefi: rollback in terms of the migration process means:

- restore of the repositories (backward the upgrade path, that's done by the zypper migration plugin)
- restore of the sle registration (also zypper migration plugin)

As soon as zypper started the transaction and the first rpm package got installed on the machine there is no reliable rollback available. The only way would be on systems with btrfs that allows to return to a former snapshot. The concept as we present it here however does not put any pre constraints on the machine. Whether or not a customer has snapshots or certain file system magic at hand is not in our hands.

So maybe it's a good idea to mention this in the documentation as well

doc/user_quide.asciidoc
Martin-Weiss: > 
> 
> Customers that have a physical machine should continue to use the "boot from install ISO" method. I don't see us covering the general case of "use this anywhere SLES may run and with whatever configuration you might have". The install ISO handles this we have no intention of duplicating that functionality.

At least for the SES usecase we have to support this method also for hardware based installations (SES is 99% on hardware). 
The "alternative method for automation of the upgrade using autoyast based upgrade or even manual upgrades using the DVD would be far to complicated and can not be standardized among all the different customer deployments out there which have different and complicated network setups".
However - lets do one step after the other and see where it goes - in the worst case we can still fallback to autoyast based or manual DVD based upgrades if this method does not work.

doc/adoc/user_guide.adoc
```diff
@@ -0,0 +1,117 @@
...
+== Installation
+
+The distribution migration system is made available in the Public Cloud module. Therefore this module has to be enable on the system to be migrated. For running on-demand instances this module is already enabled in the instance.
+
+For data center customers it is recommended to continue to use the documented offline distribution migration using the next distribution version installation media.
+
+To install the distribution migration system execute as root
+
+[listing]
+$ zypper in SLE15-Migration suse_migration_activation
```
schaefi: The activation package name is: suse-migration-sle15-activation

jesusbv: Fixed.

schaefi:
@fsundermeyer Could you have a short look on the current state ? For me it looks good, but it would be very helpful to see someone from the doc team review too. Thanks much

schaefi:
@jesusbv Could you merge the commits into one to be prepared once Frank approved ? Thanks

doc/adoc/user_guide.adoc
```diff
@@ -0,0 +1,117 @@
+= SUSE Distribution Migration System
+Marcus Schäfer; Jesús Velázquez
```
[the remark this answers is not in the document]
jesusbv: @fsundermeyer thank you for the feedback and suggestions.
Regarding the name, yes it is "Distribution Migration System" and it should remain like that.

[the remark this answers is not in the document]
schaefi: Yes that name was already send to the business people, so let's keep it

schaefi:
@jesusbv merge into one ?

@fsundermeyer thanks for your feedback. I think apart from the name which is set we've addressed everything. Are we good to go from your pov ?

Thanks

doc/adoc/user_guide.adoc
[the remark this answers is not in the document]
fsundermeyer: Sigh, I already assumed that the name is fixed. How about at least using upgrade (the verb) in the text (rather than migrate)?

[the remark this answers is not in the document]
jesusbv: @fsundermeyer, in the text., upgrade instead of migrate/migration or migrate ?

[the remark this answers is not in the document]
fsundermeyer: Yes. Not when the product name is referred to, but 

To start ~migrating/the migration~ run ...
To start upgrading/the upgrade run ...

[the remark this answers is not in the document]
jesusbv: Got it. I was applying it in that way.
I will apply the suggestions that were missed because they were hidden due to outdate, as well.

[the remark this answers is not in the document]
jesusbv: @fsundermeyer, I fixed this change and the other remaining suggestions.
Please, if you could give an extra look in case I missed something or there is a part that is still not clear, it would be great.

doc/adoc/user_guide.adoc
```diff
@@ -0,0 +1,102 @@
...
+The distribution migration system provides the live image and a component that modifies the bootloader configuration such that on the next boot the system will boot into the upgrade image. Once booted into the upgrade live image the following chain of services will be executed:
+
+- detect the system to be migrated
+- mount the necessary file systems
+- setup the network to match the network configuration of the system to be migrated
+- prepare the package manager for the migration task
+- migrate the system
+- update the boot loader configuration
+- unmount all mounted file systems
+- reboot
```
fsundermeyer: Sorry, I missed this (this will be my last comment, promised, please merge after fixing ;-)). Since this is a series of events that happen in order, we should  use an ordered list. The text also contains a few "migrations":

```
1. Detect the system to be migrated.
2. Mount the necessary file systems.
3. Setup the network to match the network configuration of the system to be upgraded.
4. Prepare the package manager for the upgrade task.
5. Upgrade the system.
6. Update the boot loader configuration.
7. Unmount all mounted file systems.
8. Reboot.
```

doc/adoc/user_guide.adoc
```diff
@@ -0,0 +1,102 @@
...
+upgrade process is via ssh with a user called `migration`:
+
+[listing]
+tux > sudo ssh migration@IP_OF_INSTANCE
+
+[NOTE]
+There is no need to provide any other information or key. The known SSH keys on the system to be upgraded have been imported into the upgrade system. Password-based login is not possible.
+
+== Debugging the Migration
+In case of a upgrade failure, consult the migration log. The file `/etc/issue` will contain a pointer to the respective log file.
```
jesusbv: @fsundermeyer I have left this ```migration``` because the name of the log file is 
```distro_migration.log```

schaefi: looks good to me

rjschwei:
Just a comment on the choice of terminology and why we ended up with "migration". First and formost it matches the `zypper` command line invocation that has been established as a name. Second we have established terminology for "updated" and in service package update, and "upgrade" a move from one service pack to another the alter is also establish by `zypper` naming as in `zypper dup`. The water gets muddled a bit as `zypper migration` is used for SP to SP upgrades and now also for major distribution upgrades. In any event we overloaded the term "upgrade" and then later we managed to overload the term "migration". This is unfortunate having `zypper upgrade` for SP to SP upgrade and `zypper migration` for major release upgrades would have made things nicer w.r.t. terminology.

Anyway, it is what it is.

PR #83, opened 2019-05-17
- Refine the user documentation for the migration system

rjschwei:
+ Include the caveates
  + Some structural changes

### DIFF
--- a/doc/user_quide.asciidoc
+++ b/doc/user_quide.asciidoc
@@ -14,62 +14,53 @@ Latest Version: {Latest_Version}
 Code available: {Repo}
 
 === Concept
-Upgrading from one release to a newer one having to skip a range of multiple
-releases introduces an upgrade problem.
-
-The Distribution Migration System provides an upgrade path from A to B,
-where A and B are to major distributions versions (for example, from SLES 12 SP4 to SLES 15 GA).
-
-=== Requirements
-The system to be migrated must be registered.
+The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from on major distribution to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For system updates from one Service Pack (SP) to another within a given major distribution release series the existing
 
 [listing]
-$ SUSEConnect --regcod [SLES_REGCODE]
+zypper migration
 
-=== How does distribution migration system do it ?
-To achieve the migration between this two releases, the entire migration process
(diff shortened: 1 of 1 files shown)


rjschwei:
@schaefi @jesusbv if you would like to discuss the rational behind my changes please let me know and we can hop on mumble

doc/user_quide.asciidoc
```diff
@@ -14,85 +14,74 @@ Latest Version: {Latest_Version}
...
 
-. fetch the suse-migration-activation package which contains an
-activation package:
-+
-[listing]
-$ zypper ar official_repo_url suse_migration_activation
+To install the distribution migration system execute as root
 
-. install it:
 +
```
jesusbv: s/+/

doc/user_quide.asciidoc
```diff
@@ -14,85 +14,74 @@ Latest Version: {Latest_Version}
...
 [listing]
 $ zypper in SLE15-Migration suse_migration_activation
```
jesusbv: should we not point to the repos to find those packages?

doc/user_quide.asciidoc
```diff
@@ -103,3 +92,12 @@ into the newer version.
 Whether the migration succeeded or not, a log file is available in
 `/var/log/distro_migration.log` and it will contain information
 about the migration process.
+
+=== Caveats and unsupported conditions
+* Files marked as configuration files in RPM packages and modified will have a corresponding '.rpmnew' version in the same location.
+** Public Cloud instances from SUSE images have a custom '/etc/motd' file that makes reference the distribution version this needs to be modified manually
+* Repositories not registered via SUSEConnect and added to the system manually will remain untouched.
+* For Public Cloud instances the metdata will not change, as far as the cloud framework is concerned you will still be running a SLES 12 SP4 instance even if you migrated the instance to SLES 15. This cannot be changed.
```
jesusbv: s/metdata/metadata

doc/user_quide.asciidoc
schaefi: Let's discuss this on the main PR branch. I agree the docs should include the location from where to get the packages. This would be the SLE12 Public Cloud Module. However in the public cloud environment and in an on-demand instance those repos will already be registered through our SMT/RMT infrastructure. I also think this should be mentioned with regards to use in public clouds

schaefi:
I will merge this back to the open doc PR branch. A few parts are still missing imho:

* mention source repos
* mention how things looks like on a public cloud instance

PR #88, opened 2019-05-22
Use one configuration file

jesusbv:
In case of user provides a configuration file,
that info will be added to the default configuration file.

### DIFF
--- a/suse_migration_services/defaults.py
+++ b/suse_migration_services/defaults.py
@@ -39,9 +39,9 @@ class Defaults(object):
         )
 
     @classmethod
-    def get_system_migration_debug_file(self):
+    def get_system_migration_custom_config_file(self):
         return os.sep.join(
-            [self.get_system_root_path(), 'etc/sle-migration-service']
+            [self.get_system_root_path(), 'etc/sle-migration-service.yml']
         )
 
     @classmethod
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -16,6 +16,8 @@
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
 import yaml
+import os
+from textwrap import dedent
 
 # project
 from suse_migration_services.defaults import Defaults
@@ -36,8 +38,12 @@ class MigrationConfig(object):
     version as part of the live migration image build.
     """
     def __init__(self):
-        self.config_data = None
-        with open(Defaults.get_migration_config_file(), 'r') as config:
+        self.config_data = {}
+        self.migration_config_file = \
+            Defaults.get_migration_config_file()
+        self.migration_custom_file = \
+            Defaults.get_system_migration_custom_config_file()
+        with open(self.migration_config_file, 'r') as config:
             self.config_data = yaml.safe_load(config)
 
     def get_migration_product(self):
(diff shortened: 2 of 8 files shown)


jesusbv:
The structure off the config files is the old one. It will be updated when [81](https://github.com/SUSE/suse-migration-services/pull/81), to do one change at the time.

test/data/optional-migration-config.yml
```diff
@@ -0,0 +1,3 @@
+migration:
+  target:
+    product: bar
```
rjschwei: This does not match with #81

jesusbv: Yes, that's what I commented [here](https://github.com/SUSE/suse-migration-services/pull/88#issuecomment-494753645), once #81 is merged.

suse_migration_services/defaults.py
```diff
@@ -39,7 +39,7 @@ def get_migration_log_file(self):
         )
 
     @classmethod
-    def get_system_migration_debug_file(self):
+    def get_system_migration_opt_config_file(self):
```
schaefi: Can we name this differently ? e.g

get_system_migration_custom_config_file()

suse_migration_services/units/mount_system.py
```diff
@@ -109,7 +110,7 @@ def main():
     mount_system(root_path, fstab)
 
     initialize_logging()
-    initialize_debugging()
+    update_config_info()
```
schaefi: Can we name this

update_migration_config_file()

It would be more in line with the other config file related methods.

suse_migration_services/units/mount_system.py
```diff
@@ -118,17 +119,10 @@ def initialize_logging():
...
-    )
-    # copy debug flag file if set on target system to migration live system
-    if os.path.exists(debug_file):
-        Command.run(
-            ['cp', debug_file, '/etc']
-        )
+def update_config_info():
+    # update the default configuration info with provided data, if any
+    if os.path.exists(Defaults.get_system_migration_opt_config_file()):
+        MigrationConfig().update_config_info()
```
schaefi: I think this method belongs to the MigrationConfig class.

test/data/optional-migration-config.yml
jesusbv: Fixed.

suse_migration_services/defaults.py
jesusbv: Fixed.

suse_migration_services/units/mount_system.py
jesusbv: Fixed the name and now it calls directly to MigrationConfig method

suse_migration_services/migration_config.py
```diff
@@ -56,3 +58,24 @@ def get_migration_product(self):
...
+            default_migration_config_file = Defaults.get_migration_config_file()
+            with open(default_migration_config_file, 'w') as config:
+                yaml.dump(self.config_data, config, default_flow_style=False)
+                message = dedent('''
+                The migration file '{0}' has been updated with '{1}' info
+                ''')
+                log.info(message.format(
+                    default_migration_config_file,
+                    migration_custom_file
+                ))
```
schaefi: great now :+1: just nitpick on the formatting. Can we change this to

```python
message = dedent('''
    The migration file '{0}' has been updated with '{1}' info
''')
log.info(
    message.format(
        default_migration_config_file, migration_custom_file
    ).lstrip()
)
```

schaefi: I will leave it up to you since it's only a style issue. Usually one avoids multiple opening/closing brackets in one line though

test/unit/migration_config_test.py
```diff
@@ -25,3 +26,21 @@ def test_get_migration_product_targets(self, mock_error):
...
+    ):
+        mock_get_system_migration_config_custom_file.return_value = \
+            '../data/optional-migration-config.yml'
+        mock_get_migration_config_file.return_value = \
+            '../data/combined-migration-config.yml'
+
+        self.config.update_migration_config_file()
+        assert self.config.get_migration_product() == 'SLES/15.1/x86_64'
+        assert mock_info.called
+        os.remove('../data/combined-migration-config.yml')
```
schaefi: This test has room for improvement. The code actually writes the file and your test re-reads it and after that deletes the file. You expect the system that runs your tests to have permissions to write on the target storage. This is an assumption which could be wrong especially for running unit tests. So normally you would mock the open call and perform any read/write check in memory. See units/kernel_load_test.py or fstab_test.py if you need examples how to mock code that opens and write files.

We can go with this version but it's not so nice :)

PR #91, opened 2019-05-22
- Enhance documentation

rjschwei:
+ Provide recommendation about ssh key access
  + State that clean up of /etc/issue is needed


schaefi:
@rjschwei looks good to me. I introduce the merge conflict with the daps changes by Frank. I'll fix your branch and merge it. Thanks

Issue #96, opened 2019-05-28
Migration system unable to boot on host with mdraid


tserong:
When using mdraid for the system root disk, there are a couple of problems:

1) /etc/grub.d/99_migration doesn't correctly pick up the root uuid and filesystem type. The `lsblk` invocation returns two items ("/dev/md1 /dev/md1"), which breaks the subsequent `blkid` invocations. This in turn results in a two grub errors when you boot into the migration system: "error: one argument expected" and "error: no such device: root."

2) Grub still manages to boot the initrd from the migration system, but this ultimately lands in an emergency shell, apparently because `/sbin/iso-scan` fails. AFAICT from digging around in the emergency shell, iso-scan is failing because there's no symlink created in /dev/disk/by-uuid that points to the migration ISO. If it helps, here's
[rdsosreport.txt](https://github.com/SUSE/suse-migration-services/files/3228007/rdsosreport.txt).

tserong:
I've made some progress with item 2:

- The `mdadm` package needs to be added to the migration image kiwi config -- without this, the initrd has no MD support
- `rd.auto` needs to be added to the end of the kernel line in the grub config.

This at least allows the migration system to *boot*, but then suse-migration-mount-system.service fails for some reason:
![failed-to-mount](https://user-images.githubusercontent.com/754700/58494204-dd453300-8174-11e9-9539-7736aeca7a68.png)

Still, it's progress. I'll investigate more tomorrow.

schaefi:
> The mdadm package needs to be added to the migration image 

done now

schaefi:
> rd.auto needs to be added to the end of the kernel line in the grub config.

This should be part of the activation script, the loopback grub entry there transports the kernel options.
If the option can be set without having a negative influence on the standard case we can simply add it there. If not the code needs to check if root runs on an md device.

Thanks for your work

tserong:
> If the option can be set without having a negative influence on the standard case we can simply add it there. If not the code needs to check if root runs on an md device.

`man dracut.cmdline` says rd.auto or rd.auto=1 "enable autoassembly of special devices like cryptoLUKS, dmraid, mdraid or lvm. Default is off as of dracut version >= 024". I don't know for certain if it will have a negative influence on the standard case, but suspect the safest thing to do is only add it if we know root runs on an md device. If we discover it's needed for other cases in future, we can always add more checks, or re-evaluate the decision.

schaefi:
[...]

PR #97, opened 2019-05-28
Use `uniq` when finding root device in grub activation script

tserong:
If the system is using mdraid on the root disk, the `lsblk` invocation
will return two lines matching "/$", e.g.:

/dev/md1
/dev/md1

This then breaks the subsequent `blkid` invocations, so $root_uuid and
$root_type are set to empty strings.  When you later boot into the
migration system, grub gives two errors:

error: one argument expected.
error: no such device: root.

Adding `uniq` fixes this.


tserong:
This fixes problem 1 in #96

PR #98, opened 2019-05-28
Add support for zypper dup mode

### LINKED ISSUE #72: Support zypper dup with manual repo setup
There are some cases where `zypper migration` can't work, for example sites with no direct access to SCC, or with manually configured mirrors/repos, or possibly when using SUMA. I think it makes sense to use `zypper migration` by default, but for cases where the admin knows this is not possible (for whatever reason), can we add a config file to the host system, which would trigger calling `zypper dup` instead? In this case it would be the admin's responsibility to add/remove the necessary repositories before rebooting into the migration system.

schaefi:
By default the migration system uses the zypper migration plugin
to migrate the system. This works if the system is SLES registered
and an upgrade path from SCC exists. For system where this is not
possible an alternative method can be switched on with the
custom configuration option use_zypper_migration_plugin set
to: false. In this mode we just call zypper dup. Please note
this requires that the user has pre setup the repositories before
the migration system gets started. This Fixes #72

### DIFF
--- a/suse_migration_services/units/migrate.py
+++ b/suse_migration_services/units/migrate.py
@@ -24,7 +24,8 @@ from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import log
 
 from suse_migration_services.exceptions import (
-    DistMigrationZypperException
+    DistMigrationZypperException,
+    DistMigrationCommandException
 )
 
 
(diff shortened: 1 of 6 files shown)


schaefi:
One can activate the mode by setting the following in ```/etc/sle-migration-service.yml``` on the host to migrate:

```yaml
use_zypper_migration_plugin: false
```

schaefi:
Please note before reboot into the live migration system the repos must be manually setup in a way that zypper dup can be successful

suse_migration_services/units/migrate.py
```diff
@@ -39,23 +40,47 @@ def main():
...
-                '--product', MigrationConfig().get_migration_product(),
-                '--root', root_path,
-                '&>>', Defaults.get_migration_log_file()
-            ]
-        )
-        Command.run(
-            ['bash', '-c', bash_command]
-        )
+        migration_config = MigrationConfig()
+        if migration_config.is_zypper_migration_plugin_requested():
```
jesusbv: Reading this part of the code, it seems that this case is not the default one.

schaefi: sorry don't understand what you mean ?

jesusbv: It's just nitpicking, it gives the impression that the use of zypper migration plugin is the special case.

schaefi: Oh that's what you mean. Well frankly speaking the migration plugin is a special case :) Usually people upgrade via ```zypper dup``` ;) If we want to change this I think also the logic of the option in the config file should change. I choose that way because the migration plugin has this clear name and is a plugin that one can use or not. ```zypper dup``` on the other hand is just an option to zypper and the long form is _distribution-upgrade_. I didn't find a nice name to express this mode in the config file and so I choose to go with "migration_plugin_requested y|n"

Currently I don't think we need to change this, but if you think it has a big potential for user confusion or misunderstanding we can of course reverse the logic ?

jesusbv: I don't think we should change it either. It is clear that unless user provides another option, the zypper migration plugin is the default. Let's keep it like that and we will check with feedback in the future.

schaefi: ok, also remember there is #99 so we also document it and that should make the default clear.
Good then I'll merge it ;)

Thanks

PR #102, opened 2019-05-29
Fix grub setup if root is on a raid device

schaefi:
If the root filesystem is on a raid device the boot parameter
rd.auto must be passed to the boot such that the dracut raid
module can setup the raid prior to iso-scan.sh searching
through the devices. Related to Issue #96


schaefi:
great thanks for testing :+1:

PR #103, opened 2019-05-29
Try to mount raid1 partitions

### LINKED ISSUE #96: Migration system unable to boot on host with mdraid

tserong:
`lsblk` returns "raid1" for /dev/md devices, so we need to attempt
to mount these as well as regular partitions in case the root disk
is on mdraid. Fixes #96.

### DIFF
--- a/suse_migration_services/units/mount_system.py
+++ b/suse_migration_services/units/mount_system.py
@@ -77,7 +77,7 @@ def main():
     log.info('Mount system service: mounting all entries')
     for entry in lsblk_call.output.split(os.linesep):
         block_record = entry.split()
-        if block_record and block_record[1] == 'part':
+        if block_record and (block_record[1] == 'part' or block_record[1].startswith('raid')):
             try:
                 Command.run(
                     ['mount', block_record[0], root_path],


schaefi:
Thanks. Do we want to limit this to raid1 particularly ? Thinking about something like

```python
block_record[1].starts_with('raid')
```

Thoughts ?

tserong:
Good point. I don't think there's any reason to limit this to just raid1. Shall tweak and re-test.

schaefi:
Thanks, if you've done the change we are ready to merge

tserong:
Tweaked and re-tested (finally :-/)

PR #104, opened 2019-05-29
- Cosmetic changes to handle error codes from zypper

rjschwei:
+ Make it more explicit to the reader that we treat specific values
    from zypper as errors. Positive side effect is that if the list gets
    longer the condition does not need to change


jesusbv:
Looks good. It needs to have 2 spaces before inline comments.

rjschwei:
> Looks good. It needs to have 2 spaces before inline comments.

Fixed, thanks

PR #105, opened 2019-05-30
Inherit ssh host keys

### LINKED ISSUE #86: Inherit ssh host key to live migration system
When accessing the Live system via SSH we always get a warning that the SSH key of the server changed which normally can indicate a man-in-the-middle attack.
Would it be possible to inject the SSH keys of the running server into the Live system?

(not sure if the current behavior is at will)

jesusbv:
The live image has no host key by default, provoking
a MITM warning message when login through ssh.

To avoid this situation, the ssh host keys
are added to the live system.

This Fixes #86

### DIFF
--- a/suse_migration_services/defaults.py
+++ b/suse_migration_services/defaults.py
@@ -75,3 +75,13 @@ class Defaults(object):
         return os.sep.join(
             [self.get_system_root_path(), prefix_path, '.ssh/authorized_keys']
         )
+
+    @classmethod
+    def get_system_ssh_host_keys_glob_path(self):
+        return os.sep.join(
+            [self.get_system_root_path(), 'etc/ssh/ssh_host_*']
+        )
+
+    @classmethod
+    def get_system_sshd_config_path(self):
+        return '/etc/ssh/sshd_config'
--- a/suse_migration_services/units/ssh_keys.py
+++ b/suse_migration_services/units/ssh_keys.py
@@ -16,10 +16,13 @@
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
 import glob
+import shutil
+import os
 
 # project
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import log
+from suse_migration_services.command import Command
 
 
 def main():
@@ -31,6 +34,8 @@ def main():
     """
     ssh_keys_glob_paths = Defaults.get_ssh_keys_paths()
     migration_ssh_file = Defaults.get_migration_ssh_file()
+    system_ssh_host_keys_glob_path = Defaults.get_system_ssh_host_keys_glob_path()
+    sshd_config_path = Defaults.get_system_sshd_config_path()
     try:
         log.info('Running ssh keys service')
         ssh_keys_paths = []
@@ -47,6 +52,35 @@ def main():
         log.info('Save keys to {0}'.format(migration_ssh_file))
         with open(migration_ssh_file, 'w') as authorized_migration_file:
             authorized_migration_file.write(authorized_keys_content)
+
+        system_ssh_host_keys = glob.glob(system_ssh_host_keys_glob_path)
+        sshd_config_host_keys_entries = []
+        log.info('Copying host ssh keys')
+        for system_ssh_host_key in system_ssh_host_keys:
+            shutil.copy(system_ssh_host_key, '/etc/ssh/')
+
+            if not system_ssh_host_key.endswith('.pub'):
+                live_private_ssh_host_key_path = os.sep.join(
+                    [
+                        os.path.dirname(sshd_config_path),
+                        os.path.basename(system_ssh_host_key)
+                    ]
+                )
+                entry = 'HostKey {0}'.format(
+                    live_private_ssh_host_key_path
+                )
+                sshd_config_host_keys_entries.append(entry)
+
+        with open(sshd_config_path, 'a') as live_sshd_config_file:
+            # write one newline to be sure any subsequent HostKey entry starts correctly
+            live_sshd_config_file.write(os.linesep)
+            live_sshd_config_file.write(
+                os.linesep.join(sshd_config_host_keys_entries)
+            )
+        log.info('Restarting sshd')
+        Command.run(
+            ['systemctl', 'restart', 'sshd']
+        )
     except Exception:
         log.error('An error ocurred when copying ssh key files.'
                   'The migration will continue without ssh access')
--- a/test/unit/units/ssh_access_test.py
+++ b/test/unit/units/ssh_access_test.py
@@ -1,5 +1,5 @@
 from unittest.mock import (
-    patch
+    patch, call
 )
 import os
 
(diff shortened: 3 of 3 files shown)


suse_migration_services/defaults.py
```diff
@@ -75,3 +75,9 @@ def _get_ssh_keys_path(self, prefix_path):
         return os.sep.join(
             [self.get_system_root_path(), prefix_path, '.ssh/authorized_keys']
         )
+
+    @classmethod
+    def get_system_ssh_host_keys_glob_path(self):
+        return os.sep.join(
+            [self.get_system_root_path(), 'etc/ssh/ssh_host_*']
+        )
```
schaefi: The process here is not fail safe. We do a glob match and the files gets copied 1:1 to the live migration system. Whether or not those ssh host keys will be really used depends on the setup in ```/etc/ssh/sshd_config```. As we go with the default setup sshd comes with a prebuilt list of ssh host keys that gets read. This list can be different compared to sles12 vs. sles15. So the concept here should be changed in a way that we copy all existing host keys and additionally add an entry of the form:

```
HostKey /path/to/key
```

to the private host key(s) that was found into ```/etc/ssh/sshd_config```

suse_migration_services/units/mount_system.py
```diff
@@ -59,6 +61,7 @@ def main():
     isoscan_loop_mount = '/run/initramfs/isoscan'
     if is_mounted(isoscan_loop_mount):
         # The system to become migrated was booted via a grub
+
```
schaefi: superfluous new line

suse_migration_services/units/mount_system.py
```diff
@@ -119,6 +123,16 @@ def initialize_logging():
...
 
+def inherit_ssh_host_keys():
+    system_ssh_host_keys_glob_path = Defaults.get_system_ssh_host_keys_glob_path()
+
+    if os.path.exists(os.path.dirname(system_ssh_host_keys_glob_path)):
+        system_ssh_host_keys = glob.glob(system_ssh_host_keys_glob_path)
+
+        for system_ssh_host_key in system_ssh_host_keys:
+            shutil.copy(system_ssh_host_key, '/etc/ssh/')
+
```
schaefi: The copy is not enough, sshd must be restarted after new host keys got added or changed

schaefi: In addition why is the ssh host key handling done in the mount_system service. This seems to be the wrong place. We have an explicit ssh_keys service. imho all of this belongs there not in mount_system

jesusbv: I put it in mount_system so keys get copied as soon as possible.

schaefi: That's not needed as you need to restart sshd there is no hurry to do this. In addition nobody could login prior to setting up authorized_keys in the ssh-keys service. So from my perspective the ssh-keys service is the right place and it should do:

1. get the host keys 
2. restart sshd
3. get the pubkeys from the users and put in authorized keys (already implemented)

schaefi: So looks better now, thanks much

jesusbv: Good. If all ok, I will squash them into one and merge.

suse_migration_services/units/ssh_keys.py
```diff
@@ -47,6 +52,34 @@ def main():
         log.info('Save keys to {0}'.format(migration_ssh_file))
         with open(migration_ssh_file, 'w') as authorized_migration_file:
             authorized_migration_file.write(authorized_keys_content)
+
+        if os.path.exists(os.path.dirname(system_ssh_host_keys_glob_path)):
+            system_ssh_host_keys = glob.glob(system_ssh_host_keys_glob_path)
```
schaefi: It's not needed to check for the presence of the directory. The glob match will return an empty list for any reason of the pattern not matching. Thus it's enough to go with:

```python
system_ssh_host_keys = glob.glob(system_ssh_host_keys_glob_path)
if system_ssh_host_keys:
    ...
```

suse_migration_services/units/ssh_keys.py
```diff
@@ -47,6 +52,34 @@ def main():
...
+                    )
+                    entry = 'HostKey {0}'.format(
+                        live_private_ssh_host_key_path
+                    )
+                    sshd_config_host_keys_entries.append(entry)
+
+            with open(sshd_config_path, 'a') as live_sshd_config_file:
+                live_sshd_config_file.write(
+                    os.linesep.join(sshd_config_host_keys_entries)
+                )
```
schaefi: One newline for safety reasons is missing here:

```python
# write one newline to be sure any subsequent HostKey entry starts correctly
live_sshd_config_file.write(os.linesep)

live_sshd_config_file.write(
    os.linesep.join(...
```

test/unit/units/ssh_access_test.py
```diff
@@ -18,23 +18,70 @@ def test_migration_continues_on_error(
...
+        mock_get_ssh_key_path.return_value = ['/path/']
+        mock_glob_glob.side_effect = [
+            [
+                '/path/authorized_azure_keys',
+                '/path/authorized_ec2_keys'
+            ],
+            [
+                '/system-root/etc/ssh/ssh_host_ecdsa_key.pub',
+                '/system-root/etc/ssh/ssh_host_ecdsa_key'
+            ]
```
schaefi: I still don't get why the test is messing with the authorized_keys testing. For the purpose of adding the host keys there should not be any change needed on the side of the auth key setup. Please explain the motivation for this change and if we need it put it in an extra pull request. Thanks

test/unit/units/ssh_access_test.py
```diff
@@ -18,23 +18,70 @@ def test_migration_continues_on_error(
...
+            assert file_handle_authorized_keys.write.call_args_list == [
+                call('keys for azure\nkeys for aws ec2')
+            ]
+            assert mock_shutil_copy.call_args_list == [
+                call('/system-root/etc/ssh/ssh_host_ecdsa_key.pub', '/etc/ssh/'),
+                call('/system-root/etc/ssh/ssh_host_ecdsa_key', '/etc/ssh/')
+            ]
+            assert file_handle_sshd_config.write.call_args_list == [
+                call('HostKey /etc/ssh/ssh_host_ecdsa_key')
+            ]
```
schaefi: The testing of the sshd config update seems incomplete to me. Actually what we do is:

1. add a newline to be safe
2. add HostKey entries

The unit test should outline this logic:

1. Did we really start with a newline
2. Did we correctly add HostKey entries, at least two entries should be tested

If I would write the test I would go with real testing files taken from test/data. The only thing that needs to be mocked would be the Defaults method to point to the test/data instead of the system. That way the test becomes almost real and reading from test/data is a safe operation for unit testing

test/unit/units/ssh_access_test.py
schaefi: github quoted this a bit unfortunate. The lines 30-37 are meant by this comment

jesusbv: There is no change in the authorized_keys testing in those lines, the glob returns the values for authorized_key (as before) and the values for host keys (new values for the new case)

test/unit/units/ssh_access_test.py
jesusbv: Ok, I opted for no using any file to avoid unnecessary testing files. I will update the test with a file.

test/unit/units/ssh_access_test.py
jesusbv: I reverted all the changes related to authorized_keys

schaefi: Thanks

schaefi:
@jesusbv Thanks looks good now, can you rebase the commits into one and then merge ? Thanks

schaefi:
Hope you don't mind I did the merge because I'd like to submit a new package to the devel space

PR #107, opened 2019-06-02
Provide detailed network information

### LINKED ISSUE #84: Provide detailed network interface/status in log file
During some issues with the network setup I realized that it is a bit difficult to find out the root cause for sub-optimal network setup while in the temporary system.

Maybe a simple "ip a" / "routes -n" and "cat /etc/resolv.conf" could be added to the /var/log/distro_migration.log after a successful network setup?

jesusbv:
After network.online target is done, we log
the network status, interfaces and relevant information.

Fixes #84

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -124,13 +124,15 @@ def main():
                 ).output
             )
         )
-        log.info(
-            'Network Bonding {0}{1}'.format(
-                os.linesep, Command.run(
-                    ['cat', '/proc/net/bonding/bond*'], raise_on_error=False
-                ).output
+        bonding_paths = Defaults.get_bonding_paths()
+        if os.path.exists(os.path.dirname(bonding_paths)):
+            log.info(
+                'Network Bonding {0}{1}'.format(
+                    os.linesep, Command.run(
+                        ['cat', bonding_paths], raise_on_error=False
+                    ).output
+                )
             )
-        )
         log.info('Running prepare service')
         system_mount = Fstab()
         system_mount.read(
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -24,6 +24,9 @@ from suse_migration_services.command import Command
 from suse_migration_services.fstab import Fstab
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import log
+from suse_migration_services.units.setup_host_network import (
+    log_network_details
+)
 
 from suse_migration_services.exceptions import (
     DistMigrationZypperMetaDataException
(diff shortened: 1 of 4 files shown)


suse_migration_services/units/prepare.py
```diff
@@ -102,6 +102,35 @@ def main():
...
+                ).output
+            )
+        )
+        log.info(
+            'Network Bonding {0}{1}'.format(
+                os.linesep, Command.run(
+                    ['cat', '/proc/net/bonding/bond*'], raise_on_error=False
+                ).output
+            )
+        )
```
schaefi: I don't think the bonding information are relevant. Also ```/proc/net/bonding``` only exists if bonding is active. So this would lead to an error in the log at any time no bond network is used. I would not log bonding data unless someone can tell me why this information is useful ?

jesusbv: It seems relevant information for Martin.
I did test it before the PR and yes, no bonding info in my case.

rjschwei: Given we provide the network information as proposed here, then the bonding information should be included if it exists.

schaefi: ok but this condition is missing:

```
cat /proc/net/bonding/bond*
cat: '/proc/net/bonding/bond*': No such file or directory
```

So the code should only run if /proc/net/bonding is present

Martin-Weiss: We have to make sure that i.e. 802.3ad bonding mode is active and bonding is setup correct. 
There are cases when you will not have a chance to communicate over the configured network in case the bonding mode is not setup properly. So for debugging "network does not work during live system" - this information would be very useful.. and based on the udev learning's maybe also the 70-persistent-net might be needed in the worst case....

suse_migration_services/units/prepare.py
```diff
@@ -102,6 +102,37 @@ def main():
...
+        log.info(
+            'DNS Resolver {0}{1}'.format(
+                os.linesep, Command.run(
+                    ['cat', '/etc/resolv.conf'], raise_on_error=False
+                ).output
+            )
+        )
+        bonding_paths = Defaults.get_bonding_paths()
+        if os.path.exists(os.path.dirname(bonding_paths)):
```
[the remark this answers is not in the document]
schaefi: Also the Defaults space is usually used for locations that are project specific and/or used several times. The location of the bonding interfaces is given by the kernel cannot be influenced by us and used only once. I think you don't need an extra method in the Defaults class for this.

test/unit/units/prepare_test.py
```diff
@@ -100,6 +110,30 @@ def test_main(
...
+                    'cat', '/etc/resolv.conf'
+                ],
+                raise_on_error=False
+            ),
+            call(
+                [
+                    'cat', '/proc/net/bonding/bond*'
+                ],
+                raise_on_error=False
+            ),
```
schaefi: I think it can be written in one line

```python
call(
    ['cat', '/proc/net/bonding/bond*'], raise_on_error=False
)
```

The 80 characters per line limit is not violated iirc

suse_migration_services/units/prepare.py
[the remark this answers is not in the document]
jesusbv: I will remove the method from Defaults. 
The os.path.exists gets an os.path.dirname of the glob pattern, I think it works.

test/unit/units/prepare_test.py
jesusbv: I tried to follow the same pattern that it was in the test already, that doesn't violate the limit either.
I will update it.

suse_migration_services/units/prepare.py
schaefi: ok Martin, thanks for the update

suse_migration_services/units/prepare.py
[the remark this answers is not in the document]
schaefi: Oh I see, dirname... ok overlooked that part. sorry

test/unit/units/prepare_test.py
schaefi: right, my bad... happy if you fix my stuff too :)

schaefi:
Looks great now :+1: Thanks

Issue #110, opened 2019-06-04
The mounting of filesystems from the system to be migrated needs to be more selective to only attempt to mount filesystems relevant to the migration


Martin-Weiss:
I have this in /etc/fstab:

`UUID=0b1b1a50-3fd8-45d3-9167-efc68926604d swap swap defaults 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 / btrfs defaults 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /boot/grub2/i386-pc btrfs subvol=@/boot/grub2/i386-pc 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /boot/grub2/x86_64-efi btrfs subvol=@/boot/grub2/x86_64-efi 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /home btrfs subvol=@/home 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /opt btrfs subvol=@/opt 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /srv btrfs subvol=@/srv 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /tmp btrfs subvol=@/tmp 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /usr/local btrfs subvol=@/usr/local 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /var/cache btrfs subvol=@/var/cache 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /var/crash btrfs subvol=@/var/crash 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /var/lib/libvirt/images btrfs subvol=@/var/lib/libvirt/images 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /var/lib/machines btrfs subvol=@/var/lib/machines 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /var/lib/mailman btrfs subvol=@/var/lib/mailman 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /var/lib/mariadb btrfs subvol=@/var/lib/mariadb 0 0
UUID=df286cfc-3da3-4361-a2a6-155489aa2035 /var/lib/mysql btrfs subvol=@/var/lib/mysql 0 0
[...]

PR #112, opened 2019-06-04
Ignore blank lines and comments when parsing fstab

tserong:
This fixes the commented / blank line issue mentioned in #110


schaefi:
nice thanks

PR #114, opened 2019-06-04
Fixed SSH identity copy

schaefi:
The SSH host identity is provided by the server credentials that
makes up the fingerprint of that host. However the copy of the
files should exclude the encrypted private key ssh_host_key which
can only be read on the origin host. In fact a copy of that file
to another host will lead to a segfault in sshd and kills the
connection without any further information.

### DIFF
--- a/suse_migration_services/units/ssh_keys.py
+++ b/suse_migration_services/units/ssh_keys.py
@@ -34,7 +34,8 @@ def main():
     """
     ssh_keys_glob_paths = Defaults.get_ssh_keys_paths()
     migration_ssh_file = Defaults.get_migration_ssh_file()
-    system_ssh_host_keys_glob_path = Defaults.get_system_ssh_host_keys_glob_path()
+    system_ssh_host_keys_glob_path = \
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/ssh_keys.py
```diff
@@ -57,22 +58,24 @@ def main():
         sshd_config_host_keys_entries = []
         log.info('Copying host ssh keys')
         for system_ssh_host_key in system_ssh_host_keys:
-            shutil.copy(system_ssh_host_key, '/etc/ssh/')
+            if not system_ssh_host_key.endswith('ssh_host_key'):
```
rjschwei: But host keys don't end with "ssh_host_key" keys are named according to the encryption algorithm being used such as "ssh_host_ecdsa_key".

I am confused

schaefi: There is one key that is encrypted and not PEM encoded named /etc/ssh/ssh_host_key. That one can't be copied

PR #115, opened 2019-06-04
- Only attempt to mount filesystems from the host that my be part of the "core system"

rjschwei:
+ Entries in fstab my include filesystems we do not care about for
    the migration, such as pointing to ISO files on local disk. Ignore such
    entries and only consider filesystems on devices

This addresses #110

### DIFF
--- a/suse_migration_services/fstab.py
+++ b/suse_migration_services/fstab.py
@@ -58,6 +58,9 @@ class Fstab(object):
                         device_path = ''.join(
                             ['/dev/disk/by-label/', device.split('=')[1]]
                         )
+                    elif device.startswith('PARTUUID'):
+                        device_path = ''.join(
+                            ['/dev/disk/by-by-partuuid/', device.split('=')[1]]
                     else:
                         device_path = device
                     self.fstab.append(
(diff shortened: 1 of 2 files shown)


rjschwei:
@schaefi @jesusbv this is one approach of dealing with the problem. Will add/fix tests if this is what we want to do. The other option is to filter by mount point, i.e. we could use:

```
system_mount_points = [ALL_MOUNT_POINTS_FROM_FHS]
for fstab_entry in fstab.get_devices():
    if fstab_entry.mountpoint not in system_mount_points:
        continue
```
The ALL_MOUNT_POINTS_FROM_FHS is defined by [1]. The list of mountpoints is rather lengthy. However the approach of using the FHS specification would make it easier to handle NFS mounts as we wouldn't have to handle "IP-ADDRESS:PATH" and "HOSTNAME:PATH", which are a bit more difficult to comprehend in the filtering loop if we ever care to handle NFS mounts.

Of course if we want to handle NFS mounts then we have a problem with the order, mounting happens before the network is up and we'd have to implement a "delayed mount" code path.

Anyway, primarily wanted to show what this could look like to avoid the issue brought in #110 which I re-opened. While I agree that we should expect a "clean" fstab file, the reality is that we are not going to get it. So IMHO we can avoid the issue by being more selective about what we mount.


[1] http://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html

schaefi:
per conversation on #110 please see my suggestion in #116

rjschwei:
Superseded by #116

PR #117, opened 2019-06-05
Update documentation

### LINKED ISSUE #99: Document zypper dup support
In reference to #98 we need to add this feature to the documentation once #82 is merged

schaefi:
Added section about possible config file values including
the explanation for the different zypper install modes.
This Fixes #99

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -28,7 +28,14 @@ Latest Version: {Latest_Version}
 Code available: {Repo}
 
 == Concept
-The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration tool is the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
+The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration plugin provides the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
+
+The upgrade is done via the network using the Zypper migration plugin which
+sends a request to the SUSE Customer Center (SCC), asking for an upgrade path.
+The request response contains the list of repositories required to upgrade the
+system. This requires the system to upgrade to be a SLES registered system.
+The upgrade system also supports an upgrade mode that works with non
+registered systems. For details see <<Customize the Migration Process>>
 
 The upgrade to a new major versions requires the system to be offline during the upgrade to avoid system inconsistencies that may leave the system in a state that does not allow recovery. This behavior is implemented using a Live Migration Image.
 
@@ -37,7 +44,7 @@ The distribution migration system provides the live image and a component that m
 1. detect the system to be upgraded
 2. mount the necessary file systems
 3. setup the network to match the network configuration of the system to be upgraded
-4. prepare the package manager for the upgrade task
+4. prepare the package manager for the upgrade task 
 5. upgrade the system
 6. update the boot loader configuration
 7. unmount all mounted file systems
@@ -74,13 +81,19 @@ tux > ssh INSTANCE_USER@IP_OF_INSTANCE 'touch /etc/sle-migration-service.yml'
 
 The custom config file allows for the following settings:
 
-Enable Debug Mode::
-If enabled, prevents the upgrade system from rewinding the setup
-steps and rebooting due to a failed upgrade, allowing the issue to
-be debugged.
+Control Zypper Installation Mode::
+If the upgrade process is used on systems that are not SLES registered
+or for which the SUSE Customer Center has no upgrade path, it's required to
+switch off the use of the migration plugin.
 
 [listing]
-debug: true|false
+use_zypper_migration_plugin: true|false
+
+[NOTE]
+If the migration plugin is not used, the upgrade process just calls
+`zypper dup` and expects all required repositories to be setup correctly.
+This means prior to the upgrade the setup of the repositories must
+be performed.
 
 Specify Migration Product::
 With the SLE15-Migration package a predefined product setup is
@@ -94,26 +107,17 @@ with this setting. Following product names can be set:
 migration_product: SLES/15/x86_64
 
 [NOTE]
-If the migration plugin is not used, the value set for the target
-product will not have any effect.
+If `use_zypper_migration_plugin: false` is set, the value for the
+target product will not be effective because it's used only in
+combination with the Zypper migration plugin.
 
-Control Zypper Installation Mode::
-By default the upgrade system uses a plugin named: `migration`. This
-plugin sends a request to the SUSE Customer Center, asking for an upgrade
-path. The request response contains the list of repositories required to
-upgrade the system. This process only works for SLES registered systems.
-If the upgrade process is used on systems that are not registered or for
-which the SUSE Customer Center has no upgrade path, it's required to
-switch off the use of the migration plugin.
+Enable Debug Mode::
+If enabled, prevents the upgrade system from rewinding the setup
+steps and rebooting due to a failed upgrade, allowing the issue to
+be debugged.
 
 [listing]
-use_zypper_migration_plugin: true|false
-
-[NOTE]
-If the migration plugin is not used, the upgrade process just calls
-`zypper dup` and expects all required repositories to be setup correctly.
-This means prior to the upgrade the setup of the repositories must
-be performed.
+debug: true|false
 
 == Run the Migration
 
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -34,7 +34,7 @@ The upgrade is done via the network using the Zypper migration plugin which
 sends a request to the SUSE Customer Center (SCC), asking for an upgrade path.
 The request response contains the list of repositories required to upgrade the
 system. This requires the system to upgrade to be a SLES registered system.
-The upgrade system also supports an upgrade mode that works with non
+The upgrade system also supports an upgrade mode that works with non SLES
 registered systems. For details see <<Customize the Migration Process>>
 
 The upgrade to a new major versions requires the system to be offline during the upgrade to avoid system inconsistencies that may leave the system in a state that does not allow recovery. This behavior is implemented using a Live Migration Image.
@@ -44,7 +44,7 @@ The distribution migration system provides the live image and a component that m
 1. detect the system to be upgraded
 2. mount the necessary file systems
 3. setup the network to match the network configuration of the system to be upgraded
-4. prepare the package manager for the upgrade task 
+4. prepare the package manager for the upgrade task
 5. upgrade the system
 6. update the boot loader configuration
 7. unmount all mounted file systems
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -28,14 +28,17 @@ Latest Version: {Latest_Version}
 Code available: {Repo}
 
 == Concept
-The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration plugin provides the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
+The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration workflow provides the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
 
-The upgrade is done via the network using the Zypper migration plugin which
-sends a request to the SUSE Customer Center (SCC), asking for an upgrade path.
+The upgrade is done via the network using the Zypper migration workflow which
+sends a request to the repository server, asking for an upgrade path.
+SUSE supported repository servers are the SUSE Customer Center (SCC), the
+Repository Management Tool (RMT), and the Subscription Management Tool (SMT).
 The request response contains the list of repositories required to upgrade the
-system. This requires the system to upgrade to be a SLES registered system.
-The upgrade system also supports an upgrade mode that works with non SLES
-registered systems. For details see <<Customize the Migration Process>>
+system. This requires the system to upgrade to be a registered system.
+The upgrade system also supports an upgrade mode that works with systems not
+registered to a repository service. For details
+see <<Customize the Migration Process>>
 
 The upgrade to a new major versions requires the system to be offline during the upgrade to avoid system inconsistencies that may leave the system in a state that does not allow recovery. This behavior is implemented using a Live Migration Image.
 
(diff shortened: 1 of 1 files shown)


doc/adoc/user_guide.adoc
```diff
@@ -62,9 +62,62 @@ To install the distribution migration system run:
...
+If enabled, prevents the upgrade system from rewinding the setup
+steps and rebooting due to a failed upgrade, allowing the issue to
+be debugged.
+
+[listing]
+debug: true|false
+
+Specify Migration Product::
+With the SLE15-Migration package a predefined product setup is
+already applied. The value for the target product can be overwritten
```
jesusbv: should we not add "......for the target product in that setup can be overwritten..." ?

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +62,62 @@ To install the distribution migration system run:
...
+* SLES/15.1/x86_64
+
+[listing]
+migration_product: SLES/15/x86_64
+
+[NOTE]
+If the migration plugin is not used, the value set for the target
+product will not have any effect.
+
+Control Zypper Installation Mode::
```
jesusbv: s/::/:

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +62,62 @@ To install the distribution migration system run:
...
+By default the upgrade system uses a plugin named: `migration`. This
```
jesusbv: s/:/

doc/adoc/user_guide.adoc
tomschr: It's a  [description list](https://asciidoctor.org/docs/user-manual/#description-list). For the terms, you need the colons.

doc/adoc/user_guide.adoc
tomschr:
```suggestion
By default, the upgrade system uses a plugin named: `migration`. This
```

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +62,62 @@ To install the distribution migration system run:
...
+switch off the use of the migration plugin.
+
+[listing]
+use_zypper_migration_plugin: true|false
+
+[NOTE]
+If the migration plugin is not used, the upgrade process just calls
+`zypper dup` and expects all required repositories to be setup correctly.
+This means prior to the upgrade the setup of the repositories must
+be performed.
```
tomschr: I think, the order is a bit off. 

```
If the migration plugin is not used, the setup of the repositories must
be performed. Once done, the upgrade process just calls `zypper dup` and expects all required
repositories to be setup correctly.

File: doc/adoc/user_guide.adoc [hunk 5]
```diff
@@ -62,9 +62,62 @@ To install the distribution migration system run:
...
+with this setting. Following product names can be set:
+
+* SLES/15/x86_64
+* SLES/15.1/x86_64
+
+[listing]
+migration_product: SLES/15/x86_64
+
+[NOTE]
+If the migration plugin is not used, the value set for the target
```
rjschwei: At this point the reader knows nothing about plugins and cannot relate the information presented to any other concept previously discussed. If we talk about "plugins" here we need to set the context earlier in the document to describe the difference between plain "zypper dup" and "zypper migration".

The other option is to re-write this in a way that simply addresses, in a brief way the differences between "zypper dup" and "zypper migration" and when might want to use one vs. the other. Then state that this setting allows the user to switch between the two methods.

doc/adoc/user_guide.adoc
schaefi: right, asciidoc style

schaefi:
Thanks for the input. I have re-structured the docs according to what has been said.

You can build and read the docs if you have daps installed:

```
daps -d DC-distribution-migration-system html --single
```

doc/adoc/user_guide.adoc
schaefi: I did not understand this part ?

doc/adoc/user_guide.adoc
schaefi: I tried that now, hope it's more self explaining now

doc/adoc/user_guide.adoc
schaefi: The point is that this repo setup has to be done prior running the migration, that's why the reason for this was mentioned first. But I follow your experience here and changed it :) Thanks

schaefi:
@tomschr regarding your suggestion on the process I'm not sure what you mean. Actually the document structures the process in sections you can go through from top to down. The section with "Customize the Migration Process" is however optional. All in there can be done but doesn't have to be done. I did not mark this in any form because other sections on the same level don't do this either, e.g "Recommendation".

So from my point of view the structure of the document is ok, actually when Frank reviewed the current state there was no concern in this regard, which is also the reason why the document is as it is right now.

So can you clarify what the actual problem is ?

Thanks

doc/adoc/user_guide.adoc
jesusbv: @tomschr thanks ! I missed the list, thinking it was a typo

doc/adoc/user_guide.adoc
jesusbv: The predefined product setup has a target product with a default value.
That value can be overwritten, as you mentioned in that part of the docs, I wanted to
be more explicit in case it was that obvious for other readers.

schaefi: I think I got it

doc/adoc/user_guide.adoc
```diff
@@ -28,7 +28,14 @@ Latest Version: {Latest_Version}
...
-The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration tool is the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
+The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration plugin provides the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
```
rjschwei: I think we can drop "plugin". To the subject at hand it is really immaterial whether the "migration" command for `zypper` is implemented as a plugin or part of the `zypper` C++ code base. 

s/Zypper migration plugin/zypper migration workflow/

doc/adoc/user_guide.adoc
```diff
@@ -28,7 +28,14 @@ Latest Version: {Latest_Version}
...
+
+The upgrade is done via the network using the Zypper migration plugin which
```
rjschwei: s/plugin/workflow/

doc/adoc/user_guide.adoc
```diff
@@ -28,7 +28,14 @@ Latest Version: {Latest_Version}
...
+sends a request to the SUSE Customer Center (SCC), asking for an upgrade path.
```
rjschwei: Hmm we should probably not limit the target where this request goes to be SCC only, that gives an incomplete impression. Maybe use

"...sends a request to the repository server, asking for an upgrade path. SUSE supported repository servers are the SUSE Customer Center (SCC), the Repository Management Tool (RMT), and the Subscription Management Tool (SMT)."

doc/adoc/user_guide.adoc
```diff
@@ -28,7 +28,14 @@ Latest Version: {Latest_Version}
...
+The request response contains the list of repositories required to upgrade the
+system. This requires the system to upgrade to be a SLES registered system.
```
rjschwei: s/SLES//

doc/adoc/user_guide.adoc
```diff
@@ -28,7 +28,14 @@ Latest Version: {Latest_Version}
...
+The upgrade system also supports an upgrade mode that works with non SLES
```
rjschwei: s/The upgrade system /The distribution migration workflow/

s/with non SLES registered systems/with systems not registered to a repository service/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
 [listing]
 tux > sudo zypper in SLE15-Migration suse-migration-sle15-activation
 
+== Customize the Migration Process
+
+The migration system reads a custom config file from the system to upgrade.
```
rjschwei: s/config/configuration/
s/to upgrade/to be upgraded/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+The contents of that file allows to influence the behavior of the upgrade
```
rjschwei: s/contents/content/
s/that/this/
s/allows to influence/modifies/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+process. Prior to the start of the upgrade process, create the following
+file:
+
+[listing]
+tux > ssh INSTANCE_USER@IP_OF_INSTANCE 'touch /etc/sle-migration-service.yml'
+
+The custom config file allows for the following settings:
```
rjschwei: s/allows for/supports/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+
+Control Zypper Installation Mode::
+If the upgrade process is used on systems that are not SLES registered
```
rjschwei: s/SLES//

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+or for which the SUSE Customer Center has no upgrade path, it's required to
```
rjschwei: s/SUSE Customer Center/repository server/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+switch off the use of the migration plugin.
```
rjschwei: s/switch/turn/
s/plugin/workflow/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+
+[listing]
+use_zypper_migration_plugin: true|false
```
rjschwei: I know this is late, I missed this in one of the commits, can we rename the option to be just `use_zypper_migration`?

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+
+[NOTE]
+If the migration plugin is not used, the setup of the repositories must
```
rjschwei: s/plugin/workflow/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+be performed manually. Once done, the upgrade process just calls `zypper dup`
+and expects all required repositories to be setup correctly.
+
+Specify Migration Product::
+With the SLE15-Migration package a predefined product setup is
+already applied. That setting can be overwritten with one of the
+following product names:
```
rjschwei: s/names/specifications/

The triplet is not the product name, it is a synonym and we should not conflate the two

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+
+* SLES/15/x86_64
+* SLES/15.1/x86_64
+
+[listing]
+migration_product: SLES/15/x86_64
+
+[NOTE]
+If `use_zypper_migration_plugin: false` is set, the value for the
```
rjschwei: drop `_plugin` please, see above.

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +69,58 @@ To install the distribution migration system run:
...
+target product will not be effective because it's used only in
+combination with the Zypper migration plugin.
```
rjschwei: s/plugin/workflow/

rjschwei:
Sorry for the flood of comments when I started I didn't expect that I'd have so many requests, I should have just sent a PR.

schaefi:
No problem this is good. I'll adapt the docs. The part with renaming the config option I'd like to handle in a followup request though. Could you open an issue for the rename of the option and the motivation behind it ? That would be great, thanks

Starting to update the rest now...

doc/adoc/user_guide.adoc
schaefi: yes we can as it's all still in development. But let's handle it in a followup request please.
Could you create an issue for it, then we don't forget. Thanks

doc/adoc/user_guide.adoc
schaefi: yep, after I updated the docs according to your comments there are only the two places left that name the option. Let's handle that change which also requires a code path to change in a new PR. Also see my comment above. Thanks

tomschr:
@schaefi 
> The section with "Customize the Migration Process" is however optional. All in there can be done but doesn't have to be done. I did not mark this in any form because other sections on the same level don't do this either, e.g "Recommendation".

I didn't know. When I looked at it---admittedly in isolation from the complete document---I wasn't aware of this fact. Therefore, I really think you should tell that your readers. A small sentence should be enough.

Experience tells, that readers do not read a document from cover to cover, but jump right into the section that they think are interesting. :wink:


> So from my point of view the structure of the document is ok, actually when Frank reviewed the current state there was no concern in this regard, which is also the reason why the document is as it is right now.

Ok, different people have different opinions. :wink:  It's totally fine when other people disagree with my conclusion. I just state my opinion. :smile: 

I think, I see now where the confusion comes from. You see the sections as kind of "steps" which the user needs to complete. I see the section titles mainly as "hint" if it is task-based or a reference-based content. So my thinking was this: "Oh, a verb ("customize"), the user has to do some actions. What steps do I need to take?" After looking at the section, I found no steps at all. :wink: That triggered some actions... :laughing: 

After reading your comments, I think you can leave it as it is. However, I would really suggest to state that the content in this section is optional.

In another PR, you could think about this:

* Is the main title really inviting? IMHO, the title "SUSE Distribution Migration System" lacks a verb ("Using"?) as the content contains some user actions and not reference material.
* Change the section title "Debugging the Migration" to "Debug the Migration" to be consistent with the other section titles.

> So can you clarify what the actual problem is ?

Hope it clarifies some points? :-)

tomschr:
@rjschwei 

> Sorry for the flood of comments when I started I didn't expect that I'd have so many requests, I should have just sent a PR.

You probably know that already, but just in case: you can send the review all at once. Have you heard about [suggestions](https://help.github.com/en/articles/reviewing-proposed-changes-in-a-pull-request) in GitHub? Look for step 6. I use it a lot and it eases the pain for reviewer and implementer. :wink:

rjschwei:
Issue for the option rename #118

doc/adoc/user_guide.adoc
```diff
@@ -28,7 +28,17 @@ Latest Version: {Latest_Version}
...
+The Distribution Migration System provides an upgrade path for an installed SUSE Linux Enterprise system from one major version to another, for example, from SUSE Linux Enterprise Server 12 SP4 to SUSE Linux Enterprise Server 15. For a service pack upgrade from one Service Pack (SP) to another within a given major version release series the existing Zypper migration workflow provides the supported upgrade path. The distribution migration system provides the ability to upgrade across major distributions without the need to use the next major version installation media to perform the system upgrade.
+
+The upgrade is done via the network using the Zypper migration workflow which
+sends a request to the repository server, asking for an upgrade path.
+SUSE supported repository servers are the SUSE Customer Center (SCC), the
+Repository Management Tool (RMT), and the Subscription Management Tool (SMT).
+The request response contains the list of repositories required to upgrade the
+system. This requires the system to upgrade to be a registered system.
```
rjschwei: s/This requires the system to upgrade to be a registered system./This requires the system to be upgraded to be registered/

doc/adoc/user_guide.adoc
```diff
@@ -28,7 +28,17 @@ Latest Version: {Latest_Version}
...
+The upgrade system also supports an upgrade mode that works with systems not
```
rjschwei: s/upgrade system/migration implementation/

This is to avoid confusion between the system we are upgrading and the system that is running the upgrade.

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +72,58 @@ To install the distribution migration system run:
...
+Control Zypper Installation Mode::
+If the upgrade process is used on systems that are not registered
+or for which the repository server has no upgrade path, it's required to
+switch off the use of the migration workflow.
+
+[listing]
+use_zypper_migration_plugin: true|false
+
+[NOTE]
+If the migration workflow is not used, the setup of the repositories must
```
rjschwei: s/If/The use of the migration workflow is the default behavior. If/

doc/adoc/user_guide.adoc
```diff
@@ -62,9 +72,58 @@ To install the distribution migration system run:
...
+be performed manually. Once done, the upgrade process just calls `zypper dup`
```
rjschwei: s/just//

rjschwei:
> @rjschwei
> 
> > Sorry for the flood of comments when I started I didn't expect that I'd have so many requests, I should have just sent a PR.
> 
> You probably know that already, but just in case: you can send the review all at once. Have you heard about [suggestions](https://help.github.com/en/articles/reviewing-proposed-changes-in-a-pull-request) in GitHub? Look for step 6. I use it a lot and it eases the pain for reviewer and implementer.

@tomschr thanks for the hint and yes. But you know how it goes you start with one, "oh let me just comment on this" and then there's another and another ....

schaefi:
Obviously I'm not able to write good docs with or without the suggestions mechanism in place ;) The few lines of text for review here has caused 10 times more the effort to come to an acceptable state. I'll integrate the other suggestions now... stay tuned

rjschwei:
> Obviously I'm not able to write good docs with or without the suggestions mechanism in place ;) The few lines of text for review here has caused 10 times more the effort to come to an acceptable state. I'll integrate the other suggestions now... stay tuned

I think it was a great start. Especially for documentation it is a lot easier to suggest changes than to come up with the original.

schaefi:
@tomschr ok I have changed the structure a bit regarding your suggestions. I agree things that are optionally should be identified as such. I also realized that what we set as required is only required under certain circumstances, this should also be clear.

Would be great if you can look at the document as a whole, daps can easily built a single-html page to look at.

I think it should be less confusing now.

Please advise. Thanks

tomschr:
@schaefi 
> Obviously I'm not able to write good docs with or without the suggestions mechanism in place ;) 

Hey, don't worry, you **are** able to write good docs! It wasn't my intention to give you a bad feeling. :)

> ok I have changed the structure a bit regarding your suggestions. [...] I also realized that what we set as required is only required under certain circumstances, this should also be clear.

Perfect! I like the new "Upgrade Prerequisites" section. Much clearer now. :+1: 

> Would be great if you can look at the document as a whole, [...] I think it should be less confusing now.

Looked as a whole document, read it, like it! :+1: 

Great job, Marcus!

schaefi:
ok thanks :)

PR #121, opened 2019-06-17
Added post mount service

### LINKED ISSUE #100: Allow preserving custom files e.g udev rules
We have situations where we have lots of NICs in a server and to ensure the device names are identical on all servers the /etc/udev/rules.d/70-persistent-net.rules is adjusted.

In case we would "just" map the /etc/sysconfig/network/ things into the live system but not the udev rules, too - the ifcfg-<device> files would not match anymore to the "real" interfaces.

So now the question is "do we use the adjusted udev rules of the installed system also while running the Live system?"

schaefi:
The post mount service is used to preserve data from the
system to be migrated into the live migration system directly
after the mounting of the system to be migrated. It is is
also responsible for making that data effective in the live
migration system. The current implementation allows for
preserving udev rules file and make them active.
This Fixes #100

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -155,6 +155,12 @@ preserve:
 ￼   - /etc/udev/rules.d/a.rules
 ￼   - /etc/udev/rules.d/b.rules
 
+[NOTE]
+udev rules that require custom drivers will not have the desired effect
+as the migration system will not include these drivers and therefore
+execution of those rules will fail. Rules with such properties should
+not be listed.
+
 Enable Debug Mode::
 If enabled, prevents the upgrade system from rewinding the setup
 steps and rebooting due to a failed upgrade, allowing the issue to
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -144,6 +144,17 @@ If `use_zypper_migration: false` is set, the value for the
 `migration_product` will not be effective because it's used only in
 combination with the Zypper migration workflow.
 
+Preserve System Data::
+Preserve custom data file(s) e.g udev rules from the system
+to be migrated into the upgrade live system and make sure
+they will become effective.
+
+[listing]
+preserve:
+  rules:
+￼   - /etc/udev/rules.d/a.rules
+￼   - /etc/udev/rules.d/b.rules
+
 Enable Debug Mode::
 If enabled, prevents the upgrade system from rewinding the setup
 steps and rebooting due to a failed upgrade, allowing the issue to
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -63,6 +63,20 @@ class MigrationConfig(object):
 
         return migration_product
 
+    def get_preserve_udev_rules_list(self):
+        """
+        Returns list of udev rule file(s)
+
+        The returned file list is used to make those rule files
+        available to the migration live system such that the
+        same behavior the rules apply on the system to be
+        upgraded also applies to the live migration system
+        while the upgrade runs
+        """
+        preserve_data = self.config_data.get('preserve')
+        if preserve_data:
+            return preserve_data.get('rules')
+
     def update_migration_config_file(self):
         """
         Update the default migration configuration with custom values
--- a/suse_migration_services/units/post_mount_system.py
+++ b/suse_migration_services/units/post_mount_system.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2019 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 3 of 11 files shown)


suse_migration_services/units/post_mount_system.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2018 SUSE Linux LLC.  All rights reserved.
```
jesusbv: s/2018/2019

suse_migration_services/migration_config.py
```diff
@@ -63,6 +63,20 @@ def get_migration_product(self):
...
+
+        The returned file list is used to make those rule files
+        available to the migration live system such that the
+        same behavior the rules apply on the system to be
+        upgraded also applies to the live migration system
+        while the upgrade runs
+        """
+        preserve_data = self.config_data.get('preserve')
+        if preserve_data:
+            return preserve_data.get('rules')
```
jesusbv: It would be safer to have a default in case of typo/error in the custom file.

jesusbv:
The documentation would need an update

suse_migration_services/units/post_mount_system.py
schaefi: I make this mistake again and again

schaefi:
yes doc update needed. I need to add that, thanks

doc/adoc/user_guide.adoc
```diff
@@ -144,6 +144,17 @@ If `use_zypper_migration: false` is set, the value for the
 `migration_product` will not be effective because it's used only in
 combination with the Zypper migration workflow.
 
+Preserve System Data::
+Preserve custom data file(s) e.g udev rules from the system
+to be migrated into the upgrade live system and make sure
+they will become effective.
```
rjschwei: I think we should point out a general limitation with this. Such as

"""
udev rules that require custom drivers will not have the desired effect as the migration system will not include these drivers and therefore execution of those rules will fail. Rules with such properties should not be listed.
"""

schaefi: yes very good point, thanks

schaefi:
@tserong it failed in Devel:PubCloud because someone had the brilliant idea to update python urllib which broke basically everything. I did not see a reason to keep a broken package in that project and just deleted it, such that urllib is taken from  project="SUSE:SLE-15-SP1:GA". Sorry if that sounds harsh but just updating packages without checking if something else broke is not a good idea.

After that I've update suse-migration-services and the image build should be at your hand soon

tserong:
Thanks @schaefi, that's an entirely reasonable decision :-)

I did a little test, and ran into a problem...  On the host system (a VM), I had eth0, eth1, eth2 and eth3, and then edited `/etc/udev/rules.d/70-persistent-net.rules` to rename these to eth0, eth11, eth12 and eth13, i.e. I kept eth0 as-is, and renamed the other interfaces (I realise that on real hardware, it would be more interesting to get the interface names to match specific devices, rather than just screwing around with interface names as I am doing, but I hope this still counts as a valid test):

```
# cat /etc/udev/rules.d/70-persistent-net.rules 
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="virtio-pci", ATTR{dev_id}=="0x0", KERNELS=="0000:00:08.0", ATTR{type}=="1", KERNEL=="eth*", NAME="eth13"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="virtio-pci", ATTR{dev_id}=="0x0", KERNELS=="0000:00:06.0", ATTR{type}=="1", KERNEL=="eth*", NAME="eth11"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="virtio-pci", ATTR{dev_id}=="0x0", KERNELS=="0000:00:07.0", ATTR{type}=="1", KERNEL=="eth*", NAME="eth12"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="virtio-pci", ATTR{dev_id}=="0x0", KERNELS=="0000:00:05.0", ATTR{type}=="1", KERNEL=="eth*", NAME="eth0"
```

I also tweaked `/etc/sysconfig/network/ifcfg-*` to match the new names.

If I reboot the host system, the new names come up as expected, e.g.:

```
# ip a|awk '/eth[0-9]+:/ {print $2}'
eth0:
eth11:
eth12:
eth13:
```

If I reboot into the migration system, with `/etc/sle-migration-service.yml` including the appropriate preserve rules line, the `/etc/udev/rules.d/70-persistent-net.rules` file in the migration system *is* copied from the host, so the correct file is preserved...  But, the system doesn't recognise the different interface names:

```
# ip a|awk '/eth[0-9]+:/ {print $2}'
eth0:
eth1:
eth2:
eth3:
```

So... Does that mean the `udevadm` invocation isn't doing the right thing? Or maybe something is happening in the wrong order? Or, maybe my test is invalid? :-)

schaefi:
@tserong thanks for testing this. Sounds all reasonable to me, so your test is surely good and correct.

We need more information. Could you provide the log file /var/log/distro_migration.log ? It would also be very interesting to debug the migration live system. So can you enable the debug switch

```
debug: true
```

in sle-migration-service.yml. That will keep you in the live system and we can do further debugging.
For example you can check if the rule file really got copied. If we are sure it is present then check

```
udevadm control --reload
udevadm trigger --type=subsystems --acction=add
```

That should have added and renamed the interfaces. If not let's check why.

It could well be that I did a mistake or missed a step, udev is a beast

Could we followup the discussion by e-mail. It's because discussing on a merged PR is a bit cumbersome. Alternatively we can re-open the connected issue

Whatever works best for you

Thanks

schaefi:
Maybe you can also allow to let me ssh into your live system once the debug switch was enabled

tserong:
Thanks, let's do email.

PR #123, opened 2019-06-26
Fix migration package name in the documentation

### LINKED ISSUE #122: Wrong package name in docs
Recent feedback has pointed out wrong named package in the docs
```
tux > sudo zypper in SLE15-Migration 
```

That led to a confusing installation with an error message, besides installed the right packages.

jesusbv:
Avoid confusion and possible error due to wrong package name.

Fix #122


schaefi:
Sorry I was too fast. The name "SLE15-Migration" appears several times. All places should be fixed

PR #124, opened 2019-07-05
Fixup udev reload on rule changes

schaefi:
The current restart sequence is missing the actual set/change
of device nodes. This also prevented network interface names
to be added/renamed.


tserong:
Just to confirm, I've tested this on my VM setup here (as described in https://github.com/SUSE/suse-migration-services/pull/121#issuecomment-507644845) and it works fine. Thanks!

PR #125, opened 2019-07-16
Avoid bootloader to run the migration

### LINKED ISSUE #108: Use kexec instead of grub to boot into live migration system
Problem Description
================

The current way to boot into the live migration system and therefore start the migration is to reboot the machine and boot a pre-configured loopback grub menu entry. That menu entry is created through an activation script we provide as a package

The problem with this approach is that it does not work on all architectures. Only plain grub supported architectures can boot the migration system. Also in clouds like the Oracle cloud which boots machines through iSCSI via a network boot infrastructure and no direct requirement for a bootloader inside of the system will not be able to boot the live migration.

Alternative Solution
===============

Due to the bootloader restriction on the current architecture my idea was to get rid of the bootloader and directly boot the live migration system via ```kexec```. The activation script that usually creates a grub entry needs to be rewritten in a way that it extracts the kernel and initrd from the live migration iso image and just kexec's with the appropriate parameters. Instead of a reboot the user would need to call an ```activation_program``` we will provide with the activation package

That approach will also overcome the restriction of the bootloader to read from e.g encrypted root partitions.

schaefi:
Instead of a reboot the customer should run the migration
by calling run_migration. This commit adds a service utility
to run the migration. The concept is based on kexec and
avoids the modification to the bootloader setup. This allows
more flexibility for clouds that runs instances not directly
through a bootloader and also avoids infinite boot/reboot
cycles in case of an early error of the migration mount
service. This Fixes #108


schaefi:
Please note this is a change in behavior and will lead to a major version upgrade on merge.
Also there will be no extra activation package anymore. So overall this makes the startup nicer and less error prune

tools/run_migration
```diff
@@ -0,0 +1,115 @@
...
+    # shellcheck disable=SC2181
+    if [ ! $? = 0 ];then
+        echo "Migration run failed: see ${logfile} for details"
+    fi
+    umount /mnt &>/dev/null
+}
+
+function get_migration_image {
+    # """
+    # Search migration ISO file
```
rjschwei: "Search for ..."

tools/run_migration
```diff
@@ -0,0 +1,115 @@
...
+    # """
+    echo /usr/share/migration-image/*-Migration.*.iso
+}
+
+function get_system_root_device {
+    # """
+    # Lookup running system root device
```
rjschwei: "Looking for..."

tools/run_migration
schaefi: I should take part in an English class. We had that at suse years before and I think a refresh would be good :) Thanks

PR #126, opened 2019-07-17
Allow two ways to start the migration

schaefi:
Partially revert the no bootloader based startup
    
After testing the kexec based implementation to start the
migration live image, we found that kexec does not work in
all cloud providers. Especially those using Xen like AWS
do not work. Thus it's required to keep the alternative
bootloader based startup sequence

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,7 @@
 # Manifest describing source package contents
 
+prune grub.d
+
 graft tools
 graft suse_migration_services/units
 graft package
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,15 @@ build: check test
 	# provide rpm rpmlintrc
 	cp package/suse-migration-services-rpmlintrc dist
 
+sle15_activation: check
+	rm -f dist/*
+	git log grub.d | helper/changelog_generator |\
+		helper/changelog_descending > \
+		dist/suse-migration-sle15-activation.changes
+	cp package/suse-migration-sle15-activation-spec-template \
+		dist/suse-migration-sle15-activation.spec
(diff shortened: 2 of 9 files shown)


schaefi:
Unfortunately I had to revert parts of the implementation that avoids the bootloader. We have to keep the old concept together with the new one. kexec is not supported on all virtualization platforms. Especially the Xen based sucks. So this impacts Amazon, unfortunately. The good news is, people seem to work on it and maybe Amazon updates their Xen stack such that at some point kexec works there too.

So we can only document the primary way through kexec and suggest the alternative through the bootloader. Luckily both is possible and can co-exist with our concept

Martin-Weiss:
Even though the current focus in this project is not real hardware - but also on real hardware environments I got feedback that kexec is not working there in all cases. So keeping the old and the new way maybe with a configuration option would be nice.

schaefi:
> Even though the current focus in this project is not real hardware - but also on real hardware
> environments I got feedback that kexec is not working there in all cases. 

You are right, I was too enthusiastic that kexec will be able to complete replace the boot through a bootloader method.

> So keeping the old and the new way maybe with a configuration option would be nice.

Yes and it is done that way now. However you don't need a configuration option. Every SLES*-Migration
package is now packaged with the ```run_migration``` tool. From there you can go and give the kexec way a try and just call run_migration.

Alternatively you can install the suse-migration-sle*-activation package which on install will create the bootloader entry as you know and just reboot. Which method you choose is a decision to be made by the customer

Unfortunately there is no safe way to pre-detect if kexec will work or not. If you run in a Xen environment I know it does not work. From the public cloud side this means on Amazon we have to reboot

schaefi:
Also the documentation has been updated in this regard.

PR #127, opened 2019-07-18
- User doc

rjschwei:
+ Clarification around the choices of invoking the migration process using
    reboot or kexec


schaefi:
Thanks much, this really improves it. Sorry I had to continue with the former PR to get the image builds done for real testing in the clouds, so I took the liberty to merge prior to your review.

Thanks for this one

PR #130, opened 2019-07-26
Detect baseproduct on system to be migrated

### LINKED ISSUE #129: Auto detect baseproduct to migrate

jesusbv:
By default, the migration will target SLES 15,
however this may not always be the case, e.g. SLES SAP.

Fixes #129

### DIFF
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -18,6 +18,7 @@
 import yaml
 import os
 from textwrap import dedent
+from xml.etree.ElementTree import ElementTree
 
 # project
 from suse_migration_services.defaults import Defaults
@@ -96,6 +97,16 @@ class MigrationConfig(object):
                     self.migration_config_file, self.migration_custom_file
                 ).lstrip()
             )
+        else:  # auto-detect baseproduct
+            baseproduct_xml_file = self._read_baseproduct()
+            baseproduct_names = baseproduct_xml_file.findall('name')
+            for name in baseproduct_names:
+                default_migration_target = self.config_data.get('migration_product').split('/')
+                migration_product = '/'.join(
+                    [name.text, default_migration_target[0], default_migration_target[1]]
+                )
+                self.config_data['migration_product'] = migration_product
+                self._write_config_file()
 
     def is_debug_requested(self):
         return self.config_data.get('debug', False)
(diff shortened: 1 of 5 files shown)


jesusbv:
The `migration_product` info has changed, as the product now should not be set by default.

Removing it from `migration_product` is an option, or change the structure to

```yaml
version: 15
arch: 'x86_64'
```
Thoughts ?

suse_migration_services/migration_config.py
```diff
@@ -96,6 +97,16 @@ def update_migration_config_file(self):
...
+        else:  # auto-detect baseproduct
+            baseproduct_xml_file = self._read_baseproduct()
+            baseproduct_names = baseproduct_xml_file.findall('name')
+            for name in baseproduct_names:
+                default_migration_target = self.config_data.get('migration_product').split('/')
+                migration_product = '/'.join(
+                    [name.text, default_migration_target[0], default_migration_target[1]]
+                )
+                self.config_data['migration_product'] = migration_product
+                self._write_config_file()
```
schaefi: So this code reads the file behind the baseproduct link, reads in the name tag from the XML and adapts the value we have set in our image for the migration_product that gets passed to the migration plugin for zypper. All of this is only done if no custom migration config file exists. The place for this code is very very misleading. The scope of the MigrationConfig config class and its single responsibility is to handle the migration config file nothing more. This is one of the core rules when writing object oriented and I think we should follow this because it makes sense.

So what you need is code to read the name tag from the base product. If you take a look at units/product_setup.py

you will see that we already have code to handle .prod files. We even have code to detect the base product file. My suggestion is to close this PR and come up with a refactoring that moves any code that handles .prod files into its own class. e.g 

```python
class SUSEProduct(object):
    ...
```

Move all code we already have to the new class and just add the little extension to read the name tag
This should be imho the first PR. Let's review that one and then continue

suse_migration_services/migration_config.py
```diff
@@ -106,3 +117,11 @@ def is_zypper_migration_plugin_requested(self):
...
         with open(self.migration_config_file, 'w') as config:
             yaml.dump(self.config_data, config, default_flow_style=False)
+
+    def _read_baseproduct(self):
+        xml = ElementTree()
+        root_path = Defaults.get_system_root_path()
+        baseproduct_metadata = os.sep.join(
+            [root_path, 'etc', 'products.d', 'baseproduct']
+        )
+        return xml.parse(baseproduct_metadata)
```
schaefi: This is not safe, there is no guarantee that there is a baseproduct link. Again see units/product_setup.py where we already solved this problem... actually you wrote that code a while back ;)

PR #131, opened 2019-07-29
Fixed network setup

schaefi:
The network setup was based on an overlay bind mount of the
entire /etc/sysconfig/network directory. However that directory
also contains script code and functions that are used by system
tools like netconfig. The scripts here are not compatible between
distributions. Thus the overmount of /etc/sysconfig/network/scripts
from SLE12 to SLE15 as one example breaks the netconfig tool and
prevents the update of the /etc/resolv.conf file on network
restart. Other negative effects of that overmounting are likely.
Therefore this commit changes the way we inherit the network setup.
Only the plugin directory and the interface configurations are
taken from the host, all the rest stays untouched.

### DIFF
--- a/suse_migration_services/units/setup_host_network.py
+++ b/suse_migration_services/units/setup_host_network.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
+import glob
 import os
 import shutil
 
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/setup_host_network.py
```diff
@@ -62,11 +66,19 @@ def main():
...
+                'mount', '--bind', sysconfig_network_providers,
+                '/etc/sysconfig/network/providers'
+            ]
         )
         system_mount.add_entry(
-            sysconfig_network, '/etc/sysconfig/network'
+            sysconfig_network_providers, '/etc/sysconfig/network/providers'
         )
+        for interface_setup in glob.glob(sysconfig_interfaces):
+            if os.path.isfile(interface_setup):
```
ikapelyukhin: Small naming nitpick: `interface_setup` is not `interface_setup` if it is a directory.

schaefi: yeah, I'll find a better name

schaefi: how about it now ?

ikapelyukhin: Yeah, it's hard to come up with a good name for this. `network_... stuff`? :D But I guess `network_setup` works.

PR #132, opened 2019-07-30
Detect baseproduct name on system to be migrated

### LINKED ISSUE #129: Auto detect baseproduct to migrate

jesusbv:
By default, the migration will target SLES 15,
however this may not always be the case, e.g. SLES SAP.

Fixes #129

### DIFF
--- a/suse_migration_services/suse_product.py
+++ b/suse_migration_services/suse_product.py
@@ -1,0 +1,119 @@
+# Copyright (c) 2019 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+import os
+import glob
+from xml.etree.ElementTree import ElementTree
+
+# project
+from suse_migration_services.defaults import Defaults
+from suse_migration_services.command import Command
+from suse_migration_services.logger import log
+from suse_migration_services.exceptions import (
+    DistMigrationSUSEBaseProductException
+)
+
+
+class SUSEBaseProduct(object):
+    def __init__(self):
+        root_path = Defaults.get_system_root_path()
+        self.products_metadata = os.sep.join(
+            [root_path, 'etc', 'products.d']
+        )
+        self.prod_filenames = glob.glob(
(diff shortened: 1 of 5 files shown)


jesusbv:
The `migration_product` info (inside `migration-config.yml`) should change, as the product now should not be set by default.

Removing it from `migration_product` is an option, or change the structure to

```yaml
version: 15
arch: 'x86_64'
```
Thoughts ?

suse_migration_services/units/product_setup.py
```diff
@@ -56,73 +49,22 @@ def main():
...
+                    'Found multiple product definitions '
+                    'without element <flavor>: \n{0}'
+                    .format('\n'.join(base_products))
+                )
             log.error(message)
             raise DistMigrationProductSetupException(message)
 
         log.info('Updating Base Product to be suitable for migration')
-        delete_target_registration(baseproducts[0])
+        SUSEProduct().delete_target_registration(base_products[0])
```
schaefi: Do you have any reason why you don't create an instance of SUSEProduct() ?

```python
suse_product = SUSEProduct()
suse_product.get_baseproduct()
suse_product.delete_target_registration(...)
```

jesusbv: No reason. I will update it with an instance.

suse_migration_services/suse_product.py
```diff
@@ -0,0 +1,97 @@
...
+import glob
+from xml.etree.ElementTree import ElementTree
+
+# project
+from suse_migration_services.defaults import Defaults
+from suse_migration_services.command import Command
+from suse_migration_services.logger import log
+
+
+class SUSEProduct(object):
```
rjschwei: The naming is very misleading. The class is named "SUSEProduct" but it only handles the baseproduct. A class named "SUSEProduct", based on the name, should be generic representation of any product, i.e. as a code reader, potential user I would expect thta I can do the following:
```
product = SUSEPRoduct('/etc/products.d/sle-module-legacy.prod`)
```

and then have an object representation of he product information for this product. However the implementation is specific to the baseproduct, which in and off itself is always a symlink to some product definition.

Also the implementation of `glob.glob(os.path.join(self.products_metadata, '*.prod')` indicates the `SUSEProduct` is not singular.

IMHO we need some refactoring for clarity of code and implementation.

schaefi: > as a code reader, potential user I would expect thta I can do the following:
> ```product = SUSEPRoduct('/etc/products.d/sle-module-legacy.prod`)```

Actually that was part of the plan. I agree the class needs a bit of refactoring and I also agree the name could be better. In the scope of this project we will only deal with the base product because that will be the relevant information for a migration. Nothing else will matter. Thus I suggest to name the class

```
SUSEBaseProduct
```

The constructor of the class should search for the baseproduct on instance creation. This is basically the code from
```get_baseproduct```. The object then provides methods working with that base product. As of now this would be

* delete_target_registration
* get_baseproduct_tag_text (which I would rename to get_tag_name)
* _backup_products_metadata (which I would make public and rename to backup_products_metadata)

Thoughts ?

rjschwei: Proposed re-name and approach to only look for he `baseproduct` link in the c-tor works for me.

jesusbv: Yep, class rename and refactor was part of the plan.

PR #135, opened 2019-10-21
Add option to configure reboot method

### LINKED ISSUE #111: kexec vs reboot at the end of the upgrade
On real hardware deployments it might be a potential problem that kexec is used at the end of the upgrade as AFAIK kexec might does not always initialize/re-init the hardware. 
I am not sure if this is / can still be a problem but in case yes - it would be great if we could also configure the final action to be reboot or kexec.

tserong:
This adds a use_reboot option, which if set invokes `reboot` to reboot into the host system, instead of using `kexec`.

Fixes #111

For some more context to this, @smithfarm and @gekios have experienced (virtualized) systems hanging after migration is complete when using `kexec` to reboot. Not every time, so it's intermittent, but it's still a problem for our SES CI testing. I built a test migration image which used `reboot -f` instead, which didn't seem to help (sometimes it just got stuck on the grub screen for no reason we could ascertain), then another version which just used `reboot` (without `-f`); the intermittent post-migration hangs went away when just using `reboot`.

For this implementation I somewhat arbitrarily went with `use_reboot: true|false`, because it was easy. If more flexibility is required for any reason we could maybe do something like `reboot_method: kexec|reboot -f|reboot` instead, but I'm not sure whether that's desirable or not. Feedback most appreciated :-)

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -194,6 +194,14 @@ be debugged.
 [listing]
 debug: true|false
 
+Configure Reboot Method::
+By default, the migration system uses `kexec` to boot back into the host
+system once migration is complete.  If this is in any way problematic,
+a regular `reboot` can be requested by setting `soft_reboot: false`.
+
+[listing]
+soft_reboot: true|false
+
 == Run the Migration
 After the install of the `SLES15-Migration` package, start the migration
 by calling the following command:
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -103,6 +103,9 @@ class MigrationConfig(object):
     def is_zypper_migration_plugin_requested(self):
         return self.config_data.get('use_zypper_migration', True)
 
+    def is_soft_reboot_requested(self):
+        return self.config_data.get('soft_reboot', True)
+
     def _write_config_file(self):
         with open(self.migration_config_file, 'w') as config:
             yaml.dump(self.config_data, config, default_flow_style=False)
--- a/suse_migration_services/units/kernel_load.py
+++ b/suse_migration_services/units/kernel_load.py
@@ -24,6 +24,7 @@ from suse_migration_services.command import Command
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import log
 from suse_migration_services.path import Path
+from suse_migration_services.migration_config import MigrationConfig
 
 from suse_migration_services.exceptions import (
     DistMigrationKernelRebootException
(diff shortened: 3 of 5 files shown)


suse_migration_services/migration_config.py
```diff
@@ -103,6 +103,9 @@ def is_debug_requested(self):
     def is_zypper_migration_plugin_requested(self):
         return self.config_data.get('use_zypper_migration', True)
 
+    def is_reboot_requested(self):
```
rjschwei: To me this reads as if it implies that a system can be migrated without a reboot. "kexec" is a form of a reboot. I think a different name may help clarify: "use_reboot_exec" may be one option to more explicitly express the intend.

doc/adoc/user_guide.adoc
```diff
@@ -194,6 +194,14 @@ be debugged.
 [listing]
 debug: true|false
 
+Configure Reboot Method::
+By default, the migration system will invoke `kexec` to boot back into the
+host system once migration is complete.  If this is in any way problematic,
+a regular `reboot` can be requested by setting `use_reboot: true`.
+
+[listing]
+use_reboot: true|false
```
rjschwei: Maybe use "use_reboot_exec" to be more explicit.

schaefi: I agree the name could be misinterpreted. We can also think of something like

```
soft_reboot: true|false
```

if true, use kexec if false hardware reboot the machine. A full explanation of the option is part of the documentation.

From a code perspective it looks all good to me, thanks :+1:

tserong:
Of `use_reboot`, `use_reboot_exec` and `soft_reboot` I think the latter is slightly clearer, so I went ahead and implemented that one.

tserong:
Don't merge this yet! I'm testing it on a VM, and I swear it's still using kexec to reboot somehow, even with `soft_reboot: false` (at least, I don't see the grub screen and it's combing back up way too quickly)

tserong:
Looks like systemd is smarter than I am (see https://github.com/SUSE/suse-migration-services/issues/111#issuecomment-544816534)

schaefi:
Yes good catch a change has to be made also in ```kernel_load.py``` at the time we load the kernel for kexec

schaefi:
blocked by #138

tserong:
Rebased on top of #138, and changed to not do the `kexec --load` if soft_reboot is false

schaefi:
> Rebased on top of #138, and changed to not do the `kexec --load` if soft_reboot is false (i'll rebase it on master once the other PR is merged, just wanted to do this to get a test build in my home project, which seems to behave correctly!)

The other PR is merged now, you changes looks great to me. Just some git Kung Fu needed :)

Thanks

tserong:
git Kung Fu done :-)

schaefi:
@tserong before merge can you please fix the flake issue:

* https://travis-ci.org/SUSE/suse-migration-services/builds/601745017?utm_source=github_status&utm_medium=notification

simple one

Thanks

tserong:
OK, that should fix the flake issue (not sure how I missed that - sorry)

schaefi:
Thanks, all good that's why the tests runs here :)

perfect, let's merge it :+1:

Issue #134, opened 2019-09-16
Grub does not find ISO with more complex disk layout (separate /boot partition and LVM for root)


Martin-Weiss:
We have an md0 for /boot and an md1 with lvm for the root filesystem.

/dev/md0 mounted to /boot with ext3
/dev/system/root (in LVM on md1) mounted to / with btrfs

In this scenario we get this error:
![grafik](https://user-images.githubusercontent.com/13362573/64978943-ee968480-d8b6-11e9-9c4e-eb81f932445b.png)

I am not sure if/how often we might see such a problem but maybe this can be fixed some day.

schaefi:
Hmm, I guess our simple root_device setup might be wrong in this case. Before you rebooted can you please share the grub config file:

```/boot/grub2/grub.cfg```

Thanks.

Also the kexec based run_contained approach as documented here:

* https://documentation.suse.com/suse-distribution-migration-system/1.0/single-html/distribution-migration-system/#_run_the_migration

Could be an alternative ?

schaefi:
@Martin-Weiss Is this still of interest ? If so could you provide the details I asked for ? Thanks

Martin-Weiss:
> 
> 
> @Martin-Weiss Is this still of interest ? If so could you provide the details I asked for ? Thanks

@schaefi - yes - AFAIK this problem still exists but unfortunately I do not have access to the original system anymore so can not provide the original grub.cfg. Probably we have to wait until I have the chance to reproduce during the next upgrade.

schaefi:
ok in this case I'm going to close this one and we take a look at it once there is an opportunity to access all data for debugging. Thanks

PR #136, opened 2019-10-22
Robust detection of root device

schaefi:
Use the findmnt tool to detect the currently active root device
This is related to Issue #134

### DIFF
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -14,9 +14,8 @@ else
 fi
 
 migration_iso=$(echo /usr/share/migration-image/*-Migration.*.iso)
-root_device=$(lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | uniq | cut -f1 -d" ")
+root_device=$(findmnt -n -f -o SOURCE /)
 root_uuid=$(blkid -s UUID -o value "${root_device}")
-root_type=$(blkid -s TYPE -o value "${root_device}")
 boot_options="rd.live.image root=live:CDLABEL=CDROM"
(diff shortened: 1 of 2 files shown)


schaefi:
This is just a start, I guess there might be more changes needed to make the loop boot entry more robust. See #134 for details. We are also hitting a limit here since there could be really strange system root setups which we can't tackle all. But anything we can do to increase the current code should be imho done.

schaefi:
The loading of the filesystem module of the rootfs should be not needed and is also flaky because the return value from blkid does not necessarily be in line with the grub module name used to support this filesystem. Therefore I deleted this part.

tserong:
I just tried those two commands on my desktop system (Tumbleweed with btrfs root), and got:

```
# lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | uniq | cut -f1 -d" "
/dev/sda3
# findmnt -n -f -o SOURCE /
/dev/sda3[/@/.snapshots/1/snapshot]
```

Is the `[/@/.snapshots/1/snapshot]` bit going to cause trouble?

schaefi:
*grmbl* I would say yes. The device string is used later in a blkid call. I don't think it will work with

```blkid /dev/sda3[/@/.snapshots/1/snapshot]```

**Comment by tserong** [comment]
> I don't think it will work with
> 
> `blkid /dev/sda3[/@/.snapshots/1/snapshot]`

Correct:

```
# blkid /dev/sda3[/@/.snapshots/1/snapshot]
# echo $?
2
```

schaefi:
yeah we can't do it the way I thought it would be nicer :( Let's close this

PR #139, opened 2019-10-25
Gracefully handle syntax errors in custom config file

tserong:
If /etc/sle-migration-service.yml is not valid YAML, or is empty, yaml.safe_load() will either return something that's not a dict, or will raise an exception.  Either way, this kills the suse-migration-mount-system service, resulting in an immediate reboot; the migration does not run, and you'll have something in the log like this:

```
    load_entry_point('suse-migration-services==1.0.4', 'console_scripts', 'suse-migration-mount-system')()
  File "/usr/lib/python3.6/site-packages/suse_migration_services/units/mount_system.py", line 89, in main
    MigrationConfig().update_migration_config_file()
  File "/usr/lib/python3.6/site-packages/suse_migration_services/migration_config.py", line 87, in update_migration_config_file
    self.config_data.update(new_config)
TypeError: 'NoneType' object is not iterable
suse-migration-mount-system.service: Main process exited, code=exited, status=1/FAILURE
```

### DIFF
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -83,8 +83,28 @@ class MigrationConfig(object):
         """
         if os.path.exists(self.migration_custom_file):
             with open(self.migration_custom_file, 'r') as custom_config:
-                new_config = yaml.safe_load(custom_config)
-                self.config_data.update(new_config)
+                try:
+                    new_config = yaml.safe_load(custom_config)
+                    if isinstance(new_config, dict):
+                        # We have a valid config file (the YAML parsed
+                        # into a dict) so it's safe to update the config
+                        self.config_data.update(new_config)
+                    elif new_config is None:
+                        # The config file is empty (or is all comments),
+                        # so there's nothing to do.
+                        return
+                    else:
+                        # The config file is slightly corrupt (e.g. if it's
+                        # one line "foo:bar" with no space after the ':', it
+                        # will be a str, not a dict, and thus unusable).
+                        raise Exception
+                except Exception:
+                    # Either the config file is *really* broken (yaml.safe_load
+                    # raised an exception), or it's slightly broken as mentioned
+                    # above.  Either way, we can't use it.
+                    log.warning("Can't use custom configuration ({} is not valid YAML)".format(
+                        self.migration_custom_file))
+                    return
 
             self._write_config_file()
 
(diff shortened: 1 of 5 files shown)


tserong:
(I hit this when I was doing random manual testing, and had commented out everything in /etc/sle-migration-service.yml, resulting in `yaml.safe_load()` returning NoneType)

schaefi:
I completely agree we should have a better validation of the input file. In the beginning there was only one simple setting (debug on/off) but now there are more options and I see even more to come. So my idea was to add a schema and run proper validation in the same way I did it on the Azure project:

* https://github.com/SUSE-Enceladus/azure-li-services/blob/master/azure_li_services/schema.py

I think we should aim for an implementation that does a proper validation before yaml.load.
What do you think ?

schaefi:
@tserong by the way you have write permissions to the project, so you don't have to go with forked repos :)

rjschwei:
@schaefi I agree it is likely that the configuration options will continue to expand and validation against a schema is a good idea.

@tserong You are not suggesting that we do the migration and ignore the config errors, correct? From reading the initial description I wasn't sure. Clarification appreciated.

tserong:
> I think we should aim for an implementation that does a proper validation before yaml.load.
> What do you think ?

Yes, a proper validation would be better than what I've got here :-)

> @tserong by the way you have write permissions to the project, so you don't have to go with forked repos :)

Thanks!

tserong:
> You are not suggesting that we do the migration and ignore the config errors, correct?

Ah... I can see how that wasn't quite clear :-) I don't want to necessarily ignore config errors, but as it stands right now, if there are any config errors, or even if the config file exists but is completely empty, or is just all comment lines (both of the latter cases being OK), this will raise an exception, which means the suse-migration-mount-system service will fail, and the system will immediately reboot, and you'll have no idea why that happened.

This commit allows the migration to proceed in all the above cases, with an error logged if the config is corrupt. As you say, this is probably not correct behaviour. I suspect we need something like:

- If the config file is empty, or is all comments, or is valid, the migration should proceed as usual
- If the config file is corrupt, we need to be able to still mount the host system so we can log the error, but should then bail out after that point.

So, this needs some reworking...

rjschwei:
@tserong thanks for the clarification I agree with the decision path

tserong:
I've reworked this to do schema validation, and will update the PR tomorrow once I've had a chance to do a few manual test runs to make sure the messages that land in the log file are actually good.

schaefi:
> Or should I close this and open a new PR, given the focus changed 

I would prefer if we can close this one and open another request. As you already said the scope of the change will be relatively different. Thanks much for your effort

tserong:
Closing in favour of #142

PR #140, opened 2019-10-30
Auto detect migration product

### LINKED ISSUE #129: Auto detect baseproduct to migrate

jesusbv:
As there could be different targets (SLES, SLES_SAP, etc)
instead of setting the target in the custom file, the migration
system auto detects the migration product.

This Fixes #129

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -8,7 +8,7 @@ Marcus Schäfer; Jesús Velázquez
 :Authors: Marcus Schäfer and Jesús Bermúdez Velázquez
 :Publication_Date: TBD
 :Latest_Version: 1.1.0
-:Contributors: 
+:Contributors:
 :Repo: https://github.com/SUSE/suse-migration-services[suse-migration-services]
 
 ifdef::env-github[]
@@ -154,12 +154,14 @@ manually. Once done, the upgrade process uses `zypper dup` and expects
 all required repositories to be setup correctly.
 
 Specify Migration Product::
-With the SLES15-Migration package a predefined product setup is
-already applied. That setting can be overwritten with one of the
+With the SLES15-Migration package an auto detected product is
+already applied. If that detection failed, the setting can be overwritten with one of the
 following product specifications:
 
 * SLES/15/x86_64
 * SLES/15.1/x86_64
+* SLES_SAP/15/x86_64
+* SLES_SAP/15.1/x86_64
 
 [listing]
 migration_product: SLES/15/x86_64
--- a/suse_migration_services/defaults.py
+++ b/suse_migration_services/defaults.py
@@ -16,6 +16,7 @@
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
 import os
+from collections import namedtuple
 
 
 class Defaults(object):
@@ -85,3 +86,16 @@ class Defaults(object):
     @classmethod
     def get_system_sshd_config_path(self):
         return '/etc/ssh/sshd_config'
+
+    @classmethod
+    def get_os_release(self):
+        with open('/etc/os-release', 'r') as handle:
+            keys, values = zip(
+                *[
+                    (key.lower(), value.strip('\'"')) for (key, value) in (
+                        line.strip().split('=', 1) for line in
+                        handle.read().strip().split(os.linesep)
+                    )
+                ]
+            )
+            return namedtuple('OSRelease', keys)(*values)
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -21,6 +21,7 @@ from textwrap import dedent
 
 # project
 from suse_migration_services.defaults import Defaults
+from suse_migration_services.suse_product import SUSEBaseProduct
 from suse_migration_services.logger import log
 from suse_migration_services.exceptions import (
     DistMigrationProductNotFoundException
@@ -53,13 +54,17 @@ class MigrationConfig(object):
         The value returned is passed to the --product option in the
         zypper migration call
         """
+        # check for a custom migration product
         migration_product = self.config_data.get('migration_product')
         if not migration_product:
-            message = (
-                'Migration product not found, aborting migration attempt.'
-            )
-            log.error(message)
-            raise DistMigrationProductNotFoundException(message)
+            migration_product = SUSEBaseProduct().get_product_name()
+            if not migration_product:
+                # auto detection went wrong
+                message = (
+                    'Migration product not found, aborting migration attempt.'
+                )
+                log.error(message)
+                raise DistMigrationProductNotFoundException(message)
 
         return migration_product
 
--- a/suse_migration_services/suse_product.py
+++ b/suse_migration_services/suse_product.py
@@ -117,3 +117,23 @@ class SUSEBaseProduct(object):
                 'Parsing XML file {0} failed with: {1}'
                 .format(self.base_product, issue)
             )
+
+    def get_product_name(self):
+        """Get the product name to be migrated to."""
+        migration_product_name = None
+        try:
+            name = self.get_tag('name')[0]
+            arch = self.get_tag('arch')[0]
+            if name and arch:
+                migration_product_name = '/'.join(
+                    [name, self.get_default_target_version(), arch]
+                )
+        except Exception as issue:
+            log.error(
+                'Base product could not be detected: {0}.'.format(issue)
+            )
+
+        return migration_product_name
+
+    def get_default_target_version(self):
+        return Defaults.get_os_release().version_id
--- a/test/data/migration-config.yml
+++ b/test/data/migration-config.yml
@@ -1,5 +1,3 @@
-migration_product: 'SLES/15/x86_64'
-
 preserve:
   rules:
      - /etc/udev/rules.d/a.rules
--- a/test/unit/defaults_test.py
+++ b/test/unit/defaults_test.py
@@ -1,3 +1,9 @@
+from unittest.mock import (
+    patch, MagicMock
+)
+import io
+from collections import namedtuple
+
 from suse_migration_services.defaults import Defaults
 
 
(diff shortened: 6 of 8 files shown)


suse_migration_services/migration_config.py
```diff
@@ -53,13 +54,20 @@ def get_migration_product(self):
...
-        migration_product = self.config_data.get('migration_product')
-        if not migration_product:
+        migration_product = None
+        suse_base_product = SUSEBaseProduct()
+        if suse_base_product.is_version_supported():
+            name = suse_base_product.get_tag('name')[0]
+            arch = suse_base_product.get_tag('arch')[0]
+            migration_product = '/'.join(
+                [name, str(suse_base_product.get_default_target_version()), arch]
+            )
```
schaefi: The creation of the product string for the later use with the zypper migration plugin is a responsibility of the SUSEBaseProduct class. I think the class should provide a method that just returns this information, e.g

```python
migration_product_name = suse_base_product.get_product_name()
```

schaefi: The ```get_product_name``` method implementation imho should contain the code that reads the tags and constructs the name

suse_migration_services/suse_product.py
```diff
@@ -117,3 +117,27 @@ def get_tag(self, tag):
...
+        return is_supported
+
+    def get_default_target_version(self):
+        return 15
+
+    def get_minimum_baseproduct_supported(self):
+        return 12
+
+    def get_minimum_patch_level_supported(self):
+        return 3
```
schaefi: I do not understand the reason why this code is needed ? I think we limit ourselves with a constraint that is not needed. If a migration exists or not is defined by the image we deliver. There is no need to check on a minimum product version. Always think about this has to be maintained, I would just drop it

schaefi:
Also missing is an update on the documentation about the product handling

suse_migration_services/suse_product.py
rjschwei: Well, yes we had this discussion in some other  thread about SCC and the defined migration paths. My question is do we really want to continue to answer all the support questions? We've had a fair bit of churn already. Making the blast radius larger is not necessarily helpful. While the product QA team is happy to help us testing I don't think it is reasonable to test every release of the migration code with every origin point (SLES 12 SLES 12 SP1|2|3|4|5) to every destination (SLES 15 SLES SP1|2|3|4|5|6) the combination of which makes 42 test cases eventually. Things get worse when we consider there will be a next SLES.

I know, I know, the answer is that if it doesn't work it's probably the SCC team's problem. Problem with that approach is that we still end up answering all the questions. I think the list traffic proofs my point on this one.

I like a defined origin and destination. If the user over-writes the target in the config then we can waive our hands.

Yes, when $NEXT-SLES gets released we need to re-work the definition of origin and destination.

In short I like the definition of the parameters.

jesusbv: > This is an important point and I would say yes. Any autodetection can fail. if it fails in the current implementation there is no way to fix this other than releasing a new image. So yes please treat an eventual given product definition from the custom config file as effective.
> 
From what I understood, I treated it as auto detection first and if failure, check custom product definition from user.
Not sure that is the preferred choice, where user could think something is wrong when their setting is ignored if auto detection worked.
IMO, the checking for a custom product should have priority and if no custom product, then auto detect the target.
Thoughts?

doc/adoc/user_guide.adoc
```diff
@@ -154,8 +154,8 @@ manually. Once done, the upgrade process uses `zypper dup` and expects
 all required repositories to be setup correctly.
 
 Specify Migration Product::
-With the SLES15-Migration package a predefined product setup is
-already applied. That setting can be overwritten with one of the
+With the SLES15-Migration package an auto detected product is
+already applied. If that detection failed, the setting can be overwritten with one of the
 following product specifications:
 
 * SLES/15/x86_64
```
schaefi: Do we want to make the list of possible product names more complete ? I was thinking on also adding ```SLES_SAP/15/x86_64``` Furthermore I think it would be good to have a link to a page that shows the products suse provides. This is because in theory this upgrade process works for any of them. Thus the list we mention here is most probably incomplete. I'm not sure if suse provides such an information though. Thoughts ?

suse_migration_services/suse_product.py
```diff
@@ -117,3 +117,24 @@ def get_tag(self, tag):
...
+            arch = self.get_tag('arch')[0]
+            if name and arch:
+                migration_product_name = '/'.join(
+                    [name, str(self.get_default_target_version()), arch]
+                )
+        except Exception as issue:
+            log.error(
+                'No auto detection of baseproduct: {0}. {1}'
+                .format(issue, 'Continue without auto detected product')
+            )
```
schaefi: I think the log message could be misleading. The reaction on what happens if get_product_name() couldn't find the product is not defined. It depends on the code that handles this result. Thus writing a message that says "... we continue without auto detected product" could be just wrong. I think the error message should be just saying what happened. For example:

```
log.error(
    'Base product could not be detected: {0}'.format(issue)
)
```

We also could write an information that the user can specify the product in the custom config file as a plus.
Does this makes sense ?

suse_migration_services/suse_product.py
```diff
@@ -117,3 +117,24 @@ def get_tag(self, tag):
...
+
+        return migration_product_name
+
+    def get_default_target_version(self):
+        return 15
```
schaefi: Ok here we have a problem. Thinking further and assuming there is SLES-16 available. We design an image that upgrades from sles15 to sles16. The live iso image that drives the process will be a sles16 system and the user installs the package that provides the live iso image on his sles15 system. This code will now construct the product name ```SOME_NAME/15/SOME_ARCH``` and it's for sure wrong.

I suggest that the target version is detected from the ```/etc/os-release``` file VERSION_ID variable taken from the live iso images, ,meaning just read ```/etc/os-release```

On the currently designed image this will be 15 on a potential later image this will be 16 and so on

If we add auto detection then real auto detection please :)

Thoughts ?

suse_migration_services/suse_product.py
schaefi: > IMO, the checking for a custom product should have priority and if no custom product, then auto detect the target.

yep

schaefi: > Cloud Module enabled can install the migration packages,

I think you are wrong here. We had long discussions with the maintenance team when we recently had to release the system. There is a switch that keeps a package internal or not. suse-migration-services is a package that gets never released. It will not be available to customers on the channel. It only exists in the internal data structure (the pool) where we can access it. customers should not have the ability to "zypper in suse-migration-services"

suse_migration_services/suse_product.py
schaefi: In addition I should also mention that the people who write the zypper migration plugin explicitly stated that it only works if the plugin is already called from the target system. This chicken & egg problem is solved by us with running it from the live iso which is always build for that target. So reading the os-release file is also the correct detection way and in line with the design concept

doc/adoc/user_guide.adoc
jesusbv: I'm more than happy to put the link with all products suse provides.

suse_migration_services/suse_product.py
jesusbv: Fixed.

doc/adoc/user_guide.adoc
rjschwei: The PRD specifies supported upgrade path, but I am not certain this ends up in the documentation of not. But a search on suse.com/documentation in the admin guide for SLES 15 should reveal the information.

I do know that the product definitions as they are expected for this process have not been documented in the past.

suse_migration_services/suse_product.py
```diff
@@ -117,3 +117,33 @@ def get_tag(self, tag):
...
+        os_release_file = Defaults.get_os_release_file()
+        with open(os_release_file, 'r') as release_file:
+            os_release_data = release_file.read()
+        os_release_data = os_release_data.split(os.linesep)
+        label = 'VERSION_ID'
+        for data in os_release_data:
+            if label in data:
+                version_id = data.split('=')[1]
+                version_id = version_id.replace('"', '')
+        return version_id
```
schaefi: I would implement this more generic. I thought a function that returns class like OSRelease tuple from the defaults space is more generic. For example:

```python
import os
from collections import namedtuple

class Defaults:

    @staticmethod
    def get_os_release():
        with open('/etc/os-release') as handle:
            keys, values = zip(
                *[
                    (key.lower(), value.strip('\'"')) for (key, value) in (
                        line.strip().split('=', 1) for line in
                            handle.read().strip().split(os.linesep)
                    )
                ]
            )
            return namedtuple('OSRelease', keys)(*values)
```

you would then access this by

```python
Defaults.get_os_release().version_id
```

Thoughts ?

suse_migration_services/defaults.py
```diff
@@ -85,3 +85,7 @@ def get_system_ssh_host_keys_glob_path(self):
     @classmethod
     def get_system_sshd_config_path(self):
         return '/etc/ssh/sshd_config'
+
+    @classmethod
+    def get_os_release_file(self):
+        return '/etc/os-release'
```
schaefi: See my comment below

doc/adoc/user_guide.adoc
schaefi: I couldn't come up with a one line location that summarizes this. So I suggest to keep the current state. Thanks @jesusbv for the list update

suse_migration_services/suse_product.py
jesusbv: I thought about a solution like that (returning a dictionary) but changed it as we want _one_ of the fields.
I will implement the approach.

schaefi: I agree, it is usually better to implement what's needed. In this particular case however I can imagine we might need more information from the os release at other parts of the code or for more fine grained selection. Also writing a full parser is less code than looking up a particular element. That's why I thought it's better to be generic from the beginning on. Thanks for changing it

jesusbv:
When all is ok, I would like to squash all commits into one.

PR #146, opened 2019-11-08
Fixed reboot call procedure

schaefi:
the systemd magic that detects whether to do kexec or not, is only
applied if we invoke reboot. systemd internally maps reboot to one or
the other of 'systemctl reboot' or 'systemctl kexec'. Therefore the
current implementation which just call 'systemctl reboot' does not
transparently detect for kexec. As we don't want to rely on the
magic behind the reboot call we explicitly call systemctl with
reboot or kecec depending on the configured soft_reboot
configuration option. This is then also in line with the code
that runs 'kexec load' which also depends on the configured
soft_reboot value.

### DIFF
--- a/suse_migration_services/units/reboot.py
+++ b/suse_migration_services/units/reboot.py
@@ -66,15 +66,14 @@ def main():
                         )
                     )
                 )
+            if not MigrationConfig().is_soft_reboot_requested():
+                restart_system = 'reboot'
+            else:
+                restart_system = 'kexec'
             log.info(
-                # reboot is performed through systemd. The call through
-                # systemd checks if there is a kexec loaded kernel and
-                # transparently turns 'systemctl reboot' into
-                # 'systemctl kexec'. Thus both ways, soft and hard
-                # reboot are managed in one call.
                 'Reboot system: {0}{1}'.format(
                     os.linesep, Command.run(
-                        ['systemctl', 'reboot']
+                        ['systemctl', restart_system]
                     )
                 )
             )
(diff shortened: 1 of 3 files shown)


suse_migration_services/units/reboot.py
```diff
@@ -66,16 +66,13 @@ def main():
                         )
                     )
                 )
+            if not MigrationConfig().is_soft_reboot_requested():
+                command_list = ['systemctl', 'reboot']
+            else:
+                command_list = ['systemctl', 'kexec']
```
jesusbv: All looks good, the only thing that is different for the `if` is the command to run
maybe something like this:

```python
            if not MigrationConfig().is_soft_reboot_requested():
                restart = 'reboot'
            else:
                restart = 'kexec'    
            command_list = ['systemctl', restart]
```

suse_migration_services/units/reboot.py
```diff
@@ -66,16 +66,13 @@ def main():
...
-                # reboot is performed through systemd. The call through
-                # systemd checks if there is a kexec loaded kernel and
-                # transparently turns 'systemctl reboot' into
-                # 'systemctl kexec'. Thus both ways, soft and hard
-                # reboot are managed in one call.
                 'Reboot system: {0}{1}'.format(
-                    os.linesep, Command.run(
-                        ['systemctl', 'reboot']
-                    )
+                    os.linesep, Command.run(command_list)
```
jesusbv: the comment about could make this
```py
os.linesep, Command.run(
    ['systemctl', restart]
)
```
so no `command_list` variable and it could be more explicit

schaefi: Good point, I've adapted that part, so it is a nicer read

schaefi:
@jesusbv how about it now ?

jesusbv:
> @jesusbv how about it now ?

Great !

PR #153, opened 2019-11-22
Add valid registration verification

### LINKED ISSUE #150: Verify that the system is registered before the migration
Having a valid registration is a precondition for a successful migration. If registration is missing or invalid, the migration will die and produce non-obvious error messages as the result.

jesusbv:
System must be registered for the migration to succeed.

This Fixes #150

### DIFF
--- a/suse_migration_services/suse_connect.py
+++ b/suse_migration_services/suse_connect.py
@@ -1,0 +1,34 @@
+# Copyright (c) 2019 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


suse_migration_services/suse_connect.py
```diff
@@ -0,0 +1,27 @@
...
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+# project
+from suse_migration_services.command import Command
+
+
+class SUSEConnect(object):
+    def is_registered(self):
+        status_command_result = Command.run(
+            ['SUSEConnect', '--list-extensions']
```
ikapelyukhin: How about capturing the output of SUSEConnect as the error message for the logs? (If the exit code is not zero, that is)

schaefi: yes the output of the SUSEConnect command should be logged in case of an error

PR #154, opened 2020-01-08
Fixed static method call

schaefi:
The method is_registered is implemented as object method
but is called as static method SUSEConnect.is_registered().
This is wrong and causes a python runtime error on missing
argument: self

### DIFF
--- a/suse_migration_services/suse_connect.py
+++ b/suse_migration_services/suse_connect.py
@@ -21,7 +21,8 @@ from suse_migration_services.logger import log
 
 
 class SUSEConnect:
-    def is_registered(self):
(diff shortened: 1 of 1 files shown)


schaefi:
I was testing the migration and it failed with

```
[INFO    ]: 10:05:24 | Update certificate pool
[ERROR   ]: 10:05:25 | Preparation of zypper metadata failed with is_registered() missing 1 required positional argument: 'self'
```

This is caused by the changes we did during the workshop. The patch provided here should fix it

rjschwei:
LGTM

jesusbv:
Good catch. LGTM.

PR #156, opened 2020-01-08
Consolidate image description in git

schaefi:
Maintain the live image description which implements the
platform for the upgrade process here in git. This allows
for better maintenance of the description data as well as
provides scope (pubcloud) to the description and also
keeps the version number in sync with the process code

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -5,3 +5,4 @@ tag = True
 
 [bumpversion:file:suse_migration_services/version.py]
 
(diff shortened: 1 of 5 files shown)


schaefi:
@jesusbv I'm going to find a place in the CrossCloud project for building our migration live image because I think the current place directly under Devel:PubCloud is wrong. Also having the image descrptions here in git is much better for maintenance. Eventually the storage team can move their stuff to git as well... let's see

PR #157, opened 2020-01-08
Fixed registration check

schaefi:
The SUSEConnect call to check if the system is registered was
called in the live migration system environment. Of course there
it will fail because the live migration system is never registered.
The call has to be performed chrooted inside of the system we
want to migrate. Thus the caller arguments as well as the call
time in the preparation unit needs a fix

### DIFF
--- a/suse_migration_services/suse_connect.py
+++ b/suse_migration_services/suse_connect.py
@@ -18,13 +18,21 @@
 # project
 from suse_migration_services.command import Command
 from suse_migration_services.logger import log
+from suse_migration_services.defaults import Defaults
 
 
 class SUSEConnect:
     @staticmethod
     def is_registered():
+        """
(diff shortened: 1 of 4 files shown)


schaefi:
I'm not sure what we did during the workshop but at least me obviously failed on a good review ;) The check for the registration did not work. In my tests it showed:

```
[ERROR   ]: 16:34:31 | To list extensions, you must first register the base product, using: SUSEConnect -r <registration code>

[ERROR   ]: 16:34:31 | System not registered. Aborting migration.
```

which is clear because we ran this on the live migration system

jesusbv:
> I'm not sure what we did during the workshop 

Pretty sure around 87% we tested it at the time.

rjschwei:
The problem was clearly that we had no beer during the workshop.

PR #163, opened 2020-01-17
Support lvm root

schaefi:
Add support for LVM managed devices
    
LVM managed devices requires the detection as such in the
mount_system service as well as the activation of the
volume group prior mounting. This Fixes bsc#1156068


schaefi:
I'm still testing for regressions and inconsistencies but so far it's ok for a first review

schaefi:
Thanks for the review, we need to keep this open until I can give it a test when the migration backend is fixed to make sure it does not cause any regressions. Thanks

schaefi:
migration worked :-) going to merge

PR #166, opened 2020-02-04
Mount only os entries

### LINKED ISSUE #165: OS mounting fails for external references
Recently we had a problem with a customer where the migration fails in the mount service.
The mount service searches the disks for an OS reads the fstab and mounts everything in the fstab under /system-root. This assumes the fstab contains only OS relevant entries. However on e.g HANA systems external storage references are mounted into the system as well. Inside of the migration live image those references are not available and mounting of those fails. In addition those external data is not relevant for the OS upgrade.

We should make the system mount unit more robust in a way that it only takes mountpoints into account that matches the disk which holds the fstab file.

Current workaround for customers is:

* cp etc/fstab away
* modify /etc/fstab to contain only OS entries
* reboot and upgrade
* move fstab back once the migrated system is up again
* mount -a

This process worked in the session with the customer and is ok as a workaround but the migration system can do better :)

jesusbv:
Before, mount service mounted everything on
/etc/fstab, now that service takes into account
the entries that matches the disk that holds the
fstab file.

This Fixes #165

### DIFF
--- a/suse_migration_services/fstab.py
+++ b/suse_migration_services/fstab.py
@@ -21,6 +21,7 @@ from collections import namedtuple
 
 # project
 from suse_migration_services.defaults import Defaults
+from suse_migration_services.command import Command
 
 log = logging.getLogger(Defaults.get_migration_log_name())
 
@@ -32,8 +33,10 @@ class Fstab:
     def __init__(self):
         self.fstab = []
         self.fstab_entry_type = namedtuple(
-            'fstab_entry_type', ['fstype', 'mountpoint', 'device', 'options']
+            'fstab_entry_type', ['fstype', 'mountpoint', 'device',
+                                 'options', 'unix_device']
         )
+        self.root_disk = None
 
     def read(self, filename):
         """
@@ -76,14 +79,18 @@ class Fstab:
                                 )
                             )
                             continue
+                    unix_device = os.path.realpath(device_path)
                     self.fstab.append(
                         self.fstab_entry_type(
                             fstype=fstype,
                             mountpoint=mountpoint,
                             device=device_path,
-                            options=options
+                            options=options,
+                            unix_device=unix_device
                         )
                     )
+
+            self._get_root_disk()
 
     def add_entry(self, device, mountpoint, fstype=None, options=None):
         self.fstab.append(
@@ -91,7 +98,8 @@ class Fstab:
                 fstype=fstype or 'none',
                 mountpoint=mountpoint,
                 device=device,
-                options=options or 'defaults'
+                options=options or 'defaults',
+                unix_device=os.path.realpath(device)
             )
         )
 
@@ -103,13 +111,32 @@ class Fstab:
         """
         with open(filename, 'w') as fstab:
             for entry in self.fstab:
-                fstab.write(
-                    '{0} {1} {2} {3} 0 0{4}'.format(
-                        entry.device, entry.mountpoint,
-                        entry.fstype, entry.options,
-                        os.linesep
+                if self._is_on_root(entry.device):
+                    fstab.write(
+                        '{0} {1} {2} {3} 0 0 {4}{5}'.format(
+                            entry.device, entry.mountpoint,
+                            entry.fstype, entry.options,
+                            entry.unix_device, os.linesep
+                        )
                     )
-                )
 
     def get_devices(self):
-        return self.fstab
+        return [
+            entry for entry in self.fstab if self._is_on_root(entry.device)
+        ]
+
+    def _get_root_disk(self):
+        """Get the disk name for / partition."""
+        for entry in self.fstab:
+            if entry.mountpoint == '/':
+                self.root_disk = Command.run(
+                    ['lsblk', '-no', 'pkname', entry.unix_device]
+                ).output
+                break
+
+    def _is_on_root(self, device_path):
+        """Check that the entry belongs to the root disk."""
+        lsblk_call = Command.run(
+            ['lsblk', '-n', '-o', 'pkname', device_path]
+        )
+        return lsblk_call.output == self.root_disk
--- a/test/data/fstab
+++ b/test/data/fstab
@@ -3,6 +3,7 @@ UUID=daa5a8c3-5c72-4343-a1d4-bb74ec4e586e  swap       swap  defaults        0  0
 UUID=FCF7-B051                             /boot/efi  vfat  defaults        0  0
 LABEL=foo  /home      ext4  defaults        0  0
 PARTUUID=3c8bd108-01 /bar ext4 defaults 0 0
+UUID=00d53ca7-11a4-4455-848e-ee52b1173518  /dif/disk  ext4  acl,user_xattr  0  1
 /dev/mynode /foo ext4 defaults 0 0
 
-# this comment line and the line above should be ignored by the parser
+# this comment line and the previous two lines above should be ignored by the parser
--- a/test/unit/fstab_test.py
+++ b/test/unit/fstab_test.py
@@ -2,7 +2,7 @@ import io
 import logging
 from pytest import fixture
 from unittest.mock import (
-    MagicMock, patch, call
+    MagicMock, patch, call, Mock
 )
 from suse_migration_services.fstab import Fstab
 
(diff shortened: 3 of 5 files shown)


rjschwei:
Wondering out loud, just some food for thought. I am wondering if we should add more comprehension to handling of fstab? 

We know which filesystems/mount-points we need to upgrade a system. Some of these are

```/boot
/usr, 
/ ....```

There's probably a list of 10 of these. If we would look through fstab and find entries that match our "system directories" and then mount those then we'd catch everything we need, and as the "non cloud use case" we'd catch the case where `/usr` is on a different disk for example. The system directories in our list must be mountable, period. If they are not mountable the system will not run and migration is not possible. Meaning we can "blindly" mount any one of those system directories if they are listed in fstab and then complain if it cannot be mounted and exit.

If I manually verify the current implementation in Azure where we have separate boot partitions things fail:

From get_root_disk()
```
-> lsblk -r -n -o 'NAME,MOUNTPOINT'
sda 
sda1 
sda2 /boot/efi
sda3 /boot
sda4 /
sdb 
sdb1 
sr0 
-> lsblk -n -o pkname /dev/sda4
sda
```
Note that I added '/dev' to the second call of lsblk, something the code does not do and an error in the PR.

And then we have, from is_on_root()

```
-> lsblk -n o pkname /dev/disk/by-uuid/a8e8b4ac-b2b7-4abe-8eac-2904ea39e5cf
lsblk: o: not a block device
lsblk: pkname: not a block device
sda3   8:3    0   1G  0 part /boot
```
where the UUID is directly from fstab

```
-> cat /etc/fstab
UUID=de852492-9570-432f-b45d-dbe503532a03 / xfs defaults 1 1
UUID=a8e8b4ac-b2b7-4abe-8eac-2904ea39e5cf /boot xfs defaults 0 0
UUID=8677-6368 /boot/efi vfat defaults 0 0
```
given that is_on_root() makes a straight comparison of the lsblk output to the previously determined root device I don't see how this would work.

Anyway if we'd loop through the fstab entries, which we already do anyway
```
for line in ....
   if mountpoint in required_system_dirs:
       system_dirs_to_mount.append(mountpoint)
```
and then we'd mount everything that's in `system_dirs_to_mount`. I realize I am simplifying things a bit.

Anyway, as far as I can tell the current proposal will not work. And by choosing a different approach we could capture an adjacent use case, i.e. supporting system mount points on different devices. Yes, the adjacent use-case is not our primary target for system migration support.

**Comment by jesusbv** [comment]
> Wondering out loud, just some food for thought. I am wondering if we should add more comprehension to handling of fstab?
> 
> We know which filesystems/mount-points we need to upgrade a system. Some of these are
> 
> ```
> /usr, 
> / ....```
> 
> There's probably a list of 10 of these. If we would look through fstab and find entries that match our "system directories" and then mount those then we'd catch everything we need, and as the "non cloud use case" we'd catch the case where `/usr` is on a different disk for example. The system directories in our list must be mountable, period. If they are not mountable the system will not run and migration is not possible. Meaning we can "blindly" mount any one of those system directories if they are listed in fstab and then complain if it cannot be mounted and exit.
> 
> If I manually verify the current implementation in Azure where we have separate boot partitions things fail:
> 
> From get_root_disk()
> ```
> 
> -> lsblk -r -n -o 'NAME,MOUNTPOINT'
> sda
> sda1
> sda2 /boot/efi
> sda3 /boot
> sda4 /
> sdb
> sdb1
> sr0
> -> lsblk -n -o pkname /dev/sda4
> sda
> 
> ```
> Note that I added '/dev' to the second call of lsblk, something the code does not do and an error in the PR.

True. I left it behind. Will fix it now.
> 
> And then we have, from is_on_root()
> ```
> 
> -> lsblk -n o pkname /dev/disk/by-uuid/a8e8b4ac-b2b7-4abe-8eac-2904ea39e5cf
> lsblk: o: not a block device
> lsblk: pkname: not a block device
> sda3 8:3 0 1G 0 part /boot
> 

It should be `-o` as in the previous `lsblk` call. 
> ```
> where the UUID is directly from fstab
> ```
> 
> -> cat /etc/fstab
> UUID=de852492-9570-432f-b45d-dbe503532a03 / xfs defaults 1 1
> UUID=a8e8b4ac-b2b7-4abe-8eac-2904ea39e5cf /boot xfs defaults 0 0
> UUID=8677-6368 /boot/efi vfat defaults 0 0
> 
> ```
> given that is_on_root() makes a straight comparison of the lsblk output to the previously determined root device I don't see how this would work.
> 
> Anyway if we'd loop through the fstab entries, which we already do anyway
> ```
> 
> for line in ....
> if mountpoint in required_system_dirs:
> system_dirs_to_mount.append(mountpoint)
> 
> ```
> and then we'd mount everything that's in `system_dirs_to_mount`. I realize I am simplifying things a bit.
> 
> Anyway, as far as I can tell the current proposal will not work. And by choosing a different approach we could capture an adjacent use case, i.e. supporting system mount points on different devices. Yes, the adjacent use-case is not our primary target for system migration support.
> ```

File: suse_migration_services/fstab.py [hunk 1]
```diff
@@ -113,3 +116,27 @@ def export(self, filename):
...
+        for entry in lsbkl_mountpoint_call.output.split(os.linesep):
+            block_record = entry.split()
+            if len(block_record) >= 2 and block_record[1] == '/':
+                lsbkl_disk_call = Command.run(
+                    [
+                        'lsbkl', '-n', '-o', 'pkname',
+                        '/dev/{}'.format(block_record[0])
+                    ]
+                )
+                return lsbkl_disk_call.output
```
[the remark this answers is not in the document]
schaefi: I also think the method can be done simpler. What we want is the unix disk device that has the root partition. From the read we know the unix device that holds the root partition. So it could be written like:

```python
def _get_root_disk(self):
    for fstab_entry in self.fstab:
        if fstab_entry.mountpoint == '/':
           self.root_disk = Command.run(
               ['lsblk', '-no', 'pkname', fstab_entry.unix_device]
           ).output
           break
```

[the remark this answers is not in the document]
schaefi: I would call _get_root_disk at the end of the read() method and initialize the class with empty self.root_disk

suse_migration_services/fstab.py
```diff
@@ -113,3 +116,27 @@ def export(self, filename):
...
+
+    def is_on_root(self, device_path):
+        """Check that the entry belongs to the root disk."""
+        lsblk_call = Command.run(
+            ['lsblk', '-n', '-o', 'pkname', device_path]
+        )
+        return lsblk_call.output == self.root_disk
```
schaefi: Looks good :+1:

schaefi: But would make the method also private: _is_on_root_disk(...)

schaefi: With the changes above the two methods export() and get_devices() needs to be adapted to write or return only devices that are _is_on_root_disk()

schaefi:
> Wondering out loud, just some food for thought. I am wondering if we should add more comprehension to handling of fstab?

The only comprehension I'd like to add in the Fstab class is the knowledge of the disk where the root filesystem lives. In the migration process we expect any components that belongs to the system we can upgrade to be on the disk that holds the root filesystem. Only those should be taken into account. Technically it's possible to have /boot on another disk than / but who in Gods name does that and in the cloud it's even impossible because we come from an image as anybody else
does

> We know which filesystems/mount-points we need to upgrade a system. Some of these are 
> There's probably a list of 10 of these. If we would look through fstab

This is an assumption you can't be sure about. I think adding code that does a selection by a static list even though that list follows standards is an unstable design

If we do it right the additional code to the fstab class will be only a couple lines of code and clear in its intention. It will then solve any of the customers manually added/attached storage XYZ components and we can also write a good logging message that we do not handle filesystem XYZ because it's not on the rootfs disk. If we now found out there are cases where we need to refine this we can do so if needed

schaefi:
But while we are talking about it I think we can solve this even simpler

The fstab class could check if the device exists at the level of the export() and get_devices() methods and only takes those into account that are present.

At the time of the fstab file reading all device activation code (e.g LVM) is already done. So this should work and is more straight forward

Thoughts ?

rjschwei:
> But while we are talking about it I think we can solve this even simpler
> 
> The fstab class could check if the device exists at the level of the export() and get_devices() methods and only takes those into account that are present.
> 
> At the time of the fstab file reading all device activation code (e.g LVM) is already done. So this should work and is more straight forward
> 
> Thoughts ?

Yes, looping trough the fstab entries and then checking if the reverenced device is present avoids the issue with the static list and will solve the "/usr" is on a different disk problem that would be a "non cloud usecase" I like the idea.

suse_migration_services/fstab.py
```diff
@@ -103,13 +111,32 @@ def export(self, filename):
...
+            if entry.mountpoint == '/':
+                self.root_disk = Command.run(
+                    ['lsblk', '-no', 'pkname', entry.unix_device]
+                ).output
+                break
+
+    def _is_on_root(self, device_path):
+        """Check that the entry belongs to the root disk."""
+        lsblk_call = Command.run(
+            ['lsblk', '-n', '-o', 'pkname', device_path]
```
schaefi: just for consistency we should decide for '-n', '-o' or '-no' in all cases. I vote for the shorter form

schaefi:
> Yes, looping trough the fstab entries and then checking if the reverenced device is present avoids the issue with the static list and will solve the "/usr" is on a different disk problem that would be a "non cloud usecase" I like the idea.

ok so I think it would also be the easier implementation. @jesusbv sorry but I think the code changes again :) sometimes a better solution is found while coding. Hope you don't mind ?

schaefi:
closed in favor of #167

PR #167, opened 2020-02-05
Mount only existent devices

### LINKED ISSUE #165: OS mounting fails for external references

jesusbv:
Before, mount service mounted everything on
/etc/fstab, now that service takes into account
the entries that have an existent path.

This Fixes #165

### DIFF
--- a/suse_migration_services/fstab.py
+++ b/suse_migration_services/fstab.py
@@ -69,21 +69,23 @@ class Fstab:
                         )
                     else:
                         device_path = device
-                        if not os.path.exists(device_path):
-                            log.warning(
-                                'Device path {0} not found and skipped'.format(
-                                    device_path
-                                )
+
+                    if os.path.exists(device_path):
+                        self.fstab.append(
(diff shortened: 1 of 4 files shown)


rjschwei:
LGTM, thanks

PR #169, opened 2020-02-17
Preserve zypper log

### LINKED ISSUE #168: Preserve /var/log/zypper.log
Even if called with the --root option zypper logs to /var/log on the migration live system. These log information should be part of the host for better debugging

jesusbv:
Mount zypper log in the host system.
This Fixes #168

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -101,6 +101,25 @@ def main():
     cloud_register_metadata = os.sep.join(
         [root_path, 'var', 'lib', 'cloudregister']
     )
+    zypper_log_file = os.sep.join(
+        [root_path, 'var', 'log', 'zypper.log']
+    )
+    if os.path.exists(zypper_log_file):
+        try:
+            zypper_host_log_file = zypper_log_file.replace(root_path, '')
+            if not os.path.exists(zypper_host_log_file):
+                with open(zypper_host_log_file, 'w'):
+                    # we bind mount the system zypper log file
+                    # but the mount target does not exist.
+                    # Create it as empty file prior bind mounting
+                    pass
+                Command.run(
+                    ['mount', '--bind', zypper_log_file, zypper_host_log_file]
+                )
+        except Exception as issue:
+            log.warning(
+                'Bind mounting zypper log file failed with: {0}'.format(issue)
+            )
     try:
         # log network info as network-online.target is done at this point
         log_network_details()
--- a/test/unit/units/prepare_test.py
+++ b/test/unit/units/prepare_test.py
@@ -85,9 +85,22 @@ class TestSetupPrepare(object):
         fstab = Mock()
         mock_Fstab.return_value = fstab
         mock_os_listdir.return_value = ['foo', 'bar']
-        mock_os_path_exists.return_value = True
+        mock_os_path_exists.side_effect = [True, True, True, True,
+                                           True, False, True, True]
         mock_is_registered.return_value = True
-        main()
+        mock_Command_run.side_effect = [
+            MagicMock(),
+            Exception('no zypper log'),
+            MagicMock(),
+            MagicMock(),
+            MagicMock(),
+            MagicMock(),
+            MagicMock(),
+            MagicMock(),
+            MagicMock()
+        ]
+        with patch('builtins.open', create=True):
+            main()
         assert mock_shutil_copy.call_args_list == [
             call('/system-root/etc/SUSEConnect', '/etc/SUSEConnect'),
             call(
@@ -110,6 +123,12 @@ class TestSetupPrepare(object):
         assert mock_Command_run.call_args_list == [
             call(
                 ['update-ca-certificates']
+            ),
+            call(
+                [
+                    'mount', '--bind', '/system-root/var/log/zypper.log',
+                    '/var/log/zypper.log'
+                ]
             ),
             call(
                 ['ip', 'a'], raise_on_error=False


suse_migration_services/units/migrate.py
```diff
@@ -40,7 +40,13 @@ def main():
...
     log = logging.getLogger(Defaults.get_migration_log_name())
     root_path = Defaults.get_system_root_path()
-
+    zypper_log = os.sep.join(
+        [root_path, 'var', 'log', 'zypper.log']
+    )
+    log.info('Bind mounting /var/log/zypper.log')
+    Command.run(
+        ['mount', '--bind', zypper_log, '/var/log/zypper.log']
+    )
```
schaefi: The bind mount is nice but it has to check if the destination '/var/log/zypper.log' exists and if not create it. It also should not mount if zypper_log does not exist

Apart from that this is the wrong place for this mounting code. It belongs to units/prepare.py

schaefi: I also forgot to mention that I think the mount should not cause the system to fail. Thus I would use raise_on_error=False in the Command call

schaefi: You could spare the check if zypper_log exists if you use raise_on_error=False, except if you want to add some descriptive error messages on the condition. I leave that up to you

schaefi: I think I would just log the raise message and be done

jesusbv: > The bind mount is nice but it has to check if the destination '/var/log/zypper.log' exists and if not create it. It also should not mount if zypper_log does not exist
> 
> Apart from that this is the wrong place for this mounting code. It belongs to units/prepare.py

I thought to put it in prepare but that code deals with the information about SCC, RMT or SMT.
So I thought it wouldn't be its responsibility. I will update it.

suse_migration_services/units/prepare.py
```diff
@@ -101,6 +101,28 @@ def main():
     cloud_register_metadata = os.sep.join(
         [root_path, 'var', 'lib', 'cloudregister']
     )
+    zypper_log_path = os.sep.join(
+        [root_path, 'var', 'log', 'zypper.log']
+    )
+    if os.path.exists(zypper_log_path):
+        if not os.path.exists('/var/log'):
+            log.info('Creating /var/log')
+            Path.create('/var/log')
```
schaefi: You don't have to check for the presence of /var/log, that will always exist because of the FHS standard. However you have to check on /var/log/zypper.log and create an empty file if it does not exist. The reason is you can't bind mount root_path/var/log/zypper.log to /var/log/zypper.log if this target does not exist. This is the same concept as if you would bind mount a directory. The target must exist

schaefi: Also I think there are too many occurrences of var/log/zypper.log

schaefi:
```python
zypper_log_file = os.sep.join(
    [root_path, 'var', 'log', 'zypper.log']
)
if os.path.exists(zypper_log_file):
    zypper_host_log_file = zypper_log_file.replace(root_path, os.sep)
    if not os.path.exists(zypper_host_log_file):
        with open(zypper_host_log_file, 'w'):
            # we bind mount the system zypper log file but the mount target
            # does not exist. Create it as empty file prior bind mounting
            pass
    
    ...rest of the code
```

suse_migration_services/units/prepare.py
```diff
@@ -101,6 +101,28 @@ def main():
...
+            ['mount', '--bind', zypper_log_path, '/var/log/zypper.log'],
+            raise_on_error=False
+        )
+        if zypper_log.return_code != 0:
+            log.warning(
+                '{0} failed with stderr: {1}, stdout: {2}.'
+                .format(
+                    zypper_log_message, zypper_log.error, zypper_log.output
+                )
+            )
```
schaefi: If you consider to just pass the exception message you can rewrite this to:

```python
try:
    Command.run(
        ['mount', '--bind', zypper_log_file, zypper_host_log_file]
    )
except Exception as issue:
    log.warning(
        'Bind mounting zypper log file failed with: {0}'.format(issue)
    )
```

There is no need to log the command information because we already log all Command.run
information since I merged #160

suse_migration_services/units/prepare.py
jesusbv: Done. 
`replace(root_path, os.sep)` would not work (it would be `//path`) so I changed it.

suse_migration_services/units/prepare.py
```diff
@@ -101,6 +101,24 @@ def main():
     cloud_register_metadata = os.sep.join(
         [root_path, 'var', 'lib', 'cloudregister']
     )
+    zypper_log_path = os.sep.join(
+        [root_path, 'var', 'log', 'zypper.log']
+    )
+    if os.path.exists(zypper_log_path):
```
schaefi: Can we be consistent in the naming, log_file vs. log_path. I vote for zypper_log_file and zypper_host_log_file

suse_migration_services/units/prepare.py
```diff
@@ -101,6 +101,24 @@ def main():
...
+        zypper_host_log_file = zypper_log_path.replace(root_path, '')
+        if not os.path.exists(zypper_host_log_file):
+            with open(zypper_host_log_file, 'w'):
+                # we bind mount the system zypper log file but the mount target
+                # does not exist. Create it as empty file prior bind mounting
+                pass
+    try:
+        Command.run(
```
schaefi: All of the try block belongs inside the former if os.path.exists(zypper_log_path):

suse_migration_services/units/prepare.py
```diff
@@ -101,6 +101,24 @@ def main():
...
+            ['mount', '--bind', zypper_log_path, '/var/log/zypper.log']
```
schaefi: avoid duplication of code: '/var/log/zypper.log' is zypper_host_log_file

suse_migration_services/units/prepare.py
jesusbv: Fixed

suse_migration_services/units/prepare.py
jesusbv: Fixed

suse_migration_services/units/prepare.py
jesusbv: Fixed

PR #170, opened 2020-02-24
State the focused system is the main one

jesusbv:


jesusbv:
> Sorry I have no idea what you are trying to express.

I tried to express that the reason because we dont mount all systems is the migration focus on the main system and not all the system mounted

jesusbv:
> Sorry I have no idea what you are trying to express.

I tried to express that the reason why we dont mount all systems is the migration focus on the main system and not all the system mounted

jesusbv:
@rjschwei does the new changes express the idea better ?

PR #172, opened 2020-03-25
Documentation update

rjschwei:
**Documentation update**
    
* Pull cloud specific caveats into their own section
* Explicitly set the supported migration path for cloud instances
* Minor woring improvements
* Remove section about setting the migration target to make it a hidden feature

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -48,10 +48,10 @@ The migration implementation also supports an upgrade mode that works with
 systems not registered to a repository service. For details,
 see <<Optional Customization of the Upgrade Process>>.
 
-The upgrade to a new major version requires the system to be offline
-during the upgrade to avoid system inconsistencies that may leave the
-system in a state that does not allow recovery. This behavior is implemented
-using a Live Migration Image.
+The upgrade to a new major version requires the system to be migrated to
+be offline during the upgrade to avoid system inconsistencies that may
+leave the system in a state that does not allow recovery. This behavior
+is implemented using a Live Migration Image.
 
 The distribution migration system provides the live image and a startup
 utility named: `run_migration` which reboots the running system into the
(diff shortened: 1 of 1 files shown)


doc/adoc/user_guide.adoc
```diff
@@ -260,3 +254,18 @@ a pointer to the respective log file.
   SUSE Linux Enterprise Server documentation.
 * In systems that contain multiple root file systems on different mount points
   only the root file system mounted on `/` (primary system) will be migrated.
+
+=== Public and Private Cloud Specific
+* Migration initiation is only supported via a reboot. The
+  suse-migration-sle15-activation package must be installed.
```
jesusbv: I think this is confusing with [installation](https://github.com/SUSE/suse-migration-services/blob/master/doc/adoc/user_guide.adoc#3-installation)  where it says Migration can be invoked without `suse-migration-sle15-activation` package, running `run_migration`

doc/adoc/user_guide.adoc
```diff
@@ -155,8 +155,8 @@ all required repositories to be setup correctly.
 
 Specify Migration Product::
 With the SLES15-Migration package an auto detected product is
-already applied. If that detection failed, the setting can be overwritten with one of the
-following product specifications:
+already applied. If that detection failed, the setting can be overwritten
+with a product specifications in the following format:
```
schaefi: Actually from our e-mail conversation we don't want people to specify the product themselves due to the concerns you raised. So what about deleting the documentation in this regard ?

doc/adoc/user_guide.adoc
schaefi: I agree suse-migration-sle15-activation is only required if you plan to activate the migration via reboot. I think what you wanted to express is that in the public cloud there is no other way than reboot to activate it because kexec method is not working in all public clouds.

schaefi: I think we can say that more generically and describe that in a table


```
 CSP     |  reboot     |  run-migration
-----------------------------------------
Azure    |  ok         |    !
Google   |  ok         |    ok
EC2      |  ok         |    !
```

schaefi:
An update process of the docs has been started here:

* https://bugzilla.suse.com/show_bug.cgi?id=1168446

PR #178, opened 2020-08-26
Update SMT cache

jesusbv:
Due to a cache inconsistency, the cache
update needs to be triggered.

This is related to bsc#1173654

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -159,6 +159,12 @@ def main():
                     '/var/lib/cloudregister'
                 ]
             )
+            update_smt_cache = '/usr/sbin/updatesmtcache'
+            if os.path.isfile(update_smt_cache):
+                log.info('Updating SMT cache')
+                Command.run(
+                    [update_smt_cache]
+                )
         system_mount.export(
             Defaults.get_system_mount_info_file()
         )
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/prepare.py
```diff
@@ -159,6 +159,10 @@ def main():
                     '/var/lib/cloudregister'
                 ]
             )
+            log.info('Updating SMT cache')
+            Command.run(
+                ['/usr/sbin/updatesmtcache']
```
schaefi, line 164: ok you are calling this from inside of the live migration system. Does this really work ? where does the script come from ? Can we be sure this script exists in the live migration image ? ...and what exactly does it ?

schaefi (commented):
Looks good to me

But I have concerns if that really works

suse_migration_services/units/prepare.py
jesusbv: Yes, tested that change and migration succeeded. I can add if script exist check.

rjschwei: The script is part of cloud-regionsrv-client. So as long as cloud-regionsrv-client is part of the image description the script will be there. However, a `if os.path.isfile('/usr/sbin/updatesmtcache')` is still prudent, IMHO

schaefi: ok cloud-regionsrv-client provides it and this package is part of the description, good

jesusbv: Done.

schaefi (approved):

PR #179, opened 2020-11-13
Set right root type fs

jesusbv:
When root type is extX, insmod is extX and it should be ext2
that way grub ext2 module supports all extX fs.

This Fixes bsc#1178737

### DIFF
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -17,6 +17,11 @@ migration_iso=$(echo /usr/share/migration-image/*-Migration.*.iso)
 root_device=$(lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | uniq | cut -f1 -d" ")
 root_uuid=$(blkid -s UUID -o value "${root_device}")
 root_type=$(blkid -s TYPE -o value "${root_device}")
+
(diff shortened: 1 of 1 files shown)


grub.d/99_migration
```diff
@@ -17,6 +17,11 @@ migration_iso=$(echo /usr/share/migration-image/*-Migration.*.iso)
 root_device=$(lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | uniq | cut -f1 -d" ")
 root_uuid=$(blkid -s UUID -o value "${root_device}")
 root_type=$(blkid -s TYPE -o value "${root_device}")
+
+if [[ $root_type =~ ^ext[[:digit:]] ]] ; then
```
schaefi, line 21: Just nitpick but for consistency shell variables should be referenced in bracket notation

```bash
if [[ "${root_type}" =~ ^ext[[:digit:]] ]] ; then
```

schaefi (approved):
Looks good :+1: 

just a minor variable access smell

schaefi:
Another minor thing. In the commit message you said "bcs#..." this is imho an unknown reference type. Should be "bsc#..."
I would change this because it might cause trouble when we submit the package to the SUSE namespace and the bug id
will not be recognized.

Thanks

PR #181, opened 2021-02-02
Extend preserved files

### LINKED ISSUE #180: Add /etc/sysconfig/proxy to the network environment
Some customers, who for whatever reason are unable to use transparent proxies in their networks, still use /etc/sysconfig/proxy to define system wide proxies. I understand the Distribution Migration System does not carry this file across, so the migration always fails because the migration system cannot find scc.suse.com without the proxy defined. 

Note that my suggestion is only a solution for what I expect to be the majority of cases - where customers set system-wide proxies in /etc/sysconfig/proxy AND do not require a password for the proxy. Some customers use environment variables rather than /etc/sysconfig/proxy and some do also require passwords for their proxies, for example stored in .curlrc.

jesusbv:
Add option to preserve /etc/sysconfig/proxy file.

This fixes #180

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -158,11 +158,21 @@ Preserve custom data file(s) e.g. udev rules from the system
 to be migrated into the upgrade live system and make sure
 they will become effective.
 
+Under preserve section, there are two subsections: rules and static.
+The difference between 'rules' and 'static' sections is that files preserved as
+udev rules will also make the DMS to reload udev and its rules to make the new
+rule set effective, while the files in the static section are copied with no
+further action.
+
 [listing]
 preserve:
   rules:
     - /etc/udev/rules.d/a.rules
     - /etc/udev/rules.d/b.rules
+  static:
+    - /etc/sysconfig/proxy
+    - /path/to/be/preserved/file
+
 
 [NOTE]
 udev rules that require custom drivers will not have the desired effect
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -95,19 +95,17 @@ class MigrationConfig:
 
         return migration_product
 
-    def get_preserve_udev_rules_list(self):
+    def get_preserve_info(self):
         """
-        Returns list of udev rule file(s)
+        Return dictionary of the different list of files to be preserved.
 
-        The returned file list is used to make those rule files
+        The returned lists, in the dictionary, are used to make those files
         available to the migration live system such that the
-        same behavior the rules apply on the system to be
+        same behavior the config applies on the system to be
         upgraded also applies to the live migration system
         while the upgrade runs
         """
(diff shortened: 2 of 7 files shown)


schaefi:
@jesusbv Thanks the implementation is good and will work. However I think we should make it a bit more flexible. What if another user wants to preserve some other file from /etc, then we need to add more code like this and have to run another submission which is an expensive/painful process as you know ;)

I was thinking about something like:

```yaml
static:
  - /etc/sysconfig/proxy
```

That way we would also support any static file to be preserved by a simple copy process.

Thoughts ?

jesusbv:
> ```yaml
> static:
>   - /etc/sysconfig/proxy
> ```
> 
> That way we would also support any static file to be preserved by a simple copy process.
> 
> Thoughts ?

I  like the idea but I limited it to `proxy` because I wasn't sure about the scope of the files we wanted to include.
Will update it.

schaefi:
yeah we can't be sure what people want/need to preserve more in the future but I think allowing this as a general feature is not a bad idea. Thanks for updating :+1:

doc/adoc/user_guide.adoc
```diff
@@ -163,6 +163,9 @@ preserve:
   rules:
     - /etc/udev/rules.d/a.rules
     - /etc/udev/rules.d/b.rules
+  static:
+    - /etc/sysconfig/proxy
+    - /path/to/be/preserved/file
```
schaefi, line 168: There should be some words about the difference between ```rules:``` and ```static:``` I think it would be good to mention that files preserved as udev rules will also make the DMS to reload udev and its rules to make the new rule set effective, whereas the files in the static section are just copied without any further action

schaefi (approved):
Looks great :+1: 

just a bit more info in the docs please :)

doc/adoc/user_guide.adoc
jesusbv: Done. Once all is good, I will squash the commits into one.

rjschwei:
Documentation will also need to be updated.

doc/adoc/user_guide.adoc
schaefi: This is great now :+1: 

Just go ahead and merge after squash. NOTE: Please remember we need a bugzilla number for this to avoid trouble with maintenance. If there is no bugzilla report we can refer this to, please add a bug in bugzilla and use the number here in the commit message as "bsc#NR" Thanks much

schaefi:
@jesusbv  Just go ahead and merge after squash. NOTE: Please remember we need a bugzilla number for this to avoid trouble with maintenance. If there is no bugzilla report we can refer this to, please add a bug in bugzilla and use the number here in the commit message as "bsc#NR" Thanks much

schaefi (approved):
perfect, see my comment on the commit message the bugzilla reference for the submisson. Thanks

PR #183, opened 2021-02-05
Allow non standard paths

jesusbv:
Create dir if not exists before copying


schaefi:
ok I know I said we don't need to check for ```os.path.exists``` prior calling mkdir but I think it would be good if we do that because I forgot the udev rules are handled in the same copy path and they all live in the same directory, so we will repeat ourselves quite often eventually. Could you add the check ?

Thanks

schaefi (approved):
great :+1:

PR #184, opened 2021-02-05
Avoid deprecation warning

jesusbv:


schaefi:
I'm missing a description

I assume you got a deprecation warning from what (tox) when running on TW right ? Because I don't see it when running on 15.2

jesusbv:
> @jesusbv I'm missing a description
> 
> I assume you got a deprecation warning from what (tox) when running on TW right ? Because I don't see it when running on 15.2

Yes, when running `make check`

```bash
WARNING: test command found but not installed in testenv
  cmd: /usr/bin/bash
  env: /home/jesusbv/SandBox/suse-migration-services/.tox/3
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above!
```

Also, when running `tox` or `make test`, so will fix those too in this PR and update this fix.

schaefi:
ok great, thanks for catching this one. If you are done feel free to merge. It does not affect the package submission to the SUSE namespace if I got it right this is only affecting the project setup. Thanks

schaefi (approved):
:+1:

PR #186, opened 2021-02-19
Do not copy /etc/resolv.conf if empty

### LINKED ISSUE #185: More robust use of name resolution setup
The DMS uses the hosts resolv.conf file to implement the DNS setup. However depending on the host configuration the resolv.conf information is cleared out on reboot of the machine. In this case the DMS system should not copy the empty resolv.conf information but rather stick with the DHCP data the DMS received at boot time.

jesusbv:
This Fixes #185
This references bsc#1182234

### DIFF
--- a/suse_migration_services/units/setup_host_network.py
+++ b/suse_migration_services/units/setup_host_network.py
@@ -51,10 +51,13 @@ def main():
         raise DistMigrationNameResolverException(
             'Could not find {0} on migration host'.format(resolv_conf)
         )
-
-    shutil.copy(
-        resolv_conf, '/etc/resolv.conf'
-    )
+    if has_host_resolv_setup(resolv_conf):
+        log.info('Copying {}'.format(resolv_conf))
+        shutil.copy(
+            resolv_conf, '/etc/resolv.conf'
+        )
+    else:
+        log.info('Empty {}, continuing without copying it'.format(resolv_conf))
 
     sysconfig_network_providers = os.sep.join(
         [root_path, 'etc', 'sysconfig', 'network', 'providers']
@@ -139,3 +142,12 @@ def log_network_details():
                 ).output
             )
         )
+
+
+def has_host_resolv_setup(resolv_conf_path):
+    with open(resolv_conf_path, 'r') as resolv:
+        for line in resolv:
+            # check there is useful information in the remaining lines
+            if line.startswith('search') or line.startswith('nameserver'):
+                return True
+    return False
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/setup_host_network.py
```diff
@@ -139,3 +142,18 @@ def log_network_details():
...
+        lines = filter(
+            lambda line: not line.startswith('#') and not line.startswith(os.linesep),
+            resolv.readlines()
+        )
+    for line in lines:
+        # check there is useful information in the remaining lines
+        if line.startswith('search') or line.startswith('nameserver'):
+            return False
+
+    return True
```
schaefi, line 159: Maybe we could write this a bit simpler ? like

```python
with open(resolv_conf_path, 'r') as resolv:
    for line in resolv:
        if line.startswith('search') or line.startswith('nameserver'):
            # looks like there is some effective resolver setup
            return False
return True
```

maybe it's also easier to read if the function is named "def have_host_resolv_setup" or similar

Thoughts ?

test/unit/units/setup_host_network_test.py
```diff
@@ -91,3 +98,80 @@ def test_main(
...
+
+    @patch('suse_migration_services.units.setup_host_network.is_resolv_empty')
+    @patch('suse_migration_services.logger.Logger.setup')
+    @patch('suse_migration_services.command.Command.run')
+    @patch('suse_migration_services.units.setup_host_network.Fstab')
+    @patch('os.path.exists')
+    @patch('shutil.copy')
+    @patch('glob.glob')
+    @patch('os.path.isfile')
+    def test_main_no_copy(
```
schaefi, line 110: I think the intention of this test is to check that etc/resolv conf is not copied. I wonder about all the other assertions for fstab and so on

schaefi: I would have expected something like ```assert not mock_shutil_copy.called``` or ```assert call... not in mock_shutil_copy.call_args_list```. Something that is more explicit to what the test is good for. Does this makes sense ?

schaefi (changes requested):
Thanks for looking into this. I think the logic makes good sense, but I think it can be reached with less code. Also think about unit tests as code, less code is better and easier to read, especially in tests. Hope that makes sense ?

test/unit/units/setup_host_network_test.py
schaefi: All the other conditions of main are part of other test cases. So I think it's fine if this test just focuses on no copy happened in the most easy way to get that condition

jesusbv: Fixed.

suse_migration_services/units/setup_host_network.py
jesusbv: Done.

schaefi (approved):
yeah, that looks really nice now :+1:

PR #187, opened 2021-02-24
Fixed umount inconsistency

schaefi:
There is the mount-system service which mounts the system taking fstab into account. If it fails for some reason it also umounts all parts of the system. However in the sequence of the error condition there is also the grub-setup service which uninstalls the migration package and restores the grub config such that the subsequent boot process does no longer boot into the migration system. With the error condition from above the grub-setup service has no chance to perform its jobs because all of system-root has been umounted already. This commit makes sure that mount-service only registers the mount paths but does not umount them in case of a failure. This allows grub-setup to do its job and also allows the reboot service to umount registered devices prior reboot.

This is related to bsc#1182520

### DIFF
--- a/suse_migration_services/units/mount_system.py
+++ b/suse_migration_services/units/mount_system.py
@@ -187,8 +187,6 @@ def mount_system(root_path, fstab):
         log.error(
             'Mounting system for upgrade failed with {0}'.format(issue)
         )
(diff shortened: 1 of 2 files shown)


jesusbv (approved):
Good catch

PR #190, opened 2021-03-02
Added sort by path order for fstab handling

### LINKED ISSUE #188: Apply sort order by path when reading fstab
In reference to:

* https://bugzilla.suse.com/show_bug.cgi?id=1182520

it seems systemd's fstab generator does some ordering such that people no longer have to care for the
chronological order of entries in the good old ```/etc/fstab``` file. If this is the case we will constantly run into issues
on the DMS end as we expect the fstab to be consistent.

I still think the DMS is not in the responsibility to fix systems, but I would be ok adding a path sorting in here to make
customers happy given the needinfo request on the mentioned bug really confirms that systemd does this ordering

schaefi:
Don't trust the order of entries in the fstab to be correct.
Instead add a canonical path order to the fstab entries
when processed by the distribution migration system.
This Fixes #188 and Fixes bsc#1182520

### DIFF
--- a/suse_migration_services/fstab.py
+++ b/suse_migration_services/fstab.py
@@ -17,7 +17,9 @@
 #
 import logging
 import os
-from collections import namedtuple
+from collections import (
+    namedtuple, OrderedDict
+)
 
 # project
 from suse_migration_services.defaults import Defaults
@@ -104,7 +106,7 @@ class Fstab:
         :param string filename: path to file name
         """
         with open(filename, 'w') as fstab:
-            for entry in self.fstab:
+            for entry in self._get_canonical_mount_list():
                 fstab.write(
                     '{0} {1} {2} {3} 0 0{4}'.format(
                         entry.device, entry.mountpoint,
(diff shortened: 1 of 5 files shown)


suse_migration_services/fstab.py
```diff
@@ -114,4 +116,53 @@ def export(self, filename):
...
+    @staticmethod
+    def _sort_by_hierarchy(path_list):
+        """
+        Sort given list of path names by their hierachy in the tree
+
+        Example:
+
+        .. code:: python
+
+            result = sort_by_hierarchy(['/var/lib', '/var'])
```
jesusbv, line 146: a bit nitpicking but it would be nice to have the call and the return value
```python
sort_by_hierarchy(['/var/lib', '/var'])
[output_sorted]
```

jesusbv (approved):
LGTM

suse_migration_services/fstab.py
schaefi: Could you be more precise what exactly you expect in the Example code ? Please remember it has to be valid python otherwise the doc generator will fail. So currently we have this:

```
Example:

.. code:: python

     result = sort_by_hierarchy(['/var/lib', '/var'])
```

How would you like to change this to be more obvious ?

Thanks

PR #193, opened 2021-04-12
Avoid multiple mount attempts

### LINKED ISSUE #191: DMS fails due to multiple mount attempt of special filesystem
This issue has come up numerous types with customers that include special filesystems explicitly in their /etc/fstab.  For instance, a customer fstab may look like:

```
/dev/hda1       /       ext4    defaults 1 1
tmpfs   /run    tmpfs   noauto 0 0
sysfs   /sys    sysfs   noauto 0 0
debugfs /sys/kernel/debug       debugfs noauto 0 0
proc    /proc   proc    defaults 0 0
devpts  /dev/pts        devpts  mode=0620,gid=5 0 0
```

During the mount_system, the DMS will itself mount devtmpfs, proc, sysfs, etc. But since the customer fstab includes the special filesystems already, the mount_system will fail and as a result, the migration aborts.

Proposing changing the fstab.py to not only ignore swap but all special filesystems (or at least the special filesystems the DMS explicitly mounts),  as it is unknown what the user may have in their /etc/fstab.

Instead of:
`if fstype != 'swap':`

Propose:
`if fstype not in ['swap','tmpfs','sysfs','debugfs','proc','devtmpfs','dev','devpts']:`

jesusbv:
- Fstab could contain one of the mount points
  DMS mount explicitly, ending up in an exception.
  To avoid this, DMS does not mount the explicit
  mount points if they are in fstab.

This Fixes #191 and bcs#1184278

### DIFF
--- a/suse_migration_services/units/mount_system.py
+++ b/suse_migration_services/units/mount_system.py
@@ -139,50 +139,36 @@ def mount_system(root_path, fstab):
     log = logging.getLogger(Defaults.get_migration_log_name())
     log.info('Mount system in {0}'.format(root_path))
     system_mount = Fstab()
+    explicit_mount_points = {
+        'devtmpfs': os.sep.join([root_path, 'dev']),
+        'proc': os.sep.join([root_path, 'proc']),
+        'sysfs': os.sep.join([root_path, 'sys'])
+    }
     try:
         for fstab_entry in fstab.get_devices():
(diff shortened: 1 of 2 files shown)


rjschwei (commented):
LGTM

schaefi (approved):
perfect :+1:

PR #194, opened 2021-04-12
Use ismount

jesusbv:
- Change the command "mount -q $mountpoint"
  for Python method "os.path.ismount($mountpoint)"


schaefi (approved):
yep, as discussed this is much nicer. I'm also ok with the limitation on bind mounts

PR #196, opened 2021-04-12
Debug lvm issue

schaefi:
Added kernel-firmware to support bare metal better
    
For the DMS to boot on bare metal system we should install the kernel-firmware package. If not present certain systems like HP with Mellanox driver fails to boot. This is related to bsc#1182520


schaefi:
NO MERGE !

PR #202, opened 2021-04-21
Fixed import of certificates

schaefi:
Certificates can exist on several places. This commit
makes sure we lookup certificates at the following
places:

* /usr/share/pki/trust/anchors
* /etc/pki/trust/anchors

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -62,9 +62,10 @@ def main():
     hosts_setup = os.sep.join(
         [root_path, 'etc', 'hosts']
     )
-    trust_anchors = os.sep.join(
-        [root_path, 'usr', 'share', 'pki', 'trust', 'anchors']
-    )
+    trust_anchors = [
+        '/usr/share/pki/trust/anchors/',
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/prepare.py
```diff
@@ -81,21 +82,24 @@ def main():
...
-                    '/usr/share/pki/trust/anchors/'
+    for trust_anchor in trust_anchors:
+        root_trust_anchor = os.path.normpath(
+            os.sep.join([root_path, trust_anchor])
+        )
+        if os.path.isdir(root_trust_anchor):
+            certificates = os.listdir(root_trust_anchor)
+            if certificates:
+                Path.create(trust_anchor)
+                for cert in certificates:
```
jesusbv: thoughts to use `copytree` instead of `copy` (and the for loop)
```python
log.info(f'Importing certificates: {", ".join(certificates)}')
shutil.copytree(
    root_trust_anchor,
    trust_anchor,
    dirs_exist_ok=True
)
```

jesusbv (approved):
LGTM

Issue #197, opened 2021-04-14
Failed to migrate using DMS


jaawasth:
Hi Marcus, 

As a continuation to our chat on the issue https://github.com/SUSE-Enceladus/azure-li-services/issues/266

filing a bug with this project.
please see above issue on MS images for more details.

Thanks !!

schaefi:
Hi Jai, would be great to start a debugging session about the DMS issue with the test system you have setup. 
I can arrange for a meeting on  Fri 16th, Mon 19th, Thu 21st between 3-7pm o'clock CEST. If you sent a teams event via ms@suse.com it will land in my calendar

Thanks

schaefi:
@jaawasth  Thanks Jai for the debugging sessions. We have been able to fix all issues and will rebuild the DMS including the fixes. For customers to access the DMS packages (SLES15-Migration and suse-migration-sle15-activation) we need to run a new release to the SUSE namespace.

As this will take time I'm trying to speedup this process with maintenance.

I will do the submissions after we have clarity everything works as expected. Therefore I will provide the packages for the submission on the jump host from which you fetched the debugging versions too.

I'll let you know when they are there and you can do a final test without us attending, agreed ?
On positive feedback I'll submit to SLES

Thanks

jaawasth:
@schaefi, yes, this sounds good.

Thanks !!

schaefi:
@jaawasth ok the hopefully final version of the updated DMS packages are available on the jump host we used for debugging.
Please fetch them and give it a try:

```
$ rpm -e suse-migration-sle15-activation SLES15-Migration
$ rpm -Uhv suse-migration-sle15-activation-2.0.23-6.21.1.noarch.rpm SLES15-Migration-2.0.23-6.x86_64.rpm
$ reboot
```

Let us know if everything works. I'll keep this open until your feedback arrives.

Again thanks much for walking with us through the system :+1:

jaawasth:
@schaefi , sure thanks, i'll be able to give it a try later today and let you know how it goes.

Thanks !!

jaawasth:
@schaefi , the migration rpms worked fine and the system got upgraded, any particular file you want from the system to ensure completeness ?

schaefi:
Great new, thanks for the feedback. I think I don't need further files. With that information I can now start the release of the DMS in SLES. Thanks this is a big step forward

jaawasth:
[...]

PR #204, opened 2021-04-23
Cracefully handle cert copy exceptions

schaefi:
If the copy process of a certificate failed this should not
cause the entire migration process to stop. This commit
handles copy errors from the cert chain and turns them into
a log message. If in the subsequent chain of tasks the
migration failed because of missing certificates we will
see that in the log and in the attempt to access the repos
which is better than the python stacktrace on an unhandled
FileNotFoundError exception. Retated to Issue #197


jesusbv (approved):

jesusbv (approved):
This should work better

PR #206, opened 2021-05-05
Update option settings in zypper dup mode

schaefi:
When using the DMS in zypper dup mode the option
settings --allow-downgrade and --no-confirm are handy.
Even though I think with the global non-interactive
option set it should not be required to set no-confirm
in addition but the worst are questions being asked
so better add it. With allow-downgrade set the ability
to downgrade packages is given. With zypper dup the
repo setup is user defined custom set and therefore
downgrades should not be off the table

### DIFF
--- a/suse_migration_services/units/migrate.py
+++ b/suse_migration_services/units/migrate.py
@@ -74,6 +74,7 @@ def main():
                     '--auto-agree-with-licenses',
                     '--allow-vendor-change',
                     '--replacefiles',
+                    '--allow-downgrade '
                     '&>>', Defaults.get_migration_log_file()
                 ]
             )
--- a/test/unit/units/migrate_test.py
+++ b/test/unit/units/migrate_test.py
@@ -133,6 +133,7 @@ class TestMigration(object):
                 '--auto-agree-with-licenses '
                 '--allow-vendor-change '
                 '--replacefiles '
+                '--allow-downgrade '
                 '&>> /system-root/var/log/distro_migration.log'
             ], raise_on_error=False
         )


suse_migration_services/units/migrate.py
```diff
@@ -74,6 +74,8 @@ def main():
                     '--auto-agree-with-licenses',
                     '--allow-vendor-change',
                     '--replacefiles',
+                    '--allow-downgrade',
+                    '--no-confirm '
```
jesusbv, line 78: Agree, with `--no-interactive`, the flag `--no-confirm` is not required

```doc
-y, --no-confirm
     Don’t require user interaction. Alias for the --non-interactive global option
```

schaefi: yeah but does not hurt right ? I really cannot put my hand on fire if --no-interactive prevents all interaction especially if a sub command offers an extra option for it. So my vote is to use both to stay super safe. Thoughts ?

rjschwei: I think having the redundancy of the option, i.e. `--non-interactive` and `--no-confirm` is confusing. My $0.02

jesusbv: I agree with Robert, I read it and think about it this way: "if the option is there is because it adds/does something that is not done/added by the previous options"

AFAIK, from the documentation, `--no-confirm` is a flag for subcommands while `--no-interactive` means the same but it is global

```
-n, --non-interactive
           Switches to non-interactive mode. In this mode zypper doesn’t ask user to type answers to various prompts,
           but uses default answers automatically. Those default answers also depend on other options like
           --no-gpg-checks or --ignore-unknown.
```

schaefi: ok, will remove it then

schaefi: done

jesusbv (approved):
LGTM

PR #207, opened 2021-05-05
Add pre checks

### LINKED ISSUE #205: Add pre-checks prior starting the migration
Some conditions e.g the one explained in #197 could be identified easily prior running the migration process.
If it's clear the migration will fail and we can easily detect this situation before reboot or run_migration it would be good to add this type of checks.

I suggest the following checks:

* Lookup /etc/zypp/repos.d for locations that are not remote (e.g CD mount)
* Lookup if any of the filesystems in fstab are using LUKS encryption

Please add to the list if you can think of more pre-conditions we could check easily

jesusbv:
- Check no remote repositories
- Check filesystem has LUKS encryption

This Fixes #205

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,23 @@ sle15_activation: check
 	cat package/suse-migration-sle15-activation-spec-template \
 		| sed -e s'@%%VERSION@${version}@' \
 		> dist/suse-migration-sle15-activation.spec
-	tar -czf dist/suse-migration-sle15-activation.tar.gz grub.d
+	# include grub.d dir in MANIFEST
+	sed -ie s'@prune grub.d@graft grub.d@' MANIFEST.in
+	mv MANIFEST.ine MANIFEST.in
+	python3 setup.py sdist
+	# create tarball with grub.d + suse-migration-services
+	mv dist/suse_migration_services-*.tar.gz  dist/suse-migration-sle15-activation.tar.gz
+	# restore MANIFEST.ini
+	git checkout MANIFEST.in
+	# check MANIFEST.in has prune grub.d
+	# raise error if not
+	$(eval current_value = $(shell grep "prune grub.d" MANIFEST.in))
+	$(eval expected_value = "prune grub.d")
+
+	@if [ "$(current_value)" != $(expected_value) ]; then\
+	   echo "MANIFEST.in does not have '$(expected_value)'";\
+           exit 1;\
+	fi
 
 .PHONY: test
 test:
--- a/package/suse-migration-sle15-activation-spec-template
+++ b/package/suse-migration-sle15-activation-spec-template
@@ -45,7 +45,13 @@ next reboot of the machine.
 install -D -m 755 99_migration \
     %{buildroot}/etc/grub.d/99_migration
 
+python3 setup.py install --prefix=%{_prefix} --root=%{buildroot}
+mv %{buildroot}/%{_bindir}/suse-migration-pre-checks /tmp/
+rm %{buildroot}/%{_bindir}/*
+mv /tmp/suse-migration-pre-checks %{buildroot}/%{_bindir}/
+
 %post
+suse-migration-pre-checks
 /usr/sbin/grub2-mkconfig -o /boot/grub2/grub.cfg
 
 %postun
@@ -55,5 +61,7 @@ install -D -m 755 99_migration \
 %defattr(-,root,root,-)
 %dir /etc/grub.d
 /etc/grub.d/*
+%{python3_sitelib}/*
+%{_bindir}/suse-migration-pre-checks
 
 %changelog
--- a/suse_migration_services/prechecks/pre_checks.py
+++ b/suse_migration_services/prechecks/pre_checks.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 SUSE Linux LLC.  All rights reserved.
+# Copyright (c) 2021 SUSE Linux LLC.  All rights reserved.
 #
 # This file is part of suse-migration-services.
 #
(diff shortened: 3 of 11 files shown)


suse_migration_services/prechecks/pre_checks.py
```diff
@@ -0,0 +1,84 @@
...
+
+# project
+from suse_migration_services.fstab import Fstab
+from suse_migration_services.command import Command
+
+
+def check_remote_repos():
+    repos_path = '/etc/zypp/repos.d'
+    no_remote_repos = []
+    print("PEPE")
```
schaefi, line 29: print PEPE :-)) ??

suse_migration_services/prechecks/pre_checks.py
```diff
@@ -0,0 +1,84 @@
...
+    if os.path.exists(repos_path):
+        repos_list = os.listdir(repos_path)
+        config = configparser.RawConfigParser()
+        if not repos_list:
+            print('No repositories in {}'.format(repos_path))
```
schaefi, line 34: Any reason why we should not use the python logging facility ?

schaefi (changes requested):
Really like it, a good start :+1:

schaefi:
One thing that came to my mind is that all code for checks is in one file ```pre_checks.py``` I think we should have one file for a test each and have pre_checks.py just import and call them. Thoughts ?

package/suse-migration-sle15-activation-spec-template
```diff
@@ -46,6 +46,7 @@ install -D -m 755 99_migration \
     %{buildroot}/etc/grub.d/99_migration
 
 %post
+suse-migration-pre-checks
```
schaefi, line 49: I think this will not be enough because calling the tool in %post creates a dependency to the suse-migration-services package and suse-migration-sle15-activation has no dependency to it. What we need is a sub-package e.g ```suse-migration-pre-checks``` which falls out of suse-migration-services. That sub package must be submitted to the SUSE namespace to SLE12-SP4 and then we need a ```Requires: suse-migration-pre-checks``` in addition to the %post change here.

Makes sense ?

suse_migration_services/prechecks/pre_checks.py
jesusbv: Given the checks should run before booting, more likely after installation, I thought it would be ok to print it in the console

package/suse-migration-sle15-activation-spec-template
jesusbv: Yes, I was not sure what was the best path/approach

I have added the changes for `suse-migration-pre-chekcs` package

Makefile
```diff
@@ -45,6 +45,22 @@ sle15_activation: check
...
+		--from package/suse-migration-pre-checks-spec-template --utc > \
+		dist/suse-migration-sle15-activation.changes
+	helper/update_changelog.py \
+		--file package/suse-migration-pre-checks.changes >> \
+		dist/suse-migration-pre-checks.changes
+	# update package version in spec file
+	cat package/suse-migration-pre-checks-spec-template \
+		| sed -e s'@%%VERSION@${version}@' \
+		> dist/suse-migration-pre-checks.spec
+	tar -czf dist/suse-migration-pre-checks.tar.gz suse_migration_services/prechecks
```
schaefi, line 62: I'm afraid this is not going to work that way. You are trying to rip-out parts of the python namespace "suse_migration_services" into an extra package. This will fail as soon as the pre-checks code uses anything else from this namespace (for example the logging)

I think the approach with a complete new spec file should be revised. My idea was to use the same source tarball and spec from ```suse-migration-services-spec-template``` and create a sub-package. The sub-package can package the full sources but only the pre-checks entry points as callable tools. This will end up in a package conflict between suse-migration-services and suse-migration-pre-checks because they will partially provide the same files. But this conflict is not an issue and actually also a good thing. By no means the pre-checks will be installed with the migration-services on the same host. 

So I think instead of touching the Makefile just touch the ```suse-migration-services-spec-template``` and create a sub-package definition in there

schaefi (changes requested):
See my comments on the spec file approach

Makefile
jesusbv: Code has been updated after our conversation.

I wonder if we want to have pre-checks available for `run_migration` as well
it seems to me currently that wouldn't be possible, IMHO it should be possible.

Thoughts ?

package/suse-migration-services-spec-template
```diff
@@ -89,6 +89,10 @@ install -D -m 644 systemd/suse-migration-ssh-keys.service \
 install -D -m 644 systemd/suse-migration-console-log.service \
     %{buildroot}%{_unitdir}/suse-migration-console-log.service
 
+
+%post
+suse-migration-pre-checks
+
```
schaefi, line 95: The %post of the suse-migration-services should not run the pre-checks, only the %post of the activation package should do it. Remember only the activation package is installed on the host to upgrade. Only there it makes sense to run pre-checks

schaefi: apart from the above the spec changes looks now really good

jesusbv: Done

Makefile
schaefi: This is a bit of a problem. If customer decides to go with the real reboot mode, the package suse-migration-sle15-activation gets installed and the pre-checks runs as part of %post. So all good in this case.

If the customer decides to go with run_migration, no activation package will be installed because run-migration is provided with the package that contains the migration live image (SLE15-Migration). In this case no pre-checks are available

If the customer installs suse-migration-sle15-activation for the reason to get the pre-checks, there will also be the grub menu entry which is unwanted but does not hurt.

For the moment we could add some information in the documentation that says to run the pre-checks, install the
activation package. No matter how the migration is started, the activation and also migration packages will be uninstalled by the migration code itself (except this stage is not reached by the DMS)

Another option would be to split the suse-migration-sle15-activation package into two packages:

* suse-migration-sle15-activation 
* suse-migration-sle15-pre-checks

The activation package would require the pre-checks, such that %post works. Customers using run_migration can install only the pre-checks package and manually call them. This needs to be documented. And a new pre-checks package must be published through the SUSE process

No matter what we do here, this should be done in a next step, not in this request.

Thoughts ?

Makefile
```diff
@@ -43,7 +43,23 @@ sle15_activation: check
 	cat package/suse-migration-sle15-activation-spec-template \
 		| sed -e s'@%%VERSION@${version}@' \
 		> dist/suse-migration-sle15-activation.spec
-	tar -czf dist/suse-migration-sle15-activation.tar.gz grub.d
+        # include grub.d dir in MANIFEST
+	sed -ie s'@prune grub.d@graft grub.d@' MANIFEST.in
```
schaefi, line 47: There seems to be space vs. tab issues. Please use tabs exclusively in the Makefile. Thanks

Makefile
schaefi: I have added #209 to tackle this one

Makefile
jesusbv: Left some comments without tabs.Fixed.

schaefi (approved):
This looks great now :+1: Thanks much

schaefi:
@jesusbv can you squash the commits into a good set and then we are ready to merge I think

PR #212, opened 2021-07-21
Use download-in-advance mode for zypper dup

schaefi:
If the DMS was configured to use 'zypper dup' we should
download packages in advance for more stability. The reason
is that zypper dup always starts with deleting packages and
then continues with installing packages. If at install time
a package cannot be downloaded for some reason the zypper
process ends and leaves the system in an inconsistent state


jesusbv (approved):
LGTM

schaefi:
[DMS  - SUMA Integration](https://trello.com/c/yIyjYssw/1743-dms-suma-integration)

PR #213, opened 2021-07-21
Update to new SUSE stylesheets + fix a small typo

sknorr:


schaefi:
Thanks looks good :+1:

PR #217, opened 2021-07-28
Edit message for pre checks repos

### LINKED ISSUE #214: These repositories locations may be an issue when migrating...
The message comes from the pre-checks and I think it does not tell the user much. We advise users to check the reported repos prior migration but I think users will not know what they should check

The message should be more clear imho

jesusbv:
- Add a more informative message if
  problematic repos are found

This Fixes #214

### DIFF
--- a/suse_migration_services/prechecks/repos.py
+++ b/suse_migration_services/prechecks/repos.py
@@ -33,6 +33,11 @@ def remote_repos():
                             no_remote_repos.append(repo_info['baseurl'])
 
     if no_remote_repos:
+        zypper_cmd = 'zypper repos --url'
         log.warning(
-            'These repositories locations may be an issue when migrating: "{}". '
-            'Please check before migration starts'.format(','.join(no_remote_repos)))
+            'The following repositories may cause the migration to fail, as they '
+            'may not be available during the migration: %s\nPlease, check those '
+            'before starting the migration\nTo see all the repositories and '
+            'their urls, you can run "%s"\n',
+            ','.join(no_remote_repos), zypper_cmd,
+        )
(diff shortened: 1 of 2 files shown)


suse_migration_services/prechecks/repos.py
```diff
@@ -33,6 +33,11 @@ def remote_repos():
                             no_remote_repos.append(repo_info['baseurl'])
 
     if no_remote_repos:
+        zypper_cmd = 'zypper repos --url'
```
schaefi, line 36: I haven't look in more detail, but does this mean only in case there are no remote repos we show the error message ? What if there are 2 remote repos and 1 local repo, will the check apply and warn about the 1 local repo to be an issue ?

schaefi (commented):
Ended up with a question... haven't looked deeper to the code though

suse_migration_services/prechecks/repos.py
jesusbv: No, it means _if there are any remote repos, those are mentioned_

```python
for section in repo_section.keys():
                repo_info = dict(config.items(section))
                if repo_info:
                    if not repo_info['baseurl'].startswith('http') and not repo_info['baseurl'].startswith('ftp'):
                        if repo_info['baseurl'] not in no_remote_repos:
                            no_remote_repos.append(repo_info['baseurl'])
```

loop through the repos
check repo is not remote
add it if it is

I can add that case in the tests, in this PR no problem

schaefi: ok, then I just got confused by the name. All fine, thanks

schaefi (approved):
Looks good :+1: 

conflict resolution needed because I merged the other open PRs

schaefi (approved):

PR #220, opened 2021-08-09
Add valid remote repo prefix

jesusbv:
- Add plugin:/susecloud prefix for repositories checking
- Remove extra Logger definitions to avoid duplicate entries

### DIFF
--- a/suse_migration_services/prechecks/repos.py
+++ b/suse_migration_services/prechecks/repos.py
@@ -3,11 +3,9 @@ import configparser
 import logging
 
 from suse_migration_services.defaults import Defaults
-from suse_migration_services.logger import Logger
 
 
 def remote_repos():
-    Logger.setup(system_root=False)
     log = logging.getLogger(Defaults.get_migration_log_name())
     repos_path = '/etc/zypp/repos.d'
     no_remote_repos = []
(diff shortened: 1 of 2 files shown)


suse_migration_services/prechecks/repos.py
```diff
@@ -19,6 +17,7 @@ def remote_repos():
             log.error('No repositories in {}'.format(repos_path))
             return
 
+        remote_prefixes = ('http', 'ftp', 'plugin:/susecloud')
```
rjschwei, line 20: This is by definition incomplete. There is nothing special about `plugin:/susecloud`. Anyone can write a URL resolver plugin for zypper and that plugin can resolve the prefix into a remote repo (http or ftp access) or a local repo. As such it would probably be good to have a comment here or find a name for the variable that is more expressive of it's intend.

schaefi: I agree, some naming that points to the fact that we accept repos of this type as "eligible to be used in a migration" would imho express the intend better

jesusbv: Updated, between `migration_accepted` or `migration_safe` I chose the former

schaefi (approved):
Looks good, good catch on the logger :+1:

PR #222, opened 2021-08-13
Add env information to the log

jesusbv:
- This information is helpful if a proxy is set (see bsc#1188594)


suse_migration_services/units/post_mount_system.py
```diff
@@ -67,3 +67,31 @@ def main():
...
+    for _, preserve_files in preserve_info.items():
+        if Defaults.get_proxy_path() in preserve_files:
+            with open(Defaults.get_proxy_path(), 'r') as proxy_file:
+                for line in proxy_file.readlines():
+                    if not (line.startswith('#') or line.startswith(os.linesep)):
+                        key_value = line.lower(). \
+                            replace(os.linesep, ''). \
+                            replace('"', ''). \
+                            split('=')
+                        proxy_env.update(dict([key_value]))
```
rjschwei, line 83: This creates a "non" compatible environment. For example when configuring a proxy then the local settings are exported into the shell for the "SOCKS" and "NO" proxy settings are both available in upper and lower case, i.e.
`NO_PROXY` and `no_proxy` as well as `SOCKS_PROXY` as well as `socks_proxy` both exist. This implementation forces everything to lower case, which is correct for https and http.

Given that the DMS only cares about https, http, and there may be ftp repos for which lower case only is sufficient at the very least we should comment this code to set the proper scope, i.e. that in DMS we only care about the 3 protocols for which lower case is sufficient.

test/unit/units/post_mount_system_test.py
```diff
@@ -34,3 +45,22 @@ def test_main(
             call(['udevadm', 'trigger', '--type=subsystems', '--action=add']),
             call(['udevadm', 'trigger', '--type=devices', '--action=add'])
         ]
+
+    @patch.dict(os.environ, {'foo': 'bar', 'a': 'b'}, clear=True)
```
rjschwei, line 49: The testing should also be setup to trigger an actual output of the `tolower()`, i.e. have the config with uppercase and then validate that lower case pops out at the end so we get the whole chain of the implementation to do something.

rjschwei (changes requested):

PR #231, opened 2021-11-29
Show config file content in the migration log

### LINKED ISSUE #228: Include the config file in the migration log
To help with debugging it is helpful to understand the customizations set in the config file. It would be nice if we could echo the config file content at the top of the log file.

jesusbv:
This Fixes #228


schaefi (approved):
looks good :+1: 

feel free to merge when you see fit. Thanks

rjschwei (approved):

PR #232, opened 2021-11-30
Add migration_product to documentation

### LINKED ISSUE #229: migration_product parameter not included in DMS doc
The migration_product parameter for sle-migration-service.yml is not included in the official Distribution Migration System documentation.  This can cause migration resolution issues that are difficult to find if an operator uses an incorrect product unless those involved with supporting the upgrade process are already aware of this feature.  This parameter should be added to the DMS doc "Optional Customization of the Upgrade Process" so its usefulness and availability can be publicly acknowledged.

jesusbv:
This Fixes #229


rjschwei (commented):
LGTM

schaefi (approved):
That makes it crystal clear imho :+1:

PR #236, opened 2022-02-09
Fix pytest 7 compatibility

KeithMnemonic:
This PR switches the pytest setup() to setup_method() to allow suse-migration-services to work with pytest 7.0. This is similar to the change made to mash that also provides pytest 7.0 compatability


KeithMnemonic:
@rjschwei @jesusbv @smarlowucf  Could you please review? I included Sean as the fix I did here is similar to the fix he did for mash to resolve the same issue.

smarlowucf (approved):

PR #237, opened 2022-02-28
Add more details around migration_proudct

KeithMnemonic:
Currently, the information on using migration_product is a bit ambiguous. This PR add some specific examples. Also if migration_product is fully supported, the last section of the doc needs to reflect migrations other than to 15-SP1 are also supported.

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -157,11 +157,22 @@ manually. Once done, the upgrade process uses `zypper dup` and expects
 all required repositories to be setup correctly.
 
 Migration product target::
-If set, the migration will use that product as a target.
-The product must be specified with the triplet PROD_NAME/VERSION/ARCH, for example:
+The default product target will be the 15-SP1 release corresponding to the
+current product. The migration_product setting can be used to change the
+product target. 
+
+[NOTE]
+Changing the product target from the 15-SP1 release is not officially
+supported
+
+If migration_product is set, the migration will use that product as a target.
+The product must be specified with the triplet name/version/arch found
+in '/etc/products.d/baseproduct' of the target product, for example:
 
 [listing]
-migration_product: 'SLES/15/x86_64'
+migration_product: 'SLES/15.1/x86_64'
+
+would set the product target to SLES-15-SP1
 
 [NOTE]
 The default target, if `migration_product` is not set, is the version


doc/adoc/user_guide.adoc
```diff
@@ -273,7 +278,6 @@ a pointer to the respective log file.
   cloud framework is concerned, you will still be running an instance
   of the SLES version you started with. This cannot be changed.
 * The only supported migration path in the Public Cloud is from the
-  final 2 service packs of a distribution to the first service pack of
-  the next distribution. For example from SLES 12 SP4 or SLES 12 SP5 to
-  SLES 15 SP1. The packages delivered by SUSE in the Public Cloud Module
+  final 2 service packs of a distribution. For example from SLES 12 SP4 or SLES 12 SP5 to
+  SLES 15 SPX or . The packages delivered by SUSE in the Public Cloud Module
```
jesusbv, line 282: `SLES 15 SPX or . ` or ?

doc/adoc/user_guide.adoc
```diff
@@ -273,7 +278,6 @@ a pointer to the respective log file.
   cloud framework is concerned, you will still be running an instance
   of the SLES version you started with. This cannot be changed.
 * The only supported migration path in the Public Cloud is from the
-  final 2 service packs of a distribution to the first service pack of
-  the next distribution. For example from SLES 12 SP4 or SLES 12 SP5 to
-  SLES 15 SP1. The packages delivered by SUSE in the Public Cloud Module
+  final 2 service packs of a distribution. For example from SLES 12 SP4 or SLES 12 SP5 to
```
jesusbv, line 281: For example,

doc/adoc/user_guide.adoc
jesusbv, line 282: This PR is an attempt to avoid ambiguity, IMHO it should state that the migration to 15 SP2 or 15 SP3 (or non default paths) happens when setting `migration_product` and not leave that to interpretation (reader could understand by default or sometimes, not being clear when)

my 2 cents

doc/adoc/user_guide.adoc
```diff
@@ -163,6 +163,11 @@ The product must be specified with the triplet PROD_NAME/VERSION/ARCH, for examp
 [listing]
 migration_product: 'SLES/15/x86_64'
 
+or 
+
+migration_product: 'SLES/15.2/x86_64' (Migrate to SLES-15-SP2)
+migration_product: 'SLES/15.3/x86_64' (Migrate to SLES-15-SP3)
```
rjschwei, line 169: And now you'd have to also show how this is done for SLES For SAL and then for SLE HPC and then for potentially other products.

I think we should give 1 example and then explain to the user how to compose the triplet. Yes, explaining how to compose the triplet is a little more difficult as the system itself does not have any of this information but we should be able to explain this in a way that makes sens without having to list every product and every possible version. Further listing a specific target in the doc may give the impression that the listed target is supported.

doc/adoc/user_guide.adoc
rjschwei: No SPX cannot be supported, well ultimately that is no longer the Public Cloud Teams decision. But ultimately if you want to support SPX, meaning any SP of the next distribution you have to build a migration iso based on every SPX due to the potential changes in zypper and the update stack. There are cases where zypper on 15 SP1 may not be able to migrated a target system to 15 SP3. The consistency of the update stack of the migration image and the migration target is important.

KeithMnemonic: I can call out one example in more detail based on Robert's feedback. But if none of these are "officially" supported we need to highlight that.  Something like

 "The default migration target  is SLES 15 SP1.  Using the migration_product the default path can be changed.  For example a migration_product set to ''SLES/15.3/x86_64' will change the target to SLES 15 SP3"

How does this sound?

KeithMnemonic: If we are providing migration_target, it would be good to list any known supported migrations. If none other than SLES-15-SP1 exist, we need to clearly state these others are unsupported

KeithMnemonic:
@rjschwei @jesusbv please re-review. I tried to just address the ambiguity around he syntax and not infer any other option is supported

jeremy-moffitt (approved):
looks simple enough and clearly clarifies the issue

doc/adoc/user_guide.adoc
```diff
@@ -157,11 +157,19 @@ manually. Once done, the upgrade process uses `zypper dup` and expects
 all required repositories to be setup correctly.
 
 Migration product target::
-If set, the migration will use that product as a target.
+The default product target is SLES-15-SP1. The migration_product setting can be used
```
rjschwei, line 160: This is not correct DMS determines the product

https://github.com/SUSE/suse-migration-services/blob/master/suse_migration_services/migration_config.py#L87

We can say that the default SP target is `SP1` but we cannot state that the default target is `SLES`

doc/adoc/user_guide.adoc
```diff
@@ -157,11 +157,19 @@ manually. Once done, the upgrade process uses `zypper dup` and expects
...
+to change the product target. 
+
+[NOTE]
+Changing the product target is not officially supported
```
rjschwei, line 164: This is somewhat misleading. While a user doesn't have to set it anymore we did have a bug a while back where the product was not detected and users had to use `migration_product`. Also if the users uses

`migration_product: 'SLE_SAP/15.1/x86_64'`

that is certainly supported. Of course there is not really a need for a user to do this anymore. However this in conjunction with line 160 gives the impression that `SLE_SAP` and `SLE_HPC` are not supported as products for migration. That is incorrect.

doc/adoc/user_guide.adoc
```diff
@@ -157,11 +157,19 @@ manually. Once done, the upgrade process uses `zypper dup` and expects
...
+
+If migration_product is set, the migration will use that product as a target.
 The product must be specified with the triplet PROD_NAME/VERSION/ARCH, for example:
 
 [listing]
-migration_product: 'SLES/15/x86_64'
+migration_product: 'SLES/15.1/x86_64'
+
+would set the product target to SLES-15-SP1
```
rjschwei, line 172: We can state the the product tripplet is composed of the information from `/etc/products.d/baseproduct` by combining the `name`, `version`, and `arch` values from the XML document.

rjschwei (changes requested):

doc/adoc/user_guide.adoc
KeithMnemonic: Thanks I will reword that as

"The default product target will be the 15-SP1 release corresponding to the current product."

Does the read better?

doc/adoc/user_guide.adoc
KeithMnemonic: Should we give the three examples, SLES,SLE_SAP and SLE_HPC and then point to  `/etc/products.d/baseproduct` ?  Or just point to  `/etc/products.d/baseproduct`  and make the example

migration_product: name/version/arch

Currently it has 

migration_product:  'SLES/15/x86_64'

which does not seem useful

KeithMnemonic:
@jesusbv  @rjschwei  Can you please re-review and see if this is good now?

rjschwei (approved):

PR #239, opened 2022-04-01
Add an option to create a 'fat' initrd

KeithMnemonic:
This change adds a new systemd service, regenerate-initrd that is
responsible for running the dracut command to build a "fat" initrd
when the config setting 'build_host_independent_initrd' is enabled. 

This service is configured to run before the kernel-load service
and allows a new initrd to be created after the migration is complete
and before the final system reboot is performed.

The dracut command that is executed is:

```
chroot  root_path dracut --no-kernel --no-host-only --no-hostonly-cmdline --regenerate-all --logfile  /var/log/YaST2/mkinitrd.log -f
```

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -227,6 +227,17 @@ If enabled, it will run the zypper migration plugin with increased verbosity.
 [listing]
 verbose_migration: true|false
 
+Configure Make initrd Method::
+The live system may not contain all necessary tools to create an initrd that
+meets the need of the system being upgraded. Building a host independent
+initrd will create an initrd in a way that contains the tools and
+modules available on the system being upgraded. If this is needed, a host
+independent initrd can be created by setting
+`build_host_independent_initrd: True`.
+
+[listing]
+build_host_independent_initrd: true|false
+
 == Run the Migration
 After the install of the `SLES15-Migration` package, start the migration
 by calling the following command:
--- a/suse_migration_services/units/regenerate_initrd.py
+++ b/suse_migration_services/units/regenerate_initrd.py
@@ -1,0 +1,129 @@
+# Copyright (c) 2022 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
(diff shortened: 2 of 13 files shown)


suse_migration_services/units/regenerate_initrd.py
```diff
@@ -0,0 +1,111 @@
...
+
+def run_dracut(root_path):
+    """Function run dracut"""
+
+    log = logging.getLogger(Defaults.get_migration_log_name())
+
+    try:
+        log.info(
+            'Running chroot {0} dracut --no-kernel --no-host-only --no-hostonly-cmdline'
+            '--regenerate-all --logfile /var/log/YaST2/mkinitrd.log -f'.format(root_path)
```
rjschwei, line 87: This should log to the migration log, but dracut probbaly creates a logfile from scratch which would destroy the other log information. as such log to `tmp/$SOMETHING` then concatenate the log to the migration log and remove `/tmp/$SOMETHING`

If we do not add the information to the migration log we'll be in an additional round trip to ask for more files when something goes wrong.

rjschwei (changes requested):

suse_migration_services/units/regenerate_initrd.py
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) 2018 SUSE Linux LLC.  All rights reserved.
```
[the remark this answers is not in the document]
KeithMnemonic: thanks for catching my attemp at a time machine

doc/adoc/user_guide.adoc
```diff
@@ -227,6 +227,14 @@ If enabled, it will run the zypper migration plugin with increased verbosity.
 [listing]
 verbose_migration: true|false
 
+Configure Make initrd Method::
+By default, the migration system does not include multipath when creating
```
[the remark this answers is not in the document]
KeithMnemonic: thanks for the wording.

suse_migration_services/units/regenerate_initrd.py
KeithMnemonic: got it, let me code that up and push up an update thanks

KeithMnemonic: please review this change and let me know if there is anything else

rjschwei (approved):

KeithMnemonic:
@jesusbv Please review when you get a chance.

suse_migration_services/units/regenerate_initrd.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if MigrationConfig().is_host_independent_initd_requested():
+        log.info('Creating a new host independent initrd')
+        root_path = Defaults.get_system_root_path()
+
+        do_bind_mounts(root_path)
+        run_dracut(root_path)
+
+
+def do_bind_mounts(root_path,):
```
jesusbv, line 48: bit nickpicking,
maybe rename to `dracut_bind_mounts` or something more specific for this case ?
as there are other bind mounts in the codebase

Another option would be refactor into having a common bind mounting method but not the purpose of this PR
and probably no much benefit

KeithMnemonic: sure I will change this

KeithMnemonic: fixed, @jesusbv please re-review

jesusbv (approved):
LGTM

PR #240, opened 2022-04-04
Draft: Split out pre-checks from suse-migration-sle15-activation

### LINKED ISSUE #209: Support pre-checks also for the run_migration method
## Problem Description

If customer decides to go with the real reboot mode, the package suse-migration-sle15-activation gets installed and the pre-checks runs as part of %post. So all good in this case.

If the customer decides to go with run_migration, no activation package will be installed because run-migration is provided with the package that contains the migration live image (SLE15-Migration). In this case no pre-checks are available

If the customer installs suse-migration-sle15-activation for the reason to get the pre-checks, there will also be the grub menu entry which is unwanted but does not hurt.

For the moment we could add some information in the documentation that says to run the pre-checks, install the activation package. No matter how the migration is started, the activation and also migration packages will be uninstalled by the migration code itself (except this stage is not reached by the DMS). However, this feels not natural to install an activation package for running pre-checks, so feels more like a workaround

Another option would be to split the suse-migration-sle15-activation package into two packages:

* suse-migration-sle15-activation
* suse-migration-sle15-pre-checks

The activation package would require the pre-checks, such that %post works. Customers using run_migration can install only the pre-checks package and manually call them. This needs to be documented. And a new pre-checks package must be published through the SUSE process

KeithMnemonic:
This PR introduces a new package , suse-migration-pre-checks which contains the pre-checks that were previously only included in suse-migration-sle15-activation. This allows users that are only using SLE15-Migration ( kexec) to also be able to use the pre-checks. Prior to this change, users had to install the suse-migration-sle15-activation (reboot) in order to have the pre-checks. 

Closes #209


doc/adoc/user_guide.adoc
```diff
@@ -90,6 +90,19 @@ If your system is not configured for it, it is recommended that at least
...
+== Upgrade Pre-Checks
+The suse-migration-pre-checks package contains the `suse-migration-pre-checks`
+script that checks for possible incompatibilities when doing a migration.
+
+These incompatibilities include:
+
+ Encrypted file systems
+ Invalid repository types
+
+This script is run during the install of suse-migration-pre-checks. It can
```
rjschwei, line 102: Per discussion we will not run the prechecks here, but from the package that contains the ISO.

doc/adoc/user_guide.adoc
```diff
@@ -103,6 +116,7 @@ version installation media.
 To install the distribution migration system run:
 
 [listing]
+tux > sudo zypper in suse-migration-pre-checks
```
rjschwei, line 119: This is not necessary since the ISO will depend on "pre-checks"

package/suse-migration-pre-checks-spec-template
```diff
@@ -0,0 +1,62 @@
+#
+# spec file for package suse-migration-pre-checks
+#
+# Copyright (c) 20022 SUSE LINUX GmbH, Nuernberg, Germany.
```
rjschwei, line 4: I had no idea I lived that long already ;) (extra 0 in the year)

package/suse-migration-pre-checks-spec-template
```diff
@@ -0,0 +1,62 @@
...
+
+# Please submit bugfixes or comments via http://bugs.opensuse.org/
+#
+%global           MinSLEVersion 12.3
+
+Name:             suse-migration-pre-checks
+Version:          %%VERSION
+Release:          0
+Url:              https://github.com/SUSE/suse-migration-services
+Summary:          SUSE Distribution SLE15 Migration Activation
```
rjschwei, line 23: pre-checks, not "Activation"

package/suse-migration-pre-checks-spec-template
```diff
@@ -0,0 +1,62 @@
...
+
+%install
+python3 setup.py install --prefix=%{_prefix} --root=%{buildroot}
+
+mv %{buildroot}/%{_bindir}/suse-migration-pre-checks /tmp/
+rm %{buildroot}/%{_bindir}/*
+mv /tmp/suse-migration-pre-checks %{buildroot}/%{_bindir}/
+
+%post
+suse-migration-pre-checks
```
rjschwei, line 55: No `%post` here

package/suse-migration-sle15-activation-spec-template
```diff
@@ -50,17 +45,12 @@ next reboot of the machine.
...
 install -D -m 755 grub.d/99_migration \
     %{buildroot}/etc/grub.d/99_migration
 
-mv %{buildroot}/%{_bindir}/suse-migration-pre-checks /tmp/
-rm %{buildroot}/%{_bindir}/*
-mv /tmp/suse-migration-pre-checks %{buildroot}/%{_bindir}/
-
 %post
-suse-migration-pre-checks
+#suse-migration-pre-checks
```
rjschwei, line 53: Remove, do not comment out

rjschwei (changes requested):

doc/adoc/user_guide.adoc
KeithMnemonic: Thanks I updated it to SLE15-Migration

package/suse-migration-pre-checks-spec-template
KeithMnemonic: Thanks, fixed

package/suse-migration-pre-checks-spec-template
KeithMnemonic: changed to "Pre-Checks"

package/suse-migration-sle15-activation-spec-template
KeithMnemonic: removed

package/suse-migration-pre-checks-spec-template
KeithMnemonic: removed

doc/adoc/user_guide.adoc
```diff
@@ -90,6 +90,19 @@ If your system is not configured for it, it is recommended that at least
...
+
+These incompatibilities include:
+
+ Encrypted file systems
+ Invalid repository types
+
+This script is run during the install of SLES15-Migration. It can
+also be run anytime using `/usr/bin/suse-migration-pre-checks`
+
+This script
```
jesusbv, line 105: maybe a blank line under this
as it is now, it produces
image

KeithMnemonic: thanks I will fix and render the doc to make usre it looks ok.

KeithMnemonic: How does this look @jesusbv ?

image

jesusbv (approved):

package/suse-migration-sle15-activation-spec-template
```diff
@@ -29,14 +29,9 @@ BuildArch:        noarch
 Conflicts:        suse-migration-services
 BuildRequires:    grub2
 BuildRequires:    python3-devel
-BuildRequires:    python3-setuptools
-BuildRequires:    python3-Cerberus
-BuildRequires:    python3-PyYAML
 Requires:         SLES15-Migration
+Requires:         suse-migration-pre-checks
```
rjschwei, line 33: Requires(post)

package/suse-migration-sle15-activation-spec-template
```diff
@@ -50,17 +45,11 @@ next reboot of the machine.
...
 
 install -D -m 755 grub.d/99_migration \
     %{buildroot}/etc/grub.d/99_migration
 
-mv %{buildroot}/%{_bindir}/suse-migration-pre-checks /tmp/
-rm %{buildroot}/%{_bindir}/*
-mv /tmp/suse-migration-pre-checks %{buildroot}/%{_bindir}/
-
 %post
-suse-migration-pre-checks
```
rjschwei, line 63: WE still want to run the script, but should have the fully qualified path

rjschwei (changes requested):

package/suse-migration-sle15-activation-spec-template
KeithMnemonic: @rjschwei currently I only had modified the image.spec.in  in the IBS package, suse-migration-rpm to run this as a %post. I can add it here. as well.  The only downside is if a customer just  runs `zypper in suse-migration-sle15-activation (allowing zypper to pull in suse-migration-pre-checks and SLE15-Migration) , it will get run twice, once when SLES15-Migration is installed and once when suse-migration-sle15-activation in installed. Fine by me but wanted to make sure it as noted.  Let me know if this changes your mind at all.

rjschwei: This is where I got confused, sorry. `-activation` shouldn't have a requires on `pre-checks` as you state `pre-checks` are run when `SLES15-Migration` is installed and one cannot install `-activation` without pulling `SLES15-Migration`. Wel, one can deliberately break things but that is a different conversation.

package/suse-migration-sle15-activation-spec-template
KeithMnemonic: the requires of suse-migration-pre-checks can be removed here  since it is inherited from SLES15-Migration

package/suse-migration-sle15-activation-spec-template
KeithMnemonic: ok So I just removed the requires. Please re-review

rjschwei (approved):

PR #241, opened 2022-04-13
Add multiversion kernel checks to pre-checks

KeithMnemonic:
This PR adds an additional check to suse-migration-pre-checks. This check is used to provide warnings when the following occur

1. multiversion.kernels is not set to latest,running
2. More than one kernel is currently installed on the system.

The next step is to add a "fix" option to resolve either of the two warnings above.  This will be done in a follow up PR as there will be some rework to pre_checks.py to handle command line arguments. Currently pre-checks does not support any additional arguments.

### DIFF
--- a/suse_migration_services/prechecks/kernels.py
+++ b/suse_migration_services/prechecks/kernels.py
@@ -1,0 +1,77 @@
+# Copyright (c) 2022
+#
+# SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+"""Checks the multiversion.kernels setting and the presence of multiple kernels
+
+Check to see if multiversion.kernels are enabled in /etc/zypp/zypp.conf and
+if there is more than one kernel currently installed. If detected, a warning is reported
+as multiversion.kernels needs to be disabled and old kernels cleaned up before
+a migration can be started."""
+
+import configparser
+import logging
+
+from suse_migration_services.defaults import Defaults
+from suse_migration_services.command import Command
+
+
+def multiversion_and_multiple_kernels():
+    """Check for the number of installed kernels"""
+    log = logging.getLogger(Defaults.get_migration_log_name())
+
+    config = configparser.ConfigParser()
+    config.read(Defaults.get_zypp_config_path())
+
+    kernel_multi_version_enabled = config.get('main', 'multiversion', fallback=None)
+
+    if 'kernel' in kernel_multi_version_enabled:
+        kernel_multi_version_settings = config.get('main', 'multiversion.kernels', fallback=None)
+
+        log.info("The config option 'multiversion' in /etc/zypp/zypp.conf "
+                 "includes the keyword 'kernel.' The current value is set "
+                 "as \n'multiversion = %s'. \nChecking the config option "
+                 "'multiversion.kernels' to see if multiple kernels are "
+                 "also enabled", kernel_multi_version_enabled)
+
+        if not kernel_multi_version_settings:
+            log.warning('Missing multiversion.kernels setting in zypp.conf. '
+                        'Please ensure it is set as:\n'
+                        "'multiversion.kernels = latest,running'")
+        elif set(kernel_multi_version_settings.split(',')) != {'running', 'latest'}:
+            log.warning('The config option multiversion.kernels is not set '
+                        'correctly in /etc/zypp/zypp.conf. It is currrently '
+                        'set as \n%s.\nPlease ensure it is set as:\n'
+                        "'multiversion.kernels = latest,running'",
+                        kernel_multi_version_settings)
+
+    kernel_type = 'kernel-default'
+    running_kernel = Command.run(["uname", "-r"])
+
+    if 'azure' in running_kernel.output:
+        kernel_type = 'kernel-azure'
+
+    installed_kernels = Command.run(["rpm", "-qa", kernel_type])
+    multiple_kernels = len(installed_kernels.output.splitlines()) > 1
+
+    if multiple_kernels:
+        log.warning(
+            'Multiple kernels have been detected on the system:\n'
+            '%s\n'
+            'Please remove all kernels other than the currrent running kernel, '
+            '%s\n', installed_kernels.output, running_kernel.output)
--- a/test/unit/pre_checks_test.py
+++ b/test/unit/pre_checks_test.py
@@ -4,19 +4,20 @@ from unittest.mock import (
     patch, Mock
 )
 
-from suse_migration_services.prechecks.pre_checks import main
 from suse_migration_services.fstab import Fstab
 import suse_migration_services.prechecks.repos as check_repos
+import suse_migration_services.prechecks.fs as check_fs
+import suse_migration_services.prechecks.kernels as check_kernels
 
 
 @patch('suse_migration_services.logger.Logger.setup')
-@patch('os.listdir')
-@patch('os.path.exists')
 class TestPreChecks():
     @fixture(autouse=True)
     def inject_fixtures(self, caplog):
         self._caplog = caplog
 
+    @patch('os.listdir')
+    @patch('os.path.exists')
     @patch('configparser.RawConfigParser.items')
     @patch('suse_migration_services.command.Command.run')
     @patch('suse_migration_services.prechecks.fs.Fstab')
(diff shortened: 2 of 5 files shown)


test/unit/pre_checks_test.py
```diff
@@ -4,19 +4,23 @@
     patch, Mock
 )
 
-from suse_migration_services.prechecks.pre_checks import main
+# from suse_migration_services.prechecks.pre_checks import main
```
rtamalin, line 7: Guessing this was commented out for debug purposes?

test/unit/pre_checks_test.py
```diff
@@ -4,19 +4,23 @@
...
 from suse_migration_services.fstab import Fstab
 import suse_migration_services.prechecks.repos as check_repos
+import suse_migration_services.prechecks.fs as check_fs
+import suse_migration_services.prechecks.kernels as check_kernels
 
 
 @patch('suse_migration_services.logger.Logger.setup')
-@patch('os.listdir')
-@patch('os.path.exists')
+# @patch('os.listdir')
```
rtamalin, line 15: Needed any more?

test/unit/pre_checks_test.py
```diff
@@ -4,19 +4,23 @@
...
+# @patch('os.path.exists')
```
rtamalin, line 16: Ditto

test/unit/pre_checks_test.py
KeithMnemonic: thanks, will remove

test/unit/pre_checks_test.py
KeithMnemonic: thanks, will remove

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,69 @@
...
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+"""Pre-Checks multiversion kernels
+
+Check to see if multiversion.kernels are enabled in /etc/zypp/zypp.conf and
+if there is more than one kernel currently installed. If detected a warning is reported
+as multiversion.kernels needs to be disabled and old kernels cleaned up before
+migrationg."""
+
+# import os
```
guangyee, line 27: may want to remove this commented out line

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,69 @@
...
+    """Check for the number of installed kernels"""
+    log = logging.getLogger(Defaults.get_migration_log_name())
+
+    config = configparser.ConfigParser()
+    config.read(Defaults.get_zypp_config_path())
+
+    current_settings = config.get('main', 'multiversion.kernels', fallback=None)
+
+    if not current_settings:
+        log.warning('Missing multiversion.kernels setting in zypp.conf. '
```
guangyee, line 45: what's the default behavior if the option is missing?

suse_migration_services/prechecks/kernels.py
KeithMnemonic: thanks will remove

suse_migration_services/prechecks/kernels.py
KeithMnemonic: The default in the cloud images is 

`multiversion.kernels = latest,latest-1,running`

the images are built this way so it would only be missing (removed or commented out)  if a customer did this change after the image was booted.

guangyee: got it. Thanks.

guangyee (approved):
Looks good!

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,68 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+"""Pre-Checks multiversion kernels
+
+Check to see if multiversion.kernels are enabled in /etc/zypp/zypp.conf and
+if there is more than one kernel currently installed. If detected a warning is reported
```
jesusbv, line 23: s/If detected/If detected,

or `if those conditions are detected`

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,68 @@
...
+as multiversion.kernels needs to be disabled and old kernels cleaned up before
+migrationg."""
```
jesusbv, line 25: s/migrationg/migration

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,68 @@
...
+                    current_settings)
+
+    kernel_type = 'kernel-default'
+    running_kernel = Command.run(["uname", "-r"])
+
+    if 'azure' in running_kernel.output:
+        kernel_type = 'kernel-azure'
+
+    installed_kernels = Command.run(["rpm", "-qa", kernel_type])
+    count = len(installed_kernels.output.splitlines())
```
jesusbv, line 61: maybe 

```python
multiple_kernels = len(...) > 1

if multiple_kernels:
...
```

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,68 @@
...
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+"""Pre-Checks multiversion kernels
```
jesusbv, line 20: and multiple kernels

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,68 @@
...
+migrationg."""
+
+import configparser
+import logging
+
+from suse_migration_services.defaults import Defaults
+from suse_migration_services.command import Command
+
+
+def multiversion_kernels():
```
jesusbv, line 34: should the name be different ?
method does check for multiversion kernels _and_ multiple kernels as well

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def multiversion_and_multiple_kernels():
+    """Check for the number of installed kernels"""
+    log = logging.getLogger(Defaults.get_migration_log_name())
+
+    config = configparser.ConfigParser()
+    config.read(Defaults.get_zypp_config_path())
+
+    current_settings = config.get('main', 'multiversion.kernels', fallback=None)
```
rjschwei, line 41: This check is not sufficient. From zypp.conf

"""
Comma separated list of kernel packages to keep installed in parallel, if the
above multiversion variable is set.
"""

Meaning the `multiversion.kernels` setting only takes effect if `multiversion = provides:multiversion(kernel)` is also set. The `multiversion` variable could be set as shown (default), could be set to `multiversion=kernel`, could be set to something other than kernel or could not be set. We have to check the setting of `multiversion` first to avoid fals positive warnings.

suse_migration_services/prechecks/kernels.py
rjschwei: Which means by default we ship a configuration that enables multi kernels, hmm. This also means we are giving users a configuration that is explicitly not supported for migration.

Last but not lease the images we deliver are not ready for DMS. We should change our images to address this issue. Please create a bug for images starting with SLES 12 SP4.

rjschwei (changes requested):

suse_migration_services/prechecks/kernels.py
KeithMnemonic: fixed ,please re-review

suse_migration_services/prechecks/kernels.py
```diff
@@ -0,0 +1,77 @@
...
+
+    config = configparser.ConfigParser()
+    config.read(Defaults.get_zypp_config_path())
+
+    kernel_multi_version_enabled = config.get('main', 'multiversion', fallback=None)
+
+    if 'kernel' in kernel_multi_version_enabled:
+        kernel_multi_version_settings = config.get('main', 'multiversion.kernels', fallback=None)
+
+        log.warning("The config option 'multiversion' in /etc/zypp/zypp.conf "
```
jesusbv, line 46: is there an scenario where the `multiversion` and `multiversion.kernels` are set in a way that it is OK ? (in the context of DMS)
If so, should this log be a warning or info ?

rjschwei: At this point we don not really know yet if this is a potential problem of not, as such this should be an "info" message, IMHO.

rjschwei (approved):

suse_migration_services/prechecks/kernels.py
jesusbv: In the other checks, IIRC the log messages are warnings because at that time it's known it _could_ be a problem, in this case
in this case, a check is going to be done, so inform the user about that check (instead of warning them) makes sense to me

KeithMnemonic: @jesusbv  I am not sure I am completely following. Are you agreeing with Robert's suggestion of making the first message "info"  (shown below)  and that is sufficient or are you suggesting we also log if the config setting `'multiversion.kernels` is correct?  

```
        log.info("The config option 'multiversion' in /etc/zypp/zypp.conf "
                    "includes the keyword 'kernel.' The current value is set "
                    "as \n'multiversion = %s'. \nChecking the config option "
                    "'multiversion.kernels' to see if multiple kernels are "
                    "also enabled", kernel_multi_version_enabled)
```

jesusbv: Sorry, I was replying to [this](https://github.com/SUSE/suse-migration-services/pull/241#discussion_r855173983) comment,
trying to explain why this _should_ be a `log.info` message and why the other checks messages were warnings

I [suggested](https://github.com/SUSE/suse-migration-services/pull/241#discussion_r854966586) it should be a log info 
or thought that was clear

So, yes that change would make more sense to me

jesusbv (approved):

PR #243, opened 2022-04-18
Draft: Remove --no-kernel from host independant initrd

KeithMnemonic:
This PR removes the `--no-kernel` option when creating the host independant initrd. In the most recent testing from the customer environment that is requesting this change, the server did not boot and complained about a missing dm_mod.  In doing further testing when removing the `--no-kernel` option, the kernel modules are also included in the initrd, as seen by lsinitrd.

Once we get confirmation this resolves the customer issue completely, I will remove the Draft state.


rjschwei (approved):
Not sure why this is marked as draft, thus not merging

KeithMnemonic:
> this

It was draft as we did not get confirmation from the customer that removing this resolved the issue. Feel free to remove draft if you believe this is the correct method.  The use of `--no-kernel` came from a comment in the Bugzilla.  https://bugzilla.suse.com/show_bug.cgi?id=1194320#c17 but it seems this was incorrect and the `--no-kernel` should not be used.

rjschwei:
> > this
> 
> It was draft as we did not get confirmation from the customer that removing this resolved the issue. Feel free to remove draft if you believe this is the correct method. The use of `--no-kernel` came from a comment in the Bugzilla. https://bugzilla.suse.com/show_bug.cgi?id=1194320#c17 but it seems this was incorrect and the `--no-kernel` should not be used.

We definitely need dracut to include the appropriate kernel modules. It is a bit weird that it skipped the required modules with `--no-kernel` but that appears to be consistent with the documented behavior.

PR #244, opened 2022-04-19
Enable cli options for pre checks

KeithMnemonic:
This PR adds CLI option parsing to suse-migration-pre-checks and also adds a new --fix/-f option to suse-migration-pre-checks. When this option is used, suse-migration-pre-checks will set

'multiversion.kernels = latest,running' 

in /etc/zypp/zypp.conf

and also remove all kernels, other than the current running kernel.

In addition,   A `-t/--target` option has been added and is used to run the pre_checks once the migration
has been started and the system is running from the migration iso image. With this option, the pre-checks are performed using the pre-migration '/system-root' file structure when determining the zypper configuration, installed kernels, and active kernel of the system that is being migrated.


suse_migration_services/prechecks/pre_checks.py
```diff
@@ -32,9 +35,26 @@ def main():
...
     log.info('Running pre migration checks')
+
+    cli_parser = argparse.ArgumentParser(
+        prog='suse-migration-pre-checks',
+        description='Check for existing conditions that will cause a migration to fail'
+    )
+
+    cli_parser.add_argument(
+        '-f',
+        '--fix',
```
jesusbv, line 52: can this be `--fix-kernel` or `--fix-multikernels` ?
Something more specific, I would understand `fix` would fix _all_ checks instead of only the kernel

rjschwei: We do not know what else will need to be fixed, if we make the option kernel specific now we will have an interface change later.

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +56,50 @@ def multiversion_and_multiple_kernels():
...
+                        "correctly in /etc/zypp/zypp.conf. It is currrently "
+                        "set as:\n'multiversion.kernels = %s'.", kernel_multi_version_settings)
+            if fix:
+                log.warning("The '--fix' option was provided, "
+                            "setting 'multiversion.kernels = latest,running' ")
+                existing = 'multiversion.kernels = ' + kernel_multi_version_settings
+                correction = 'multiversion.kernels = latest,running'
+                sed_arg = 's/' + existing + '/' + correction + '/'
+                try:
+                    Command.run(["sed", "-i", sed_arg, Defaults.get_zypp_config_path()])
```
rjschwei, line 69: On success we need to return from here.

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +56,50 @@ def multiversion_and_multiple_kernels():
...
 
     if multiple_kernels:
-        log.warning(
-            'Multiple kernels have been detected on the system:\n'
-            '%s\n'
-            'Please remove all kernels other than the currrent running kernel, '
-            '%s\n', installed_kernels.output, running_kernel.output)
+        log.warning('Multiple kernels have been detected on the system:\n'
+                    '%s', '\n'.join(installed_kernels))
+        if fix:
```
rjschwei, line 89: Ditto w.r.t. the order.

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +56,50 @@ def multiversion_and_multiple_kernels():
...
+            running_kernel_version = kernel_type + '-' + \
+                re.search(r'([0-9.-]*(?=[-][a-zA-Z]))', running_kernel).group()
+
+            log.warning("The '--fix' option was provided, "
```
rjschwei, line 93: This is `info` the user told us to do something and we are giving feedback

rjschwei (changes requested):

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +56,50 @@ def multiversion_and_multiple_kernels():
...
+                except DistMigrationCommandException:
+                    log.error("ERROR: Unable to update /etc/zypp/zypp.conf")
+
+            else:
+                log.warning("Please ensure it is set as:\n"
+                            "'multiversion.kernels = latest,running'")
 
     kernel_type = 'kernel-default'
-    running_kernel = Command.run(["uname", "-r"])
+    running_kernel = Command.run(["uname", "-r"]).output
```
rjschwei, line 78: Note in the context of the life system this will return the name of the kernel in the ISO image

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +56,50 @@ def multiversion_and_multiple_kernels():
...
 
-    if 'azure' in running_kernel.output:
+    if 'azure' in running_kernel:
         kernel_type = 'kernel-azure'
 
-    installed_kernels = Command.run(["rpm", "-qa", kernel_type])
-    multiple_kernels = len(installed_kernels.output.splitlines()) > 1
+    installed_kernels = Command.run(["rpm", "-qa", kernel_type]).output.splitlines()
+    multiple_kernels = len(installed_kernels) > 1
```
rjschwei, line 84: In the context of the live system this will always be 1 since when we build the ISO there will always be only 1 kernel. SO there is a mismatch w.r.t. finding out what we really want to know, which is whether or not the system that is being migrated has multiple kernel packages installed.

suse_migration_services/prechecks/pre_checks.py
KeithMnemonic: marking resolved per Robert's comments

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +56,50 @@ def multiversion_and_multiple_kernels():
...
+                log.warning("The '--fix' option was provided, "
+                            "setting 'multiversion.kernels = latest,running' ")
+                existing = 'multiversion.kernels = ' + kernel_multi_version_settings
+                correction = 'multiversion.kernels = latest,running'
+                sed_arg = 's/' + existing + '/' + correction + '/'
+                try:
+                    Command.run(["sed", "-i", sed_arg, Defaults.get_zypp_config_path()])
+                except DistMigrationCommandException:
+                    log.error("ERROR: Unable to update /etc/zypp/zypp.conf")
+
```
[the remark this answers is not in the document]
KeithMnemonic: I have updated this, please review

suse_migration_services/prechecks/kernels.py
KeithMnemonic: I have updated this, please review

suse_migration_services/prechecks/kernels.py
KeithMnemonic: I have updated this, please review

suse_migration_services/prechecks/kernels.py
KeithMnemonic: I have updated this, please review

suse_migration_services/prechecks/kernels.py
KeithMnemonic: I have updated this, please review

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +56,50 @@ def multiversion_and_multiple_kernels():
...
+            else:
+                log.warning("Please ensure it is set as:\n"
+                            "'multiversion.kernels = latest,running'")
```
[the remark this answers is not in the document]
KeithMnemonic: I have updated this, please review

suse_migration_services/prechecks/kernels.py
KeithMnemonic: Please see if this is better now.

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +69,63 @@ def multiversion_and_multiple_kernels():
...
+
+    running_kernel = os.readlink(running_kernel_path)
 
-    if 'azure' in running_kernel.output:
+    if 'azure' in running_kernel:
         kernel_type = 'kernel-azure'
 
-    installed_kernels = Command.run(["rpm", "-qa", kernel_type])
-    multiple_kernels = len(installed_kernels.output.splitlines()) > 1
+    installed_kernels = Command.run(rpm_command).output.splitlines()
```
jesusbv, line 113: why did we change the use of `multiple_kernels` to go back to len(....) > 1 in the if ? 
did we not change it before ?

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +69,63 @@ def multiversion_and_multiple_kernels():
...
+        rpm_command = \
+            [
+                "chroot",
+                Defaults.get_system_root_path(),
+                "rpm",
+                "-qa",
+                kernel_type
+            ]
+
+    running_kernel = os.readlink(running_kernel_path)
```
jesusbv, line 108: can we move this statement inside the `if multiple_kernels` ?

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +69,63 @@ def multiversion_and_multiple_kernels():
...
 
-    if 'azure' in running_kernel.output:
+    if 'azure' in running_kernel:
```
jesusbv, line 110: ditto, can we move this statement inside the `if fix` block ?

suse_migration_services/prechecks/kernels.py
KeithMnemonic: Jesus,  I need to be able to print the list lof installed_kernels for the user. if I do 

```
    installed_kernels = Command.run(rpm_command).output.splitlines()
    multiple_kernels = len(installed_kernels) > 1
```    
that seems like declaring an extra variable.  Perhaps I did not understand fully, if so can you please explain.

suse_migration_services/prechecks/kernels.py
```diff
@@ -54,24 +69,63 @@ def multiversion_and_multiple_kernels():
...
+                            "'multiversion.kernels = latest,running'",
+                            kernel_multi_version_settings)
+
+
+def multiple_kernels_installed(fix, target, log):
+    """Check for multiple kernels installed"""
 
     kernel_type = 'kernel-default'
-    running_kernel = Command.run(["uname", "-r"])
+    running_kernel_path = '/' + Defaults.get_target_kernel()
```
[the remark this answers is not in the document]
KeithMnemonic: Thanks for the tip, I will switch to this.

suse_migration_services/prechecks/kernels.py
KeithMnemonic: actually it is in the wrong place. but it should not be in the 'if fix' since kernel_type is used in the rpm_command variable near the top of the function. It should be moved up. Thanks for having me review this and catching the error. I need to see why my test did not catch this error.

suse_migration_services/prechecks/kernels.py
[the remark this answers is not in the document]
KeithMnemonic: Thanks for the tip, I will switch to this.

Makefile
```diff
@@ -2,6 +2,7 @@ version := $(shell python3 -c 'from suse_migration_services.version import __VER
 
 build: check test
 	rm -f dist/*
+        # copy in the correct MANIFEST.in
```
rtamalin, line 5: This line uses expanded tab, but the other lines use tab, so the indentation is not consistent

KeithMnemonic: Thanks, that was not supposed to be there, it was leftover from exploring different options to resolve the "conflicts issue'

rtamalin (approved):
LGTM

suse_migration_services/prechecks/kernels.py
KeithMnemonic: @rjschwei I think this might be the only thing remaining. Please re-review and merge if no other objections. Thank you

guangyee (approved):
lets do this

rjschwei (approved):

PR #245, opened 2022-04-26
DRAFT: Add pre-checks as a systemd process

KeithMnemonic:
This PR completes the updated implementation of suse-migration-pre-checks.  A new suse-migration-run-pre-checks-service has been added that will be started right after the instance is rebooted to the migration image and the filesystem is mounted at /system-root. 

One PR #244 is merged, this PR will be updated to include the `--fix` option as the default behavior when the suse-migration-run-pre-checks-service is started.  A new option config has been added to allow this to be turned off. In that mode, the pre-checks are still run but without the `--fix` fix option.


rtamalin:
Looks good.

package/suse-migration-services-spec-template
```diff
@@ -26,6 +26,7 @@ Group:            System/Management
 Source:           suse-migration-services.tar.gz
 BuildRoot:        %{_tmppath}/%{name}-%{version}-build
 Conflicts:        suse-migration-sle15-activation
+Conflicts:        suse-migration-pre-checks
```
rjschwei, line 29: The pre-checks and the migration services both need to be installed in the Live ISO system that does the migration, they cannot conflict.

rjschwei (changes requested):

KeithMnemonic:
closing this in favor of https://github.com/SUSE/suse-migration-services/pull/247  . There were some rebase issues so I applied the change to a fresh pull of master

PR #246, opened 2022-04-29
Add an EnvironmentFile to suse-migration-prepare.service bsc#1199028

KeithMnemonic:
This PR adds an EnvironmentFile entry to suse-migration-prepare.service that ensure running SUSEConnect as chroot still has access to /etc/sysconfig/proxy and be able to communicate via a proxy.

In addition, while testing this fix, an issue was found with how the suse-migration-pre-checks was being created. This lead to a conflict as both suse-migration-pre-checks and suse-migration-services contained python modules. When suse-migraton-services was installed, it complained about some python files already being installed. The best solution without creating new packages (suse-migration-services-common, for example ) was to leverage the existing spec file and have it create a "sub package" of suse-migration-pre-checks. This allows proper control of ensuring both packages are installing a unique set of python files. 

The version was updated as this is needed for the `Conflicts < 2.0.33` in case a customer installed the original 2.0.32 and then upgraded to 2.0.33 before doing the migration


rtamalin:
Looks good to me

package/suse-migration-services-spec-template
```diff
@@ -26,6 +26,7 @@ Group:            System/Management
 Source:           suse-migration-services.tar.gz
 BuildRoot:        %{_tmppath}/%{name}-%{version}-build
 Conflicts:        suse-migration-sle15-activation
+Conflicts:         suse-migration-pre-checks
```
rjschwei, line 29: This cannot conflict. We will need to build both the pre-checks and the migration services into the Live ISO that is used to drive the migration as the migration services will need to run the pre-checks in `fix` mode, in the appropriate context, i.e. in the context of the system that is being upgraded and that is currently not running since we are running from within the ISO image.

rjschwei (changes requested):

package/suse-migration-services-spec-template
KeithMnemonic: The "Conflicts" is not blocking the install of suse-migration-pre-check and suse-migration-services together on the iso as I have tested this and it works . Rather since suse-migration-pre-checks is installed in the same python lib dir as suse-migration-services, the build server is reporting a conflict and aborts. When I use the "Conflicts" it allows the SLE15-Migration rpm to build correctly.    I think Markus ran into this when the original pre-checks were added and that is why he had to do this do this PR:

https://github.com/SUSE/suse-migration-services/pull/210  

https://github.com/SUSE/suse-migration-services/pull/210/files#diff-0b55022eedc44d6b4db6f9cf8a542fbeb013f9572bbd0162fa4b03c222ba87f4R28


https://github.com/SUSE/suse-migration-services/pull/210/files#diff-b8f5e9f254ff34608785645332e830f97640e3c924f81f53ee8563fd31e3a660R29

guangyee: So the workaround proposed by Keith should work fine since the precedence has already been established.  Problem with all three templates sharing the same setup.py is that there will be conflicts amongst all of them. One alternative would be that isolated the shared code into a common package and have all three packages depended on it, as Keith proposed in the email thread. Otherwise, we can try to restructure setup.py and the directories to produce multiple packages.

KeithMnemonic: Based on conversations between @rtamalin and I , a new solution, leveraging one spec file with "sub pacakges" was implemented and tested.

guangyee (approved):
I like this approach.

package/suse-migration-services-spec-template
```diff
@@ -104,9 +110,20 @@ install -D -m 644 systemd/suse-migration-console-log.service \
 # packed into a readonly squashfs layer. Therefore there is no
 # need for use of the "standard" systemd service handling macros.
 
+ls %{python3_sitelib}
```
rtamalin, line 113: Do we still need this `ls` command? I think it was intended for debugging purposes?

rjschwei (changes requested):

package/suse-migration-sle15-activation-spec-template
```diff
@@ -26,9 +26,7 @@ Group:            System/Management
 Source:           suse-migration-sle15-activation.tar.gz
 BuildRoot:        %{_tmppath}/%{name}-%{version}-build
 BuildArch:        noarch
-Conflicts:        suse-migration-services
```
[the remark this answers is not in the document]
KeithMnemonic: I will put this back

package/suse-migration-services-spec-template
```diff
@@ -25,10 +25,10 @@ License:          GPL-3.0+
 Group:            System/Management
 Source:           suse-migration-services.tar.gz
 BuildRoot:        %{_tmppath}/%{name}-%{version}-build
-Conflicts:        suse-migration-sle15-activation
```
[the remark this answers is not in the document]
KeithMnemonic: I will put this back

package/suse-migration-services-spec-template
```diff
@@ -38,14 +38,23 @@ Requires:         ca-certificates
...
 
 %description
 Systemd services to prepare and run a distribution migration process.
 
+%package          -n suse-migration-pre-checks
+Summary:          The pre-checks code used with suse-migration-services
+Group:            System/Management
+BuildRoot:        %{_tmppath}/%{name}-%{version}-build
+Requires:         python3-Cerberus
+Conflicts:        suse-migration-sle15-activation < 2.0.33
```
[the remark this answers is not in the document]
KeithMnemonic: It does conflict with the old activation as suse-migration-sle15-activation-2.0.32 installs the python modules that are part of suse-migration-services.  It does this since the pre-checks were originally part suse-migration-sle15-activation. I should have removed the python modules from suse-migration-sle15-activation when the suse-migration-pre-checks was first created.  I originally only removed the suse-migration-pre-checks console script and not all of other python stuff pulled in from suse-migration-services.  The new suse-migration-sle15-activation-2.0.33 does not include this any more (Ideally suse-migration-sle15-activation could be done differently now as it only contains the grub stuff and using the sdist is not needed.) Without this "Conflicts: suse-migration-sle15-activation < 2.0.33" you get a zypper conflict message if the customer would somehow upgrade from 2.0.32 to 2.0.33 before doing he migration.

The error without "Conflicts: suse-migration-sle15-activation < 2.0.33"
````
File /usr/lib/python3.4/site-packages/suse_migration_services/prechecks/pre_checks.py
  from install of
     suse-migration-pre-checks-2.0.33-1.83.1.noarch (Plain RPM files cache)
  conflicts with file from package
     suse-migration-sle15-activation-2.0.32-6.27.1.noarch (@System)

File conflicts happen when two packages attempt to install files with the same name but different contents. If you continue, conflicting files will be replaced losing the previous content.
Continue? [yes/no] (no): no

```
The message with "Conflicts: suse-migration-sle15-activation < 2.0.33" added in  (you get the expected if they try to upgrade)

```
Problem: suse-migration-pre-checks-2.0.33-1.84.1.noarch conflicts with suse-migration-sle15-activation < 2.0.33 provided by suse-migration-sle15-activation-2.0.32-6.27.1.noarch
 Solution 1: deinstallation of suse-migration-sle15-activation-2.0.32-6.27.1.noarch
 Solution 2: do not install suse-migration-pre-checks-2.0.33-1.84.1.noarch
```
If you think this is a corner case we do not need to consider,I can remove it. I do not think there is any harm in leaving it in as it provides a safety net.

PR #247, opened 2022-05-25
Enable prechecks as a systemd process

KeithMnemonic:
This PR completes the updated implementation of suse-migration-pre-checks. A new suse-migration-run-pre-checks-service has been added that will be started right after the instance is rebooted to the migration image and the filesystem is mounted at /system-root.  This service will use the  `--fix` option as the default behavior when the suse-migration-run-pre-checks-service is started. A new option config has been added to allow this to be turned off. In that mode, the pre-checks are still run but without the --fix fix option.


package/suse-migration-services-spec-template
```diff
@@ -132,6 +135,8 @@ install -D -m 644 systemd/suse-migration-console-log.service \
 %{_bindir}/suse-migration-mount-system
 %{_unitdir}/suse-migration-mount-system.service
 
+%{_unitdir}/suse-migration-pre-checks.service
```
guangyee, line 138: suse-migration-run-pre-checks.service?

KeithMnemonic: Are you asking about the typo I had (thanks for finding),  I just fixed the typo or the service. .  If you are asking about the services this PR will add a "pre-checks"  systemd service that runs suse-migration-pre-checks when the system boots from the migration iso.  It checks if the original system (mounted as /system-root) passed the pre-checks. if it detects the multi kernel issue, it will remediate before starting the upgrade/

guangyee: I was asking about the typo. Looks good now.

doc/adoc/user_guide.adoc
```diff
@@ -254,6 +254,13 @@ If enabled, it will run the zypper migration plugin with increased verbosity.
 [listing]
 verbose_migration: true|false
 
+Enable the fix option for pre_checks::
+If enabled (default), the run_pre_checks systemd process will use the `--fix`
+option to automatically remediate applicable issues before the migration is started. 
+
+[listing]
+pre_checks_fix: true|false
```
guangyee, line 262: What happen if this is set to false? Will the service remove the -f option from the command line?

```ExecStart=/usr/bin/suse-migration-pre-checks -f -t```

Comment (KeithMnemonic) [reply doc/adoc/user_guide.adoc:262]: @guangyee oh yeah i need to fix that ;-) I originally had a second pre-checks script that was used by the service but it seems redundant to have two python files doing the same thing. When I removed the second script, I forgot to add the logic back in to handle this variable. Thanks for catching. I will fix this and request a re-review once done

**Comment by KeithMnemonic** [comment]
Tested iso and see the following results:

Message when installing SLES15-Migration:

```
Running suse-migration-pre-checks with options: fix: False and live: False
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/aeff9ddd-df2c-4fae-b717-b5fed9622aea']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/1aef0f3b-e5f7-407c-b48c-68971164614b']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/cloud/azure_resource-part1']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/0DFD-1C1E']
The config option 'multiversion' in /etc/zypp/zypp.conf includes the keyword 'kernel.' The current value is set as
'multiversion = provides:multiversion(kernel)'.
Checking the config option 'multiversion.kernels' to see if multiple kernels are also enabled
The config option multiversion.kernels is not set correctly in /etc/zypp/zypp.conf. It is currrently set as:
'multiversion.kernels = latest,latest-1,running'.
Please ensure it is set as:
'multiversion.kernels = latest,running'
Calling: ['rpm', '-qa', 'kernel-default']
Multiple kernels have been detected on the system:
kernel-default-4.12.14-122.121.2.x86_64
kernel-default-4.12.14-122.106.1.x86_64
Please remove all kernels other than the currrent running kernel: vmlinuz-4.12.14-122.121-default
```

Message during migration:

```
Running suse-migration-pre-checks with options: fix: True and live: True
Device path /dev/disk/cloud/azure_resource-part1 not found and skipped
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/aeff9ddd-df2c-4fae-b717-b5fed9622aea']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/1aef0f3b-e5f7-407c-b48c-68971164614b']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/0DFD-1C1E']
The config option 'multiversion' in /etc/zypp/zypp.conf includes the keyword 'kernel.' The current value is set as
'multiversion = provides:multiversion(kernel)'.
Checking the config option 'multiversion.kernels' to see if multiple kernels are also enabled
The '--fix' option was provided, setting 'multiversion.kernels = latest,running'
Calling: ['sed', '-i', 's/multiversion.kernels = latest,latest-1,running/multiversion.kernels = latest,running/', '/etc/zypp/zypp.conf']
Calling: ['chroot', '/system-root', 'rpm', '-qa', 'kernel-default']
Multiple kernels have been detected on the system:
kernel-default-4.12.14-122.121.2.x86_64
kernel-default-4.12.14-122.106.1.x86_64
The '--fix' option was provided, removing old kernels
Removing: kernel-default-4.12.14-122.106.1.x86_64

```

File: doc/adoc/user_guide.adoc [hunk 3]
```diff
@@ -111,11 +111,9 @@ A `-f/--fix` option that will set multiversion.kernels to the correct value
...
 suse-migration-pre-checks.
 
-A `-t/--target` option is used to run the pre_checks once the migration
-has been started and the system is running from the migration iso image.
-With this option, the pre-checks are performed using the pre-migration
-file system when determining the zypper configuration, installed kernels,
-and active kernel of the system that is being migrated.
+A `-l/--live` option is used when suse-migration-pre-checks are run from the
+live migration iso image. With this option, suse-migration-pre-checks are
+validated against the mounted pre-migration file system .
```
rjschwei, line 116: I am wondering if we should not make this harder to access. We have two use cases that do not really overlap. One use case is the user wants to check the system before the migration. In this case the `-l` switch is never needed. The other use case is from within the migration system itself. When running from the migration image we always want the prechecks to run, period. As such I think it would be nicer to implement this with an option that the user cannot trip over. Meaning with the current implementation the user could have a cut and paste error, or a fat finger event and somehow `-l` ends up on the command line. That would cause all kinds of weird problems. However, if we keep this out of the user interface and set a flag file in the migration image that the code checks then we have a better separation of the 2 use cases, IMHO. Thoughts?

KeithMnemonic: That seems very reasonable.  One option is to use an environment variable that is set in the units file and then check for that within the code.  If we pick an obscure setting, it is unlikely a customer will have this set in their env

For example:

```
[Unit]
Description=Run suse-migration-pre-checks before starting the migration
After=suse-migration-mount-system.service suse-migration-ssh-keys.service
Requires=suse-migration-mount-system.service suse-migration-ssh-keys.service

[Service]
Type=oneshot
Environment="suse-migration-pre-checks-mode=live-iso-image"
ExecStart=/usr/bin/suse-migration-pre-checks --fix --live

[Install]
WantedBy=multi-user.target
What type of flag would I use.?The presences of some file we bake into the image. or a hidden env setting that is set within the unit file before calling suse-migration-pre-checks.  

Let me know your thoughts and I can modify the code to use this if this is acceptable

Comment (rjschwei) [reply doc/adoc/user_guide.adoc:116]: Environment variable in the unit file works for me. This can be injected in the build of the migration image, or the migration image build can simply have a custom unit file in the overlay tree.

Comment (KeithMnemonic) [reply doc/adoc/user_guide.adoc:116]: Thanks adding it in now and will build a new image and test it out.

Comment (KeithMnemonic) [reply doc/adoc/user_guide.adoc]: @rjschwei  please re-review now. This seems to be working 

```
Running suse-migration-pre-checks with options: fix: True
Using live mode
Device path /dev/disk/cloud/azure_resource-part1 not found and skipped
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/aeff9ddd-df2c-4fae-b717-b5fed9622aea']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/1aef0f3b-e5f7-407c-b48c-68971164614b']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/0DFD-1C1E']
The config option 'multiversion' in /etc/zypp/zypp.conf includes the keyword 'kernel.' The current value is set as
'multiversion = provides:multiversion(kernel)'.
Checking the config option 'multiversion.kernels' to see if multiple kernels are also enabled

The '--fix' option was provided, setting 'multiversion.kernels = latest,running'
Calling: ['sed', '-i', 's/multiversion.kernels = latest,latest-1,running/multiversion.kernels = latest,running/', '/etc/zypp/zypp.conf']
Calling: ['chroot', '/system-root', 'rpm', '-qa', 'kernel-default']
Multiple kernels have been detected on the system:

kernel-default-4.12.14-122.121.2.x86_64
kernel-default-4.12.14-122.106.1.x86_64

The '--fix' option was provided, removing old kernels
Removing: kernel-default-4.12.14-122.106.1.x86_64
Calling: ['chroot', '/system-root', 'rpm', '-e', 'kernel-default-4.12.14-122.106.1.x86_64']
```

so the only option is --fix now, and the env variable 

SUSE_MIGRATION_PRE_CHECKS_MODE=live-iso-image

is what triggered this message:

"Using live mode"

A migration from suse:sles-12-sp5-byos:gen2:2022.01.27 was successfull

**Review by rjschwei — APPROVED** [review]
LGTM. I suggest `s/live/migration_system/ g` or something similarly more explicit.

**Comment by KeithMnemonic** [comment]
> 

for the message 

```
        log.info("Using live mode")`
```

> LGTM. I suggest `s/live/migration_system/ g` or something similarly more explicit.

for the message

```
        log.info("Using live mode")
```

or somewhere else?

rjschwei:
> > 
> 
> > LGTM. I suggest `s/live/migration_system/ g` or something similarly more explicit.
> 
> @rjschwei for the message
> 
> ```
>         log.info("Using live mode")
> ```
> 
> or somewhere else?

I would do the substitution everywhere. Meaning arguments that are currently named `live` would be `migration_system` (or similar descriptive name). The variable name `live` within the context of the code has very limited meaning and expects the code reader to know that this code is built into a "live-iso".

KeithMnemonic:
> > > 
> > 
> > 
> > > LGTM. I suggest `s/live/migration_system/ g` or something similarly more explicit.
> > 
> > 
> > @rjschwei for the message
> > ```
> >         log.info("Using live mode")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or somewhere else?
> 
> I would do the substitution everywhere. Meaning arguments that are currently named `live` would be `migration_system` (or similar descriptive name). The variable name `live` within the context of the code has very limited meaning and expects the code reader to know that this code is built into a "live-iso".

Done @rjschwei  please re-review

rjschwei:
LGTM

rtamalin (approved):
LGTM

PR #251, opened 2022-08-25
Add fix for cloudregister location

### LINKED ISSUE #250: accomodate cloud-regionsrv-client file location cahnges
With version 10.0.4 of cloud-regionsrv-client files have been relocated from `/var/lib/cloudregister/` to `/var/cache/cloudregister/` to support SLE-Micro read only setup. In DMS we bind mount the cache of the to make update server information available to the migration system. The code needs to accomodate the new location, i.e. check if either exists and has content and then mount the proper location into DMS.

KeithMnemonic:
This PR addresses an issue with the th cloud-regionsrv-client > 10.04. In that version the SMT cache files have been relocated from  /var/lib/cloudregister to /var/cache/cloudregister.  To accommodate this change, a new function has been added, get_regionsrv_client_file_location which will search both locations to find the cache files. If none are found, an exception is created as the migration would fail otherwise.

In addition, a missing dependency on the suse-migration-services-pre-checks was added in.

closes #250

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -66,6 +66,9 @@ def main():
         '/usr/share/pki/trust/anchors/',
         '/etc/pki/trust/anchors/'
     ]
+    cache_cloudregister_path = '/var/cache/cloudregister'
+    cloud_register_metadata = ""
+
     if os.path.exists(suse_connect_setup):
         shutil.copy(
             suse_connect_setup, '/etc/SUSEConnect'
@@ -92,7 +95,6 @@ def main():
                 Path.create(trust_anchor)
                 for cert in certificates:
                     cert_file = os.sep.join([root_trust_anchor, cert])
-                    print(cert_file)
                     if os.path.islink(cert_file):
                         cert_file = os.sep.join(
                             [root_path, os.readlink(cert_file)]
@@ -117,18 +119,19 @@ def main():
         [root_path, 'usr', 'lib', 'zypp', 'plugins', 'services']
     )
 
-    try:
-        cloud_register_metadata = \
-            get_regionsrv_client_file_location(root_path)
-    except DistMigrationZypperMetaDataException as issue:
-        log.error('Failed locating regionsrv-client cache files: {0}'.format(
-            issue)
-        )
-        raise
+    with open(hosts_setup, 'r', encoding="utf-8") as fp:
+        if 'susecloud.net' in fp.read():
+            try:
+                cloud_register_metadata = \
+                    get_regionsrv_client_file_location(root_path)
+            except DistMigrationZypperMetaDataException as issue:
+                log.error(
+                    'Failed locating regionsrv-client cache files: {0}'.format(
+                        issue
+                    )
+                )
+                raise
 
-    # cloud_register_metadata = os.sep.join(
-    #     [root_path, 'var', 'lib', 'cloudregister']
-    # )
     zypper_log_file = os.sep.join(
         [root_path, 'var', 'log', 'zypper.log']
     )
@@ -174,12 +177,16 @@ def main():
             zypp_plugins_services, '/usr/lib/zypp/plugins/services'
         )
         if os.path.exists(cloud_register_metadata):
-            log.info('Bind mounting /var/cache/cloudregister')
-            Path.create('/var/cache/cloudregister')
+            log.info(
+                'Bind mounting {0} from {1}'.format(
+                    cache_cloudregister_path, cloud_register_metadata
+                )
+            )
+            Path.create(cache_cloudregister_path)
             Command.run(
                 [
                     'mount', '--bind', cloud_register_metadata,
-                    '/var/cache/cloudregister'
+                    cache_cloudregister_path
                 ]
             )
             update_smt_cache = '/usr/sbin/updatesmtcache'
@@ -271,20 +278,14 @@ def get_regionsrv_client_file_location(root_path):
     (pre cloud-regionsrv-client-10.0.4) to /var/cache/cloudregister
     (post cloud-regionsrv-client-10.0.4)
     """
-    old_cloud_register_path = os.sep.join(
-        [root_path, 'var', 'lib', 'cloudregister']
-    )
-    new_cloud_register_path = os.sep.join(
-        [root_path, 'var', 'cache', 'cloudregister']
-    )
-    if os.path.isdir(new_cloud_register_path) and \
-            any('SMT' in x for x in os.listdir(new_cloud_register_path)):
-        return new_cloud_register_path
-    if os.path.isdir(old_cloud_register_path) and \
-            any('SMT' in x for x in os.listdir(old_cloud_register_path)):
-        return old_cloud_register_path
+    for cloud_register_path in [
+        os.sep.join([root_path, 'var', 'cache', 'cloudregister']),
+        os.sep.join([root_path, 'var', 'lib', 'cloudregister'])
+    ]:
+        if os.path.isdir(cloud_register_path) and \
+                any('SMT' in x for x in os.listdir(cloud_register_path)):
+            return cloud_register_path
     raise DistMigrationZypperMetaDataException(
-        'No cloud-regionsrv-client cache files found in {0} or {1}'.format(
-            new_cloud_register_path, old_cloud_register_path
-        )
+        'No cloud-regionsrv-client cache files found in '
+        '/var/cache/cloudregister or /var/lib/cloudregister'
     )
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -34,7 +34,7 @@ from suse_migration_services.units.setup_host_network import (
 
 from suse_migration_services.exceptions import (
     DistMigrationZypperMetaDataException,
-    DistMigrationSystemNotRegisteredException
+    DistMigrationSystemNotRegisteredException,
 )
 
 
@@ -92,6 +92,7 @@ def main():
                 Path.create(trust_anchor)
                 for cert in certificates:
                     cert_file = os.sep.join([root_trust_anchor, cert])
+                    print(cert_file)
                     if os.path.islink(cert_file):
                         cert_file = os.sep.join(
                             [root_path, os.readlink(cert_file)]
@@ -115,9 +116,19 @@ def main():
     zypp_plugins_services = os.sep.join(
         [root_path, 'usr', 'lib', 'zypp', 'plugins', 'services']
     )
-    cloud_register_metadata = os.sep.join(
-        [root_path, 'var', 'lib', 'cloudregister']
-    )
+
+    try:
+        cloud_register_metadata = \
+            get_regionsrv_client_file_location(root_path)
+    except DistMigrationZypperMetaDataException as issue:
+        log.error('Failed locating regionsrv-client cache files: {0}'.format(
+            issue)
+        )
+        raise
+
+    # cloud_register_metadata = os.sep.join(
+    #     [root_path, 'var', 'lib', 'cloudregister']
+    # )
     zypper_log_file = os.sep.join(
         [root_path, 'var', 'log', 'zypper.log']
     )
@@ -163,12 +174,12 @@ def main():
             zypp_plugins_services, '/usr/lib/zypp/plugins/services'
         )
         if os.path.exists(cloud_register_metadata):
-            log.info('Bind mounting /var/lib/cloudregister')
-            Path.create('/var/lib/cloudregister')
+            log.info('Bind mounting /var/cache/cloudregister')
+            Path.create('/var/cache/cloudregister')
             Command.run(
                 [
                     'mount', '--bind', cloud_register_metadata,
-                    '/var/lib/cloudregister'
+                    '/var/cache/cloudregister'
                 ]
             )
             update_smt_cache = '/usr/sbin/updatesmtcache'
@@ -251,3 +262,29 @@ def get_root_disk_device(root_path):
                 block_type = block_record[1]
                 if block_type in considered_block_types:
                     return block_record[0]
+
+
+def get_regionsrv_client_file_location(root_path):
+    """
+    Find the correct path for the cloud-regionsrv-client cache files
+    as the location moves from /var/lib/cloudregister
+    (pre cloud-regionsrv-client-10.0.4) to /var/cache/cloudregister
+    (post cloud-regionsrv-client-10.0.4)
+    """
+    old_cloud_register_path = os.sep.join(
+        [root_path, 'var', 'lib', 'cloudregister']
+    )
+    new_cloud_register_path = os.sep.join(
+        [root_path, 'var', 'cache', 'cloudregister']
+    )
+    if os.path.isdir(new_cloud_register_path) and \
+            any('SMT' in x for x in os.listdir(new_cloud_register_path)):
+        return new_cloud_register_path
+    if os.path.isdir(old_cloud_register_path) and \
+            any('SMT' in x for x in os.listdir(old_cloud_register_path)):
+        return old_cloud_register_path
+    raise DistMigrationZypperMetaDataException(
+        'No cloud-regionsrv-client cache files found in {0} or {1}'.format(
+            new_cloud_register_path, old_cloud_register_path
+        )
+    )
--- a/test/unit/units/prepare_test.py
+++ b/test/unit/units/prepare_test.py
@@ -2,7 +2,7 @@ import shutil
 from tempfile import NamedTemporaryFile
 from configparser import ConfigParser
 from unittest.mock import (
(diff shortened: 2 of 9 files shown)


suse_migration_services/units/prepare.py
```diff
@@ -251,3 +262,29 @@ def get_root_disk_device(root_path):
...
+    """
+    Find the correct path for the cloud-regionsrv-client cache files
+    as the location moves from /var/lib/cloudregister
+    (pre cloud-regionsrv-client-10.0.4) to /var/cache/cloudregister
+    (post cloud-regionsrv-client-10.0.4)
+    """
+    old_cloud_register_path = os.sep.join(
+        [root_path, 'var', 'lib', 'cloudregister']
+    )
+    new_cloud_register_path = os.sep.join(
```
jesusbv, line 277: you do not need two variables

maybe consider this approach ?

```python
    # old path
    cloud_register_path = os.sep.join(
        [root_path, 'var', 'lib', 'cloudregister']
    )
    if os.path.isdir(cloud_register_path) and \
            any('SMT' in x for x in os.listdir(cloud_register_path)):
        return cloud_register_path
    # new path
    cloud_register_path = os.sep.join(
        [root_path, 'var', 'cache', 'cloudregister']
    )
    if os.path.isdir(cloud_register_path) and \
            any('SMT' in x for x in os.listdir(cloud_register_path)):
        return cloud_register_path
```

test/unit/units/prepare_test.py
```diff
@@ -152,7 +185,7 @@ def test_main(
             )
         ]
         mock_Path.call_args_list == [
-            call(['/var/lib/cloudregister']),
+            call(['/var/cache/cloudregister']),
```
jesusbv, line 188: did the test for `/var/lib/cloudregister` disappear ? 
If so, should we not make sure that is still OK ?

suse_migration_services/units/prepare.py
KeithMnemonic: i fixed this, please check

suse_migration_services/units/prepare.py
```diff
@@ -251,3 +262,29 @@ def get_root_disk_device(root_path):
...
+def get_regionsrv_client_file_location(root_path):
+    """
+    Find the correct path for the cloud-regionsrv-client cache files
+    as the location moves from /var/lib/cloudregister
+    (pre cloud-regionsrv-client-10.0.4) to /var/cache/cloudregister
+    (post cloud-regionsrv-client-10.0.4)
+    """
+    # cloud_register_path = os.sep.join(
+    #     [root_path, 'var', 'cache', 'cloudregister']
+    # )
```
rjschwei, line 276: Please remove these comments

suse_migration_services/units/prepare.py
```diff
@@ -251,3 +262,29 @@ def get_root_disk_device(root_path):
...
+    for cloud_register_path in [
+        os.sep.join([root_path, 'var', 'cache', 'cloudregister']),
+        os.sep.join([root_path, 'var', 'lib', 'cloudregister'])
+    ]:
+        if os.path.isdir(cloud_register_path) and \
+                any('SMT' in x for x in os.listdir(cloud_register_path)):
+            return cloud_register_path
+        cloud_register_path = os.sep.join(
```
rjschwei, line 284: The assignment is superfluous. If we do not find the cache files an exception is raised.

That said on a BYOS instance no such cache would exist and we would trigger the exception. Therefore we need to handle this in an intelligent way.

The problem is that we will have a chicken and egg problem the `registerutils.is_registered()` implementation needs the cache but we are trying to figure out whether the cache should exist and where it may be coming from. We can look at `/etc/hosts` to see if there is an entry for `.susecloud.net` if that exists it is fair to assume that the instance is connected to the update infrastructure and as such we need to look for the cache.

rjschwei (changes requested):

suse_migration_services/units/prepare.py
KeithMnemonic: @rjschwei  thanks I forgot to remove the comment and superfluous stuff after addressing @jesusbv feedback.  If I wrap this check around a check  for "/etc/hosts to see if there is an entry for .susecloud.net " is that sufficient?

KeithMnemonic: for example

```
    with open(hosts_setup, 'r', encoding="utf-8") as fp:
        if 'susecloud.net' in fp.read():
            try:
                cloud_register_metadata = \
                    get_regionsrv_client_file_location(root_path)
            except DistMigrationZypperMetaDataException as issue:
                log.error('Failed locating regionsrv-client cache files: {0}'.format(
                    issue)
                )
                raise
        else:
            cloud_register_metadata = os.sep.join(
                [root_path, 'var', 'lib', 'cloudregister']
            )
```

rjschwei: @KeithMnemonic yes that's sufficient

suse_migration_services/units/prepare.py
KeithMnemonic: removed

suse_migration_services/units/prepare.py
KeithMnemonic: fixed

suse_migration_services/units/prepare.py
```diff
@@ -115,9 +116,22 @@ def main():
...
+        if 'susecloud.net' in fp.read():
+            try:
+                cloud_register_metadata = \
+                    get_regionsrv_client_file_location(root_path)
+            except DistMigrationZypperMetaDataException as issue:
+                log.error('Failed locating regionsrv-client cache files: {0}'.format(
+                    issue)
+                )
+                raise
+        else:
```
rjschwei, line 130: There is no else clause. A system that is not connected to the update infrastructure does not have a cache thus nothing needs to be mounted.

suse_migration_services/units/prepare.py
```diff
@@ -163,12 +177,12 @@ def main():
             zypp_plugins_services, '/usr/lib/zypp/plugins/services'
         )
         if os.path.exists(cloud_register_metadata):
-            log.info('Bind mounting /var/lib/cloudregister')
-            Path.create('/var/lib/cloudregister')
+            log.info('Bind mounting /var/cache/cloudregister')
+            Path.create('/var/cache/cloudregister')
```
rjschwei, line 181: The log message should now tell us where the bind mount is coming from i.e. should include `cloud_register_metadata`

suse_migration_services/units/prepare.py
```diff
@@ -163,12 +177,12 @@ def main():
             zypp_plugins_services, '/usr/lib/zypp/plugins/services'
         )
         if os.path.exists(cloud_register_metadata):
```
rjschwei, line 179: With the change above this changes to `if cloud_register_metadata and os.path.exists(cloud_register_metadata):`

rjschwei (changes requested):

suse_migration_services/units/prepare.py
KeithMnemonic: i get it now

suse_migration_services/units/prepare.py
KeithMnemonic: got it

suse_migration_services/units/prepare.py
KeithMnemonic: this could be more streamlined if I move it all to line  165 for example:  Any objections?
```
        with open(hosts_setup, 'r', encoding="utf-8") as fp:
            if 'susecloud.net' in fp.read():
                try:
                    cloud_register_metadata = \
                        get_regionsrv_client_file_location(root_path)
                    log.info('Bind mounting /var/cache/cloudregister from {0}'.format(
                        cloud_register_metadata)
                    )
                    Path.create('/var/cache/cloudregister')
                    Command.run(
                        [
                            'mount', '--bind', cloud_register_metadata,
                            '/var/cache/cloudregister'
                        ]
                    )
                    update_smt_cache = '/usr/sbin/updatesmtcache'
                    if os.path.isfile(update_smt_cache):
                        log.info('Updating SMT cache')
                        Command.run(
                            [update_smt_cache]
                        )
                except DistMigrationZypperMetaDataException as issue:
                    log.error('Failed locating regionsrv-client cache files: {0}'.format(
                        issue)
                    )
                    raise
 ```

Comment (KeithMnemonic) [reply suse_migration_services/units/prepare.py:179]: I will update it per your changes. f you want me to switch to the above that is fine or we can keep it as is.

File: suse_migration_services/units/prepare.py [hunk 8]
```diff
@@ -92,6 +92,7 @@ def main():
                 Path.create(trust_anchor)
                 for cert in certificates:
                     cert_file = os.sep.join([root_trust_anchor, cert])
+                    print(cert_file)
```
jesusbv, line 95: do we want to show the cert ?
I think we should remove this line

suse_migration_services/units/prepare.py
```diff
@@ -115,9 +116,18 @@ def main():
...
-        [root_path, 'var', 'lib', 'cloudregister']
-    )
+
+    with open(hosts_setup, 'r', encoding="utf-8") as fp:
+        if 'susecloud.net' in fp.read():
+            try:
+                cloud_register_metadata = \
+                    get_regionsrv_client_file_location(root_path)
+            except DistMigrationZypperMetaDataException as issue:
+                log.error('Failed locating regionsrv-client cache files: {0}'.format(
```
jesusbv, line 126: just a bit nitpicking, maybe consider
```python
              log.error(
                    'Failed locating regionsrv-client cache files: {0}'.format(
                          issue
                    )
              )
```
so you keep closing brackets consistently

suse_migration_services/units/prepare.py
```diff
@@ -162,13 +172,14 @@ def main():
         system_mount.add_entry(
             zypp_plugins_services, '/usr/lib/zypp/plugins/services'
         )
-        if os.path.exists(cloud_register_metadata):
-            log.info('Bind mounting /var/lib/cloudregister')
-            Path.create('/var/lib/cloudregister')
+        if cloud_register_metadata and os.path.exists(cloud_register_metadata):
+            log.info('Bind mounting /var/cache/cloudregister from {0}'.format(
```
jesusbv, line 176: same here, 
```python
log.info(
       'Bind mounting /var/cache/cloudregister from {0}'.format(
             cloud_register_metadata
       )
)
```
otherwise, let's try to keep it consistent with the previous log
```python
log.info(
       'Bind mounting /var/cache/cloudregister from {0}'.format(
             cloud_register_metadata)
)
```

suse_migration_services/units/prepare.py
```diff
@@ -251,3 +262,23 @@ def get_root_disk_device(root_path):
...
+    for cloud_register_path in [
+        os.sep.join([root_path, 'var', 'cache', 'cloudregister']),
+        os.sep.join([root_path, 'var', 'lib', 'cloudregister'])
+    ]:
+        if os.path.isdir(cloud_register_path) and \
+                any('SMT' in x for x in os.listdir(cloud_register_path)):
+            return cloud_register_path
+    raise DistMigrationZypperMetaDataException(
+        'No cloud-regionsrv-client cache files found in '
+        '/var/cache/cloudregister or /var/cache/cloudregister'
```
jesusbv, line 283: One of the paths in the exception should be `/var/lib/cloudregister` (I guess copypasta)

suse_migration_services/units/prepare.py
jesusbv, line 176: maybe consider using a var for `/var/cache/cloudregister`
so you could do
```python
log.info("...{}...".format(cache_cloudregister_path),...)
Path.create(cache_cloudregister_path)
Command.run(
          [
                  ...
                   cache_cloudregister_path
          ]
)
```

suse_migration_services/units/prepare.py
```diff
@@ -162,13 +172,14 @@ def main():
         system_mount.add_entry(
             zypp_plugins_services, '/usr/lib/zypp/plugins/services'
         )
-        if os.path.exists(cloud_register_metadata):
-            log.info('Bind mounting /var/lib/cloudregister')
-            Path.create('/var/lib/cloudregister')
+        if cloud_register_metadata and os.path.exists(cloud_register_metadata):
```
jesusbv, line 175: this condition can be
```python
if os.path.exists(path)
```
the double check does not add safety or value,
if code reaches that part and `cloud_register_metadata` has not been populated, checking as it is:
```python
if path and os.path.exists(path)
```
produces an exception same as `if os.path.exists(path)` 
there needs to be a new test similar to `test_main_raises_on_get_regionsrv_client_file_location` but with a different value (for example `foo`) [here](https://github.com/SUSE/suse-migration-services/pull/251/files#diff-244e5e611b27f16d2db5163a4a3806e6352bf26d6ca8fbf2ca4a3bac2bd4ea2bR69) to capture this scenario

adding
```python
cloud_register_metadata = ""
with open(hosts_setup, 'r', encoding="utf-8") as fp:
      ....
```
is enough
```python
cloud_register_metadata = ""
os.path.exists(cloud_register_metadata)
False
```

test/unit/units/prepare_test.py
```diff
@@ -43,10 +46,37 @@ def test_main_raises_on_zypp_bind(
...
+            [DistMigrationZypperMetaDataException('foo')]
+        mock_os_path_exists.return_value = True
+        mock_Command_run.side_effect = [
+            MagicMock(),
+            MagicMock(),
+            MagicMock(),
+            MagicMock(),
+            Exception
+        ]
+        with patch('builtins.open', mock_open(read_data='foo.susecloud.net')):
```
jesusbv, line 69: as mentioned above, there should be a test with a different value that does not include `susecloud.net`

suse_migration_services/units/prepare.py
KeithMnemonic: sorry left that in during debugging a test failure. i will remove it. thanks for catching

suse_migration_services/units/prepare.py
KeithMnemonic: no problem

suse_migration_services/units/prepare.py
KeithMnemonic: will do

suse_migration_services/units/prepare.py
KeithMnemonic: yup, copy paste error, thanks for catching

suse_migration_services/units/prepare.py
KeithMnemonic: i will switch this out.

suse_migration_services/units/prepare.py
KeithMnemonic: i will  `cloud_register_metadata = ""` in the code right after i defined `cache_cloudregister_path = '/var/cache/cloudregister'`

suse_migration_services/units/prepare.py
```diff
@@ -163,12 +177,16 @@ def main():
             zypp_plugins_services, '/usr/lib/zypp/plugins/services'
         )
         if os.path.exists(cloud_register_metadata):
-            log.info('Bind mounting /var/lib/cloudregister')
-            Path.create('/var/lib/cloudregister')
+            log.info(
+                'Bind mounting /var/cache/cloudregister from {0}'.format(
```
jesusbv, line 181: maybe use `cache_cloudregister_path` ?
```python
log.info(
      'Bind mounting {0} from ....'.format(cache_cloudregister_path,...)
)
```

jesusbv (approved):
Thanks for the changes
LGTM

suse_migration_services/units/prepare.py
KeithMnemonic: updated

suse_migration_services/units/prepare.py
```diff
@@ -163,12 +177,16 @@ def main():
...
+                'Bind mounting {0} from {1}'.format(
+                    cache_cloudregister_path, cloud_register_metadata
+                )
+            )
+            Path.create(cache_cloudregister_path)
             Command.run(
                 [
                     'mount', '--bind', cloud_register_metadata,
-                    '/var/lib/cloudregister'
+                    cache_cloudregister_path
```
rjschwei, line 189: While the cache directory is created by the `cloud-regionsrv-client` package we should be prepared for the directory not being there. So either create it conditionally or look for and handle the failure of the mount command.

rjschwei (changes requested):

suse_migration_services/units/prepare.py
KeithMnemonic: it gets created one line above  or are you referring to something else? 

``` 
   >>>>Path.create(cache_cloudregister_path)
            Command.run(
                [
                    'mount', '--bind', cloud_register_metadata,
                    cache_cloudregister_path
                ]
            )
```

rjschwei (approved):

PR #254, opened 2022-12-12
Use bind mount to ensure the resolv.conf is not empty

### LINKED ISSUE #253: Output of /usr/bin/suse-migration-pre-checks
Dear Team,

I would like to know if it's possible and feasible to add the output of `/usr/bin/suse-migration-pre-checks` as an example in the documentation in [this part](https://documentation.suse.com/suse-distribution-migration-system/15/single-html/distribution-migration-system/#id-upgrade-pre-checks). For example:

```
sles12-sp5:/etc/zypp # /usr/bin/suse-migration-pre-checks -f
Running suse-migration-pre-checks with options: fix: True
Device path 10.123.45.678:/usrsaptrans/EWM not found and skipped
Device path 10.123.45.678:/symbackup/ewx not found and skipped
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/56574c71-9967-4f8b-b8ff-b2b914de277f']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/dc679d7a-bd33-4f70-8221-878911791f12']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/disk/by-uuid/CA25-C086']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/sap/mnt']
Calling: ['blkid', '-s', 'TYPE', '-o', 'value', '/dev/sap/usrsap']
The config option 'multiversion' in /etc/zypp/zypp.conf includes the keyword 'kernel.' The current value is set as
'multiversion = provides:multiversion(kernel)'.
Checking the config option 'multiversion.kernels' to see if multiple kernels are also enabled
Calling: ['rpm', '-qa', 'kernel-default']
```

The above output is supposed to be interpreted as "it passed, no issues found and you are ready to migrate"?

Thank you.

KeithMnemonic:
The PR changes the behavior of setup_host_network when the resolv.conf in the target system is "empty" (It does not contain any entries.)  The current behavior was to not copy the file over if it was empty.   Due to a recent change in wickedd, when the migration is started via the "Reboot" method, the target /etc/resolv.conf is reset and does not contain any entries.  (This does not happens if the /usr/sbin/run_migration is used.)  This means when the migration iso is booted,  the /system-root/etc/resolv.conf is empty and therefore any commands that require DNS and are run under `chroot /system-root` will fail. 

This change will now "bind mount" /etc/resolv.conf to /system-root/etc/resolv.conf when the `/system-root/etc/resolv.conf` is empty. 

In addition, this PR contains:
- A requested documentation update and 
- Bump version: 2.0.34 → 2.0.35

closes #253

### DIFF
--- a/suse_migration_services/units/setup_host_network.py
+++ b/suse_migration_services/units/setup_host_network.py
@@ -44,6 +44,11 @@ def main():
     log = logging.getLogger(Defaults.get_migration_log_name())
     root_path = Defaults.get_system_root_path()
 
+    system_mount = Fstab()
+    system_mount.read(
+        Defaults.get_system_mount_info_file()
+    )
+
     resolv_conf = os.sep.join(
         [root_path, 'etc', 'resolv.conf']
     )
@@ -51,14 +56,6 @@ def main():
         raise DistMigrationNameResolverException(
             'Could not find {0} on migration host'.format(resolv_conf)
         )
-    if has_host_resolv_setup(resolv_conf):
-        log.info('Copying {}'.format(resolv_conf))
-        shutil.copy(
-            resolv_conf, '/etc/resolv.conf'
-        )
-    else:
-        log.info('Empty {}, continuing without copying it'.format(resolv_conf))
-
     sysconfig_network_providers = os.sep.join(
         [root_path, 'etc', 'sysconfig', 'network', 'providers']
     )
@@ -67,10 +64,6 @@ def main():
     )
     try:
         log.info('Running setup host network service')
-        system_mount = Fstab()
-        system_mount.read(
-            Defaults.get_system_mount_info_file()
-        )
         Command.run(
             [
                 'mount', '--bind', sysconfig_network_providers,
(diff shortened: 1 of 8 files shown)


suse_migration_services/units/setup_host_network.py
```diff
@@ -44,33 +44,27 @@ def main():
...
-        )
-    else:
-        log.info('Empty {}, continuing without copying it'.format(resolv_conf))
-
-    sysconfig_network_providers = os.sep.join(
-        [root_path, 'etc', 'sysconfig', 'network', 'providers']
-    )
-    sysconfig_network_setup = os.sep.join(
-        [root_path, 'etc', 'sysconfig', 'network', '*']
-    )
```
rjschwei, line 67: There's no need to move the variable assignment into the `try` block. These will not fail and from a code locality perspective it is only 1 line different, but it makes the `try` block significantly longer due to the formatting. As such distracting what the `try` block is actually supposed to catch.

suse_migration_services/units/setup_host_network.py
```diff
@@ -88,6 +82,28 @@ def main():
         Command.run(
             ['systemctl', 'reload', 'network']
         )
+        sysconfig_network_providers = os.sep.join(
+            [root_path, 'etc', 'sysconfig', 'network', 'providers']
+        )
+        sysconfig_network_setup = os.sep.join(
+            [root_path, 'etc', 'sysconfig', 'network', '*']
+        )
```
rjschwei, line 90: Probably cut and paste issue, re-assignment of `sysconfig_network_providers` and `sysconfig_network_setup`, not necessary.

rjschwei (changes requested):

suse_migration_services/units/setup_host_network.py
KeithMnemonic: Thanks for catching this.  I adjusted

suse_migration_services/units/setup_host_network.py
KeithMnemonic: Thanks for catching this.  I removed it

rjschwei (approved):

PR #260, opened 2023-06-01
Add missing metadata for docs.suse.com

tomschr:
# Situation

The document at https://documentation.suse.com/suse-distribution-migration-system/15/single-html/distribution-migration-system/ is publicly available, but misses "Edit source" and "Report bug" links.

The SLE documentation contains some metadata which is used to create these links.

# Solution

This change adds metadata to create "Report bug" links for each title. This is how it looks like when you hover the mouse:

Screenshot_20230601_160748

This is how it looks, @amach4.


KeithMnemonic (approved):
LGTM

PR #262, opened 2023-07-07
Add update_bootloader service and cache file location debug

KeithMnemonic:
This PR adds an update_bootloader service to ensure a proper bootloader is installed after the system migration is complete. This resolves an issue where the following error may occur when rebooting the migrated node. 

```
Loading Linux 4.12.14-150100.197.142-default ...
error: symbol `grub_efi_allocate_fixed' not found.
Loading initial ramdisk ...
error: symbol `grub_efi_allocate_fixed' not found.
```

In addition, some additional debug was added to the prepare module to provide better output for troubleshooting issues that may occur when a system has been registered with the cloudregionsrv-client files and SUSEConnect.   This issue is related to:

https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/118

The additional debug provides help in identifying the issue.

Bump version: 2.0.36 → 2.0.37

PR #263, opened 2023-07-10
update spec template file to include update_bootloader service

KeithMnemonic:
This commit adds an entry to the suse-migration-services-spec-template file to install the new update_bootloader service. This was inadvertently omitted in #262


rtamalin (approved):
LGTM

PR #264, opened 2023-07-10
Add entries to setup.py and the spec file

KeithMnemonic:
This PR updates the setup.py and spec file to add a few more missing entries needed when adding a new service.  The update_bootloader service was added in #262 and #263 addressed one aspect of adding a new service. This PR completes the update.


rtamalin (approved):
LGTM

rjschwei (approved):

PR #269, opened 2023-07-18
Fixing variablelists that caused validation errors

fsundermeyer:
Complex list elements consisting of more than one block may not contain empty lines, you need to concatenate the blocks with a single "+" on a line. Did that for the variablelists, which otherwise caused an error when building html/pdf.

Also added block markeres to the listings which solved the problem of a superflous "+" in the output for me.


KeithMnemonic (approved):
LGTM

jeremy-moffitt (approved):
LGTM... @KeithMnemonic I dont think Frank has merge permissions, you may need to merge the request. Thank you!

PR #270, opened 2023-07-24
Clarify supported versions of SLES

jeremy-moffitt:
Update the prerequisites section to specify that SLES12 SP4 and SLES12 SP5 are the supported versions to migrate from, and that SLES15 SP1 is the supported target version of the migration.


jeremy-moffitt:
going to redo this, my fork thought it was current but it is not... will open a new PR once its ready

PR #274, opened 2023-08-08
Added mailmap

schaefi:
Don't lose contributions of the past ;)


schaefi:
Hi All, since I did a lot of the work on this project from the past it feels odd to be no longer listed as a contributor. Thanks much

PR #279, opened 2024-02-01
Bind mount the configured certs location bsc#1219004

rtamalin:
When the migration is running ensure that the correct certs directory is bind mounted from the /system-root into the ISO boot's runtime environment, by retrieving the configured server.certlocation setting from the /system-root/etc/regionserverclnt.cfg file.

Add unit tests for new get_regionsrv_certs_path() function and tweak some existing tests, that depend upon mocking calls to builtin.open(), so that they mock any calls to get_regionsrv_certs_path() so as to avoid triggering additional open() calls that would break those tests.

### DIFF
--- a/suse_migration_services/units/prepare.py
+++ b/suse_migration_services/units/prepare.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
+import glob
 import logging
 import os
 import shutil
@@ -67,9 +68,7 @@ def main():
         '/etc/pki/trust/anchors/'
     ]
     cache_cloudregister_path = '/var/cache/cloudregister'
-    cloud_register_certs_path = '/var/lib/regionService/certs'
-    cloud_register_certs_bind_mount_path = \
-        root_path + cloud_register_certs_path
+    cloud_register_certs_path_fallback = '/usr/lib/regionService/certs'
 
     cloud_register_metadata = ""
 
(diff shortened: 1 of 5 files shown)


guangyee (approved):
lgtm

smarlowucf (approved):
LGTM, although I'm not totally familiar with cloud reg client code.

jesusbv (approved):
LGTM

suse_migration_services/units/prepare.py
```diff
@@ -67,9 +67,7 @@ def main():
         '/etc/pki/trust/anchors/'
     ]
     cache_cloudregister_path = '/var/cache/cloudregister'
-    cloud_register_certs_path = '/var/lib/regionService/certs'
-    cloud_register_certs_bind_mount_path = \
-        root_path + cloud_register_certs_path
+    cloud_register_certs_path_fallback = '/var/lib/regionService/certs'
```
jesusbv, line 70: note to keep in mind that path is valid for SLES 12
that path is different for SLES15 (s/var/usr `/usr/lib/regionService/certs`)

rtamalin: This path is valid for SLE 12 SP5 pre the most recent update (v2.0.1) to the region service tools, that switched the path to /usr/lib, then same as it is for SLE 15 streams.

If we feel it would be better to default to the new path, I'm happy to do so. ;-)

rjschwei: The default should be the new location. The user **must** update their system to the latest packages before migration. During the upgrade the certs will be moved to the new location, as such that's where we should expect them. I think we should add a _warning_ if we cannot find any `*.pem` files in whatever location `cloud_register_certs_path` returns. We cannot exit he process because the instance my be BYOS and it maybe registered to SCC in which case the cert path may not exist in the first place, but migration is still expected to work. Please insert a warning.

rjschwei (changes requested):

suse_migration_services/units/prepare.py
rjschwei: And let me change the type of the message, it should be an INFO message not a warning.

rtamalin: I've changed the fallback default path to the /usr/lib one, and the code now logs an info message if no pem files are detected at the configured cert location under /system-root, or if no certs are detected after bind mounting the cert location path into the ISO runtime environment.

rjschwei (approved):

rtamalin:
I'm going to squash the commits into as single commit before merging; helps with keeping the changes log management simpler ;-)

PR #280, opened 2024-02-15
Update DMS to support migrating directly to SLE15 SP3

yarunachalam:
Update DMS to support migrating directly to SLE15 SP3


schaefi:
pubcloud/sle15/config.kiwi is already at SP3, this one can be closed

PR #281, opened 2024-02-27
Update DMS to support migrating directly to SLE15 SP5

yarunachalam:
Update DMS to support migrating directly to SLE15 SP5


schaefi:
closed on behalf of #298

PR #283, opened 2024-04-03
Fix update-bootloader to run after grub-setup  bsc#1222258

rtamalin:
Also checked to ensure that the sequencing for all systemd services now matches the ordering specified in helper/system.tree.


jesusbv (approved):
LGTM

rjschwei (approved):

PR #285, opened 2024-08-08
Fix update-bootloader to run after grub-setup  bsc#1222258

rtamalin:
Also checked to ensure that the sequencing for all systemd services now matches the ordering specified in helper/system.tree.


rtamalin:
Closing as a duplicate of #283

PR #286, opened 2024-08-08
Remove custom legacy ruby based zypper-migration script

rtamalin:
The install of the legacy zypper-migration-plugin script was previously removed from the kiwi file.


aosthof (approved):
LGTM

rjschwei (approved):

PR #296, opened 2024-10-20
Update README

schaefi:
Fix contents of the README, point to the SUSE docs and fix symlinks and other information


README.rst
```diff
@@ -0,0 +1,64 @@
+Distribution Migration System
+=============================
+
+.. |GitHub Action Unit Types| image:: https://github.com/SUSE/suse-migration-services/actions/workflows/ci-testing.yml/badge.svg
+   :target: https://github.com/SUSE/suse-migration-services/actions
+.. |Doc| replace:: `Documentation <https://documentation.suse.com/suse-distribution-migration-system/15/html/distribution-migration-system/index.html>`__
+
+|GitHub Action Unit Types|
+
+**Migrate from one major SUSE version to another with the DMS**
```
rjschwei, line 10: With SLE 16 on the horizon and the migration from 15 to 16 not being determined yet it is probably a good idea to make this more specific at this point and change the heading to "Migrate a SUSE distribution from 12 to 15 with DMS"

rjschwei (approved):

README.rst
schaefi: good point, I'll change that

schaefi: done

PR #297, opened 2024-10-20
Add success indicator flag file

schaefi:
Added /var/log/distro_migration.exitcode to be written by the migrate service. The file contains the exit code from the zypper process and is used in the migrate tool to know about the overall state of the migration process. The migrate tool initializes the flag file with a failed value.


schaefi:
This additional information helps to identify the overall state of the migration and helps with debugging as well as serves as indicator in a container based upgrade workflow when there is no reboot involved.

rjschwei (approved):

PR #298, opened 2024-10-20
Update DMS live ISO product to SLE15 SP5

schaefi:
Allow to upgrade directly to SP5

### DIFF
--- a/image/pubcloud/sle15/config.kiwi
+++ b/image/pubcloud/sle15/config.kiwi
@@ -27,16 +27,28 @@
         <source path='obsrepositories:/'/>
     </repository>
     <repository type="rpm-md">
-	 <source path="obs://SUSE:SLE-15-SP3:Update/standard"/>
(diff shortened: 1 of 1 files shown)


schaefi:
Needs testing first

rjschwei:
We are going to stick with 15 SP3 based on the conversation with PM. We also need to serve our more conservative customers. If someone wants to move to 15 SP3 and they end up first at 15 SP5 because it is the default they have a high risk of breaking their system becuase downgrades are more dangerous than upgrades. If a customer wants to get to 15 SP5 but ends up on 15 SP3 because that's the default setting they still have a supported path forward. The agreement with PM was that next year when 15 SP7 gets released we'd move forward to 15 SP4 and then 15 SP5 the year after. Then SLE 12 will reach EOL.

schaefi:
ok thanks for clarification, then this one can be closed

PR #299, opened 2024-10-21
Fix/Refactor DMS packaging

schaefi:
* Add tox target to test implementation against py 3.6 and 3.11 stack Actually also wanted to add py3.4 but none of the github runners offers such an old python stack

* Add mypy tests. This is enabled to make sure we get notified if static type checking is enabled in the code. In this case we will be able to check if the affected code part is allowed to use type checking because the pre-checks python code must not use this feature of python as it is expected to run on SLE12 with py3.4

* Fixup the overall python packaging. The former version of the package stack duplicates the module code in conflicting packages and also packaged python byte code binaries. This commit refactors the packaging such that the python module code is packages as a proper %{pythons}-migration package which can build for different python versions. The spec file change is done in a similar way to allow SLE12 and SLE15 builds as we do it for other PCT python modules. The respective DMS packages suse-migration-services, suse-migration-pre-checks can now simply require the DMS python stack and package only the main entry-point commands and services. This also allows to add further implementations to the DMS without ending in a package nightmare

* Fixup unit tests to finally get back to 100% test coverage as it was before I left the project. I'm not sure who tought it would be a good idea to weaken --cov-fail-under

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -3,24 +3,28 @@ name: CILint
 on: [push]
 
 jobs:
-  build:
-
-    #runs-on: ubuntu-latest
-    runs-on: ubuntu-20.04
+  unit_tests:
+    name: Unit tests for DMS
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.8]
+        python-version: ["3.11"]
+        os: [ubuntu-latest]
+        include:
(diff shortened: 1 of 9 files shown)


schaefi:
The entire stack including this change builds here for double checking:

* https://build.opensuse.org/project/show/home:marcus.schaefer:dms:live

I have done intensive testing on the stack regarding regressions.

(Ignore the container based packages you will find there too, it's for the nextgen PR for which we still need to find a home and proper wording as it's outside of the documented offline process)

PR #303, opened 2024-10-30
Install migration live image to /migration-image

### LINKED ISSUE #277: Workaround for LVM support
Recently while testing DMS to help a customer migrate its VMs,  I noticed that most of their VMs use a complex layout with LVM, and thus DMS would fail (as described in the documentation). 

The issue is that GRUB cannot find the ISO after rebooting, as their /usr lives on a separate LV, and the logic in 99_migration would evaluate the final path as "/share/migration-image/<ISO name>", as it assumes /usr is under the root partition.

I managed to get it working with a workaround. Right before rebooting, run these commands:
```
# mkdir /share
# cp -av /usr/share/migration-image /share/

# sed -i '/insmod %s/a printf \"    insmod lvm\\n\"' /etc/grub.d/99_migration 
# grub2-mkconfig -o /boot/grub2/grub.cfg
``` 

So, with these I'm: 
* copying the migration-image directory to the root partition, so the path to the ISO as-is will evaluate to a valid ISO
* adding the "lvm" GRUB module so it can find the LVM root partition 

This will, of course, occupy an extra 400MB or so (the migration ISO image size) on the root partition, but this space can be reclaimed after the migration.

I could have written the proper GRUB logic to find the /usr partition, but I think this could vary quite a lot. Maybe we could leave the ISO on a different directory at the root partition, like I did with the workaround? Something like "/migration-image" is very unlikely to exist as a separate LV on any setup.

schaefi:
This change helps to run migrations on systems which stores /usr/share in a special way, e.g. as LVM volume or other type that cannot be read via the loopback grub or kexec method. This Fixes #277

### DIFF
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -14,7 +14,7 @@ else
     ) ${CLASS}"
 fi
 
-migration_iso=$(echo /usr/share/migration-image/*-Migration.*.iso)
+migration_iso=$(echo /migration-image/*-Migration.*.iso)
 
 root_device=$(
     lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | uniq | cut -f1 -d" "
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -74,6 +74,7 @@ if grub_file_is_not_garbage "${migration_iso}"; then
             "/@/boot/grub2/x86_64-efi"
         printf "    insmod loopback\n"
     fi
+    printf "    insmod lvm\n"
     printf "    insmod %s\n" "${image_fs_type}"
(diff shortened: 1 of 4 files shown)


schaefi:
The DMS stack for testing including this change can be found here:

doccaz:
I'd like to add that you need to add a "insmod lvm" to the 99-migration file to recognize LVM volumes.

rjschwei (approved):

schaefi:
> I'd like to add that you need to add a "insmod lvm" to the 99-migration file to recognize LVM volumes.

yes you mean if there is a root volume for (/) we need to load the lvm module such that we can read the image from
(root)/migration-image. I added a commit to capture this case

PR #308, opened 2025-01-16
Add option to create a solver test case

rjschwei:
During the migration we may run into a situation where a package shows up on the migrated system and we need to understand why this happened. Creating a solver test case allows us to debug such issues.


schaefi (approved):
yay this looks nice thanks :+1:

PR #309, opened 2025-01-20
use findmnt instead of blkid for root detection

fcrozat:
it works better with btrfs setup

(this is required for SLES16 migration)

### DIFF
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -16,12 +16,10 @@ fi
 
 migration_iso=$(echo /migration-image/*-Migration.*.iso)
 
-root_device=$(
-    lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | uniq | cut -f1 -d" "
-)
+root_device=$(findmnt -n -v -o SOURCE /)
 
-image_fs_uuid=$(blkid -s UUID -o value "${root_device}")
-image_fs_type=$(blkid -s TYPE -o value "${root_device}")
+image_fs_uuid=$(findmnt -n -o UUID /)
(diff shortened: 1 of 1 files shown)


rjschwei (approved):

schaefi:
@fcrozat I have concerns with this change. I know findmnt has issues on SLE12 which is one target that it would run on. Has this change been integration tested on SLE12 systems ?

fcrozat:
> @fcrozat I have concerns with this change. I know findmnt has issues on SLE12 which is one target that it would run on. Has this change been integration tested on SLE12 systems ?

No, I only tested on SLE15 to SLE16 migration. The current code doesn't handle at all btrfs subvolumes layout we have in SLE15.

schaefi:
ok so I will take your change and test SLE12 migrations, if they still work we are good

schaefi:
ok testing with the change in the cloud and in my test VM worked. @fcrozat thanks for the change, I'll merge and submit a new DMS build to my home:marcus.schaefer:dms build

Please note, we currently have a DMS submission running which is not yet processed. I assume you don't need a submission of the packages to some SLE15 target immediately ?

fcrozat:
> ok testing with the change in the cloud and in my test VM worked. @fcrozat thanks for the change, I'll merge and submit a new DMS build to my home:marcus.schaefer:dms build

Thanks

> Please note, we currently have a DMS submission running which is not yet processed. I assume you don't need a submission of the packages to some SLE15 target immediately ?

Not yet. I have many other features branches I'll submit one by one and we can then discuss on a SLE15 target.

PR #310, opened 2025-01-23
Increase reliability of host network setup

fcrozat:
Handle if /etc/resolv.conf doesn't exist on system to migrate or if it is a symlink

This is required to migrate SLE 15 system and should be safe on SLE12 ones (untested)

### DIFF
--- a/suse_migration_services/units/setup_host_network.py
+++ b/suse_migration_services/units/setup_host_network.py
@@ -52,10 +52,6 @@ def main():
     resolv_conf = os.sep.join(
         [root_path, 'etc', 'resolv.conf']
     )
-    if not os.path.exists(resolv_conf):
-        raise DistMigrationNameResolverException(
-            'Could not find {0} on migration host'.format(resolv_conf)
-        )
     sysconfig_network_providers = os.sep.join(
         [root_path, 'etc', 'sysconfig', 'network', 'providers']
     )
@@ -64,15 +60,16 @@ def main():
     )
     try:
         log.info('Running setup host network service')
-        Command.run(
-            [
-                'mount', '--bind', sysconfig_network_providers,
-                '/etc/sysconfig/network/providers'
-            ]
-        )
-        system_mount.add_entry(
-            sysconfig_network_providers, '/etc/sysconfig/network/providers'
-        )
+        if os.path.exists(sysconfig_network_providers):
+            Command.run(
+                [
+                    'mount', '--bind', sysconfig_network_providers,
+                    '/etc/sysconfig/network/providers'
+                ]
+            )
+            system_mount.add_entry(
+                sysconfig_network_providers, '/etc/sysconfig/network/providers'
+            )
         for network_setup in glob.glob(sysconfig_network_setup):
             if os.path.isfile(network_setup):
                 shutil.copy(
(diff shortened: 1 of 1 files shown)


suse_migration_services/units/setup_host_network.py
```diff
@@ -87,7 +84,14 @@ def main():
                 resolv_conf, '/etc/resolv.conf'
             )
         else:
-            log.info('Empty {0}, bind mounting /etc/resolv.conf to {0}'.format(resolv_conf))
+            # ensure resolv.conf won't be a symlink on migrated system, create an empty file instead to allow bind mount
+            if os.path.islink(resolv_conf):
+                log.info('{0} is a symlink, renaming it to {0}.pre-migration, bind mounting /etc/resolv.conf to {0}'.format(resolv_conf))
+                os.replace(resolv_conf,resolv_conf+'.pre-migration')
+                open(resolv_conf,'w').close()
```
rjschwei, line 91: But if `resolv.conf` is a symlink then the user could have pointed it to any location that has resolution information. If we stick an empty file in it's place there is a good chance that network access will not work.

fcrozat: By default on SLES15, /etc/resolv.conf is a symlink to /run/netconfig/resolv.conf (in tmpfs), which is not available on SLES16. This is why we need to move the symlink away (we can't bind mount on top of a symlink).

On SLES16, /etc/resolv.conf is no longer a symlink but a file, which is managed by NetworkManager. If the file is empty, it will replaced by NetworkManager by the information from DHCP (or NM configuration).

rjschwei: Shouldn't we at least check if the symlink points to the default location? If it doesn't we should probably not write an empty file and instead copy the content from the pointed to file.

fcrozat: Indeed, I'll add some checks to see if the symlink is resolvable (ie not pointing to /run). If it is, it should be left intact, if not, we could do the renaming.

WDYT ?

schaefi: The proposed solution will not work for non NetworkManager controlled networks. For example if you use systemd-networkd and systemd-resolved, resolv.conf is a symlink to a different location than the default one. I would really go that far to check first if `/etc/resolv.conf` belongs to a package which it usually does. `sysconfig-netconfig` at that point I also believe it should be a task of the package to handle resolv.conf properly according to the new target (SLE16). I strongly recommend to not mess with resolv.conf as part of the DMS code. It's should actually not be a job of the DMS to perform such changes.

If we have to mess with resolv.conf then I agree with @rjschwei we should do it only if we know to which network management system it belongs to such that the move is safe.

schaefi: In the current logic if there is no /etc/resolv.conf at all it's a fatal error. Why was this condition deleted ?

schaefi: why should bind mounting a symlink be a problem ?

schaefi: @fcrozat as we control the migration environment I would also be in favor to "fix" the migration live image and the container to provide a proper environment rather than adding code to do this.

fcrozat: I didn't want to bother you with my initial tests but I'm fine following your regular workflow.

PR #311, opened 2025-02-04
ensure /run is available in chroot

fcrozat:
Some symlinks on migrated host can point to temporary files in /run (for instance /etc/resolv.conf, by default on SLE15, points to /run/netconfig/resolv.conf). 

We must ensure /run is available when running chrooted application on migrated host.


fcrozat:
This PR replaces https://github.com/SUSE/suse-migration-services/pull/310

schaefi (approved):

PR #312, opened 2025-02-04
use requisite for systemd units

fcrozat:
ensure some migration tasks are only executed if previous one were succesfull

### DIFF
--- a/systemd/suse-migration-grub-setup.service
+++ b/systemd/suse-migration-grub-setup.service
@@ -1,10 +1,12 @@
 [Unit]
 Description=Recreate grub configuration file from migrated version
 After=suse-migration.service
+Requisite=suse-migration.service
 
 [Service]
 Type=oneshot
 ExecStart=/usr/bin/suse-migration-grub-setup
+RemainAfterExit=true
 
 [Install]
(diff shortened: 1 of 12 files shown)


rjschwei (approved):

systemd/suse-migration-grub-setup.service
```diff
@@ -1,10 +1,12 @@
 [Unit]
 Description=Recreate grub configuration file from migrated version
 After=suse-migration.service
+Requisite=suse-migration.service
```
schaefi, line 4: I don't think you want that. The job of the grub-setup service is to uninstall the Migration live image as well as the activation package and produce a grub config that can boot the system. This action is valid in a successful and in a failed state. This is because if you install the activation package but you never return to a grub setup that can boot into the system, a reboot of either a successful migrated system or a failed upgrade will be able to boot into anything else than the migration live system.

That's why I never set any requires to the grub-setup service. It has to run. The only condition is it, has to run after the suse-migration has been tried

schaefi (changes requested):
Looks all good except for grub-setup imho

schaefi (approved):

PR #313, opened 2025-02-04
Create a SLE 16 migration image

fcrozat:


rjschwei (approved):
LGTM

schaefi (changes requested):
Great work :+1:  see my comment

package/suse-migration-sle16-activation-spec-template
```diff
@@ -0,0 +1,60 @@
...
+Url:              https://github.com/SUSE/suse-migration-services
+Summary:          SUSE Distribution SLE16 Migration Activation
+License:          GPL-3.0+
+Group:            System/Management
+Source:           99_migration
+BuildRoot:        %{_tmppath}/%{name}-%{version}-build
+BuildArch:        noarch
+Conflicts:        suse-migration-services
+BuildRequires:    grub2
+Requires:         SLES15-Migration >= 2.1.2
```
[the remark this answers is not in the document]
fcrozat: missed in search and replaced. fixed

image/generic/sle16/config.sh
```diff
@@ -0,0 +1,93 @@
...
+# COPYRIGHT     : (c) 2023 SUSE Software Solutions Germany GmbH
+#               :
+# AUTHORS       : Marcus Schaefer <ms@suse.de>
+#               : Keith Berger <kberger@suse.com>
+#               :
+# BELONGS TO    : Operating System images
+#               :
+# DESCRIPTION   : configuration script for SUSE based
+#               : operating systems
+#----------------
```
schaefi, line 15: Please drop this header. It looks like cut&paste and also provides wrong author information ;)

image/generic/sle16/config.sh
```diff
@@ -0,0 +1,93 @@
...
+#======================================
+# Functions...
+#--------------------------------------
+test -f /.kconfig && . /.kconfig
+test -f /.profile && . /.profile
```
schaefi, line 20: Please drop the reading of .profile. We do that in kiwi automatically for every script call

Also please add

```
set -eux
```

At the beginning of the script code, we want to fail on errors and not silently ignore them

image/generic/sle16/config.sh
```diff
@@ -0,0 +1,93 @@
...
+#======================================
+# Deactivate services
+#--------------------------------------
+# cloud-regionsrv-client packages(s) activates that service
+# on install. For migration we don't need this service.
+# The installation of that packages and its relatives happens
+# to satisfy the dependencies on the zypper plugin side but
+# not for actually registering a guest to the public cloud
+# infrastructure. We expect that step to be already done
+suseRemoveService guestregister
```
schaefi, line 46: Please drop the use of suseRemoveService, it's deprecated and will be removed in newer versions of kiwi

On a general advice, do not use any of the functions that we provide via `test -f /.kconfig && . /.kconfig` We want to get rid of any of those `suseXXX` methods. 

In the above case call systemctl directly

image/generic/sle16/config.sh
```diff
@@ -0,0 +1,93 @@
...
+systemctl disable lvm2-lvmetad.service
+systemctl mask lvm2-lvmetad.socket
+systemctl mask lvm2-lvmetad.service
+
+#======================================
+# Activate services
+#--------------------------------------
+suseInsertService sshd
+suseInsertService haveged
+
```
schaefi, line 59: same here, no use of suseXXX methods please

image/generic/sle16/config.sh
```diff
@@ -0,0 +1,93 @@
...
+suseInsertService suse-migration-prepare
+suseInsertService suse-migration
+suseInsertService suse-migration-console-log
+suseInsertService suse-migration-grub-setup
+suseInsertService suse-migration-update-bootloader
+suseInsertService suse-migration-product-setup
+suseInsertService suse-migration-regenerate-initrd
+suseInsertService suse-migration-kernel-load
+suseInsertService suse-migration-reboot
+suseInsertService NetworkManager
```
schaefi, line 77: same here

schaefi (changes requested):
Sorry for the late replay, too many things going on.

I found a few other places which I believe should be changed. Thanks

PR #314, opened 2025-02-04
Apparmor to selinux

fcrozat:
On SLES 16, apparmor is no longer supposed.

If package migration was succesful, change grub default configuration from apparmor to selinux on migrated host, before updating grub.cfg


fcrozat:

schaefi:
I believe this requires #313 to be merged first

schaefi (approved):

PR #315, opened 2025-02-04
Wicked to NetworkManager migration

fcrozat:
On SLES16, wicked is no longer supported.

If packages migration was succesful, wicked systemd service is disabled and NetworkManager systemd service is enabled instead.

There is no configuration migration done here. This will be handled by another team.


fcrozat:

schaefi:
I believe this requires #313 to be merged first

PR #316, opened 2025-02-04
Create btrfs snapshots before and after migration

fcrozat:
If system to be migrated has btrfs for root fs and snapper configured, create snapshots before and after migration, somehow similar to the one created by zypper migration.

The only difference is, if migration was started by installing suse-migration-service-sle16-activation, the pre migration snapshot will correspond to the snapshot before suse-migration-services rpms are installed on the system to be migrated. Due to this handling, we are not able to use pre/post snapper snapshot types.


fcrozat:

package/suse-migration-services-spec-template
```diff
@@ -174,6 +180,8 @@ install -D -m 755 tools/migrate \
 %files
 %{_sbindir}/run_migration
 %{_bindir}/suse-migration-ssh-keys
+%{_bindir}/suse-migration-btrfs-snapshot-pre-migration
+%{_bindir}/suse-migration-btrfs-snapshot-post-migration
```
schaefi, line 184: I like it :+1:

rjschwei (approved):

PR #319, opened 2025-02-19
ensure suse-migration-apparmor-selinux is build with poetry

fcrozat:
Ensure selinux migration tool is build when using poetry


package/suse-migration-services-spec-template
```diff
@@ -32,6 +32,9 @@ Source:           suse-migration-services.tar.gz
 BuildRoot:        %{_tmppath}/%{name}-%{version}-build
 BuildRequires:    fdupes
 BuildRequires:    python-rpm-macros
+%if 0%{?suse_version} >= 1600
+BuildRequires:    %{pythons}-poetry
```
dirkmueller, line 36: why is `poetry-core` not enough? according to pyproject it depends on poetry-core.

fcrozat: Indeed, fixed

PR #320, opened 2025-02-28
Fix update bootloader

fcrozat:
* Ensure bootloader update is always run in case of migration failure
* only install shim on UEFI system
* enforce path for update-bootloader call, it can results in no file found otherwise.

### DIFF
--- a/suse_migration_services/units/update_bootloader.py
+++ b/suse_migration_services/units/update_bootloader.py
@@ -17,6 +17,7 @@
 #
 """systemd service to install the shim package and update the bootloader"""
 import logging
+import os.path
 
 # project
 from suse_migration_services.command import Command
@@ -38,10 +39,11 @@ def main():
 
     root_path = Defaults.get_system_root_path()
 
-    log.info('Installing the shim package')
-    install_shim_package(root_path)
-    log.info('Updating the shimbootloader')
-    install_secure_bootloader(root_path)
+    if os.path.exists('/sys/firmware/efi'):
+        log.info('Installing the shim package')
+        install_shim_package(root_path)
+        log.info('Updating the shimbootloader')
+        install_secure_bootloader(root_path)
     log.info('Updating the bootloader')
     update_bootloader_config(root_path)
 
--- a/suse_migration_services/units/update_bootloader.py
+++ b/suse_migration_services/units/update_bootloader.py
@@ -112,13 +112,13 @@ def update_bootloader_config(root_path):
 
     try:
         log.info(
-            'Running chroot {0} update-bootloader --reinit'.format(root_path)
+            'Running chroot {0} /sbin/update-bootloader --reinit'.format(root_path)
         )
         Command.run(
             [
                 'chroot',
                 root_path,
-                'update-bootloader',
+                '/sbin/update-bootloader',
                 '--reinit'
             ]
         )
--- a/systemd/suse-migration-btrfs-snapshot-post-migration.service
+++ b/systemd/suse-migration-btrfs-snapshot-post-migration.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Snapper btrfs snapshot after migration
-After=suse-migration.service
+After=suse-migration.service suse-migration-apparmor-selinux.service suse-migration-wicked-networkmanager.service
 Requisite=suse-migration.service
 Before=suse-migration-grub-setup.service
 ConditionPathExists=/run/suse_migration_snapper_btrfs_pre_snapshot_number
--- a/systemd/suse-migration-reboot.service
+++ b/systemd/suse-migration-reboot.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Reboot System
-After=suse-migration-kernel-load.service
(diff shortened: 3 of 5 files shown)


systemd/suse-migration-reboot.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Reboot System
-After=suse-migration-kernel-load.service
+After=suse-migration-kernel-load.service suse-migration-update-bootloader.service
```
rtamalin, line 3: This probably should have an accompanying "documentation" change in https://github.com/SUSE/suse-migration-services/blob/master/helper/service.tree

fcrozat: This part was already in the "documentation", but not really enforced. I've updated service.tree file anyway with more services.

systemd/suse-migration-btrfs-snapshot-post-migration.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=Snapper btrfs snapshot after migration
-After=suse-migration.service
+After=suse-migration.service suse-migration-apparmor-selinux.service suse-migration-wicked-networkmanager.service
 Requisite=suse-migration.service
-Before=suse-migration-grub-setup.service
+Before=suse-migration-grub-setup.service
```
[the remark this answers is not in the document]
fcrozat: fixed

schaefi (approved):
Looks good to me :+1:

suse_migration_services/units/update_bootloader.py
```diff
@@ -38,10 +39,11 @@ def main():
...
 
-    log.info('Installing the shim package')
-    install_shim_package(root_path)
-    log.info('Updating the shimbootloader')
-    install_secure_bootloader(root_path)
+    if os.path.exists('/sys/firmware/efi'):
+        log.info('Installing the shim package')
+        install_shim_package(root_path)
+        log.info('Updating the shimbootloader')
+        install_secure_bootloader(root_path)
```
rjschwei, line 46: This means that if a customer has an instance running from an older image and upgrades they will not be able to switch to an instance type that supports UEFI boot. While images in the Public Cloud have to have settings that indicate whether or not the image supports UEFI I am wondering if we are not making it more difficult for a customer to switch to UEFI instances.

schaefi: Hmm, if you start an instance in the cloud the selection if this instance boots via legacy or via EFI is been made at instance creation time. Depending on that selection the kernel knows about `/sys/firmware/efi` or not. This means if we DMS upgrade an EFI capable system we are technically able to perform secure boot(shim) and set it up. If the system is technically not able to do it because e.g booted in BIOS mode, no shim settings will be made

To my knowledge it is not possible to switch the boot method in the cloud for already created instances. If you start/stop them they will boot in the same way as it was configured in the cloud backend at instance creation time. This is also the reason why the DMS in offline upgrade mode boots into the live system via grub loopback in the same way as the instance itself (just a reboot with different data)

As the DMS can only upgrade existing instances I wonder how we would be able to change the boot method after the upgrade has been performed ?

If you upgrade an old image that has been booted in BIOS mode and expect this instance now to boot as secure boot instance, you would probably be disappointed as I don't think AWS would flip to it automatically. I believe the same applies to all other CSPs. The boot method is a fix selection at instance creation time or even hard wired into the image at image creation time.

From my perspective there are the following operation modes

* Image capable for EFI and legacy BIOS boot, instance boot up in EFI mode, DMS will upgrade to EFI + shim
* Image capable for EFI and legacy BIOS boot, instance boot up in CSM mode (bios), DMS will keep this and don't setup shim
*  Image not EFI capable (disk layout improper), instance must be legacy, DMS will keep this

If you say the boot method may change between boot cycles of the same instance you are right and we have to change the code

Thoughts ?

rjschwei: >To my knowledge it is not possible to switch the boot method in the cloud for already created instances. If you start/stop them they will boot in the same way as it was configured in the cloud backend at instance creation time.

This is not correct. In Google for example there is an explicit switch that can be set for pretty much any instance type. A user can start the instance and then realize later that they forgot to turn on the switch for UEFI secrure boot. The instance can then be stopped, the switch can be turned on and the instance will boot with UEFI secure boot. In AWS this is driven by a flag on the image and the instance type being used. All our images are set to `uefi-preferred`. This means on instances that have UEFI support the instance will boot with UEFI but on instance types that only support BIOS boot, and there are some left the instance will boot using BIOS. If the instance type is later changed to an instance that supports UEFI the instance will boot using UEFI. A similar transition is possible in Azure.

During upgrade wee need to preserve the capability of instance type switching with different firmware. This change breaks this capability.

rjschwei: > If you say the boot method may change between boot cycles 

Yes that is what I am saying

fcrozat: Ok, this is not how it is done on baremetal. Feel free to revert this PR.

schaefi: ok I reverted this part

PR #323, opened 2025-04-02
Add LTSS pre check

schaefi:
Check if the system to migrate has LTSS enabled and print an error message if it is the case

### DIFF
--- a/suse_migration_services/prechecks/ltss.py
+++ b/suse_migration_services/prechecks/ltss.py
@@ -1,0 +1,44 @@
+# Copyright (c) 2025 SUSE LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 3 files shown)


schaefi:
@rjschwei I believe the error message can be better and maybe just print the command to disable LTSS via SUSEConnect. I don't know the exact command maybe you can help me improving the message ? Thanks in advance

I don't want the pre-check to fix it automatically though, that is imho too intrusive

suse_migration_services/prechecks/ltss.py
```diff
@@ -0,0 +1,44 @@
...
+    if migration_system:
+        system_root = Defaults.get_system_root_path()
+
+    ltss_prod = os.path.normpath(
+        '{}/etc/products.d/SLES-LTSS.prod'.format(system_root)
+    )
+
+    if os.path.exists(ltss_prod):
+        log.error(
+            'LTSS is enabled and must be disabled for the migration'
```
rjschwei, line 43: I think it would be nice to form the product triplet the user will need to remove LTSS. With that we could structure the message as follows:
"""
LTSS is a version specific extension and cannot be migrated. To enable migration remove the LTSS add-on product using the appropriate command with the following product triplet $PRODUCT_TRIPLET
"""

jesusbv:
we could check for `available` and `not free` for the list of products
```bash
jesus-test-payg:/home/jesusbv > suseconnect -l --json
......
......
"name":"SUSE Linux Enterprise Server LTSS 12 SP5 x86_64","activated":true,"available":true,"free":false,"extensions":[]}]}
```

instead of checking a hardcoded value in `/etc/products.d` 

Thoughts ?

schaefi:
I had a long conversation with Tom and Alex from the SCC team and I believe we can do a much better pre-check which does not check for a specific installed product but actually performs the API call to the /connect/systems/products/offline_migrations API.

I'm closing this one and will come up with a new PR

PR #324, opened 2025-04-03
Fix and cleanup update_bootloader service

schaefi:
The service tries to install the shim package by a chroot call into the upgraded system. This is not going to work for several reasons. First and foremost the call is interactive and was missing several zypper options to allow this to be done non-interactive. Second zypper from the DMS host should be called in the same way as it was done during the actual migrate service for consistency and stability reasons. When zypper has worked during the migrate service it's likely to work for subsequent calls too. In addition to the actual zypper call I reduced the code at several stages. There is no need to catch an exception from Command.run to raise the same exception (DistMigrationCommandException) that would be done by Command.run itself. Also the command interface logs the executed command. There is no reason to log the same information again. This reduces the implementation by a number of superfluous code and test code.

### DIFF
--- a/suse_migration_services/units/update_bootloader.py
+++ b/suse_migration_services/units/update_bootloader.py
@@ -23,10 +23,6 @@ from suse_migration_services.command import Command
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import Logger
 
-from suse_migration_services.exceptions import (
-    DistMigrationCommandException
-)
-
 
 def main():
     """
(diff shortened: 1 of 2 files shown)


schaefi:
I updated 

with this patch and will give the migration from SLE12 to SLE15 a try

rjschwei (approved):

schaefi:
Ok migration succeeded

```
...

Installing the shim package
Calling: ['bash', '-c', 'zypper --no-cd --non-interactive --gpg-auto-import-keys --root /system-root in --auto-agree-with-licenses --allow-vendor-change --download in-advance --replacefiles --allow-downgrade shim &>> /system-root/var/log/distro_migration.log']
Refreshing service 'Basesystem_Module_x86_64'.
Refreshing service 'Containers_Module_x86_64'.
Refreshing service 'Desktop_Applications_Module_x86_64'.
Refreshing service 'Development_Tools_Module_x86_64'.
Refreshing service 'Legacy_Module_x86_64'.
Refreshing service 'Public_Cloud_Module_x86_64'.
Refreshing service 'Python_2_Module_x86_64'.
Refreshing service 'SUSE_Linux_Enterprise_Server_x86_64'.
Refreshing service 'Server_Applications_Module_x86_64'.
Refreshing service 'Web_and_Scripting_Module_x86_64'.
Loading repository data...
Calling: ['mkdir', '-p', '/var/tmp/kexec']
Running kernel load service
Loading the target kernel
Getting cmdline for vmlinuz
Calling: ['kexec', '--load', '/system-root/boot/vmlinuz', '--initrd', '/var/tmp/kexec/initrd', '--kexec-file-syscall', '--command-line', 'root=UUID=8af94f93-1497-4535-8a9d-29dd942e0564 console=ttyS0 multipath=off net.ifnames=0 nvme_core.io_timeout=4294967295 nvme_core.admin_timeout=4294967295 8250.nr_uarts=4 dis_ucode_ldr']
Reading installed packages...
'shim' is already installed.
No update candidate for 'shim-15.4-4.7.1.x86_64'. The highest available version is already installed.
Nothing to do.
Updating the shimbootloader
Calling: ['chroot', '/system-root', 'shim-install', '--removable']
Updating the bootloader
Calling: ['chroot', '/system-root', '/sbin/update-bootloader', '--reinit']
Config file content:
{}
Calling: ['systemctl', 'stop', 'suse-migration-console-log']
Umounting system
Calling: ['umount', '--lazy', '/system-root/boot/efi']
Umounting /system-root/boot/efi: command(output='', error='', returncode=0)
Calling: ['umount', '--lazy', '/system-root/']
Umounting /system-root/: command(output='', error='', returncode=0)
Calling: ['systemctl', 'kexec']
Reboot system: 
```

PR #325, opened 2025-04-07
Add product migration check

schaefi:
Add check which sends an API call to the registration server migrations API to ask for the possible upgrade path. This is acually the same call done by the zypper migration plugin but prior the actual migration starts. The request can only be issued if product and credentials data as well as the registration server URL could be found on the system to migrate. If the required information to issue the request cannot be found, the check is skipped. In case the regsitration server responds with an error, the pre-check prints this information such that the user knows about the issue prior starting the migration. The error information sent from the registration server could also include instructions how to fix the situation such that a migration can succeed. For example on LTSS registered systems the request will respond with an error message that products with activated LTSS cannot be migrated and how to call SUSEConnect to disable LTSS such that a migration will become possible.

### DIFF
--- a/suse_migration_services/prechecks/scc.py
+++ b/suse_migration_services/prechecks/scc.py
@@ -1,0 +1,179 @@
+# Copyright (c) 2025 SUSE LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>#
+from lxml import etree
+import platform
+import requests
+import yaml
+import logging
+import os
+import glob
+
+from suse_migration_services.command import Command
+from suse_migration_services.defaults import Defaults
+
+log = logging.getLogger(Defaults.get_migration_log_name())
+
+
+def migration(migration_system=False):
+    """
+    Send API call to ask the registration server for an upgrade path
+    and report any error this request might come back with
+    """
+    if migration_system:
+        # This check must not be called inside of the
+        # migration system live iso or container image.
+        # It will only be useful on the system to migrate
+        # at install time of the Migration package or on
+        # manual user invocation of suse-migration-pre-checks
+        # prior the actual migration.
+        return
+
+    installed_products = get_installed_products()
+    installed_products['target_base_product'] = get_migration_target()
+
+    if installed_products.get('installed_products') and \
+       installed_products.get('target_base_product'):
+        registration_server = get_registration_server_url()
+        registration_credentials = get_scc_credentials()
+        if registration_server and registration_credentials:
+            request_migration_offer(
+                installed_products,
+                registration_credentials,
+                registration_server
+            )
+            return
+    log.info('System not registered, skipped')
+
+
+def request_migration_offer(
+    installed_products, registration_credentials, registration_server
+):
+    data = {
+        'installed_products':
+            installed_products['installed_products'],
+        'target_base_product':
+            installed_products['target_base_product']
+    }
+    headers = {
+        'accept': 'application/json',
+        'Content-Type': 'application/json'
+    }
+    response = None
+    try:
+        uri = '{}/connect/systems/products/offline_migrations'.format(
+            registration_server
+        )
+        user = registration_credentials['username']
+        secret = registration_credentials['password']
+        response = requests.post(
+            uri, json=data, auth=(user, secret), headers=headers
+        )
+        offline_migrations = response.json()
+        result_error = offline_migrations.get('error')
+        if result_error:
+            log.error(result_error)
+    except Exception as issue:
+        log.error(
+            'Post request to {} failed with {}'.format(
+                registration_server,
+                response.text if response else issue
+            )
+        )
+
+
+def get_scc_credentials():
+    scc_credentials = '/etc/zypp/credentials.d/SCCcredentials'
+    result = {}
+    if os.path.isfile(scc_credentials):
+        try:
+            with open(scc_credentials, 'r') as config:
+                for line in config.read().splitlines():
+                    if line.startswith('username'):
+                        result['username'] = line.split('=')[1]
+                    if line.startswith('password'):
+                        result['password'] = line.split('=')[1]
+        except Exception as issue:
+            message = 'Loading {0} failed: {1}: {2}'.format(
+                scc_credentials, type(issue).__name__, issue
+            )
+            log.error(message)
+    return result
+
+
+def get_registration_server_url():
+    suseconnect_config = '/etc/SUSEConnect'
+    if os.path.isfile(suseconnect_config):
+        try:
+            with open(suseconnect_config, 'r') as config:
+                scc_data = yaml.safe_load(config) or {}
+                return scc_data.get('url')
+        except Exception as issue:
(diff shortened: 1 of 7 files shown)


suse_migration_services/prechecks/scc.py
```diff
@@ -0,0 +1,168 @@
...
+                        result['password'] = line.split('=')[1]
+        except Exception as issue:
+            message = 'Loading {0} failed: {1}: {2}'.format(
+                scc_credentials, type(issue).__name__, issue
+            )
+            log.error(message)
+    return result
+
+
+def get_scc_url():
```
rjschwei, line 106: It is not necessarily scc, I suggest `get_registration_server_url` as name for the function.

suse_migration_services/prechecks/scc.py
```diff
@@ -0,0 +1,168 @@
...
+                return scc_data.get('url')
+        except Exception as issue:
+            message = 'Loading {0} failed: {1}: {2}'.format(
+                suseconnect_config, type(issue).__name__, issue
+            )
+            log.error(message)
+
+
+def get_migration_target():
+    migration_config = '/etc/sle-migration-service.yml'
```
rjschwei, line 121: When only the `-pre-checks` package is installed `/etc/sle-migration-service.yml` does not exist unless the user planned ahead and created it. But hen the user does not know our defaults. Should we, as part of the `-pre-checks` package installation write out our defult configuration as `/etc/sle-migration-service.yml`?

suse_migration_services/prechecks/scc.py
schaefi: yes that's better, thanks and renamed

suse_migration_services/prechecks/scc.py
schaefi: I did apply our default in code here

*  https://github.com/SUSE/suse-migration-services/blob/ltss_pre_check/suse_migration_services/prechecks/scc.py#L142

my idea was if the target product cannot be found by a configuration file we check for the installed live migration package and depending on that check a default applies. Currently I check for `SLES15-Migration` and if found SLES/15.3/x86_64` applies. Whereas while writing this I believe the arch value should be taken from the host arch

For SLES16 the package would be named `SLES16-Migration` and another default can apply

For the container case no pre-checks service is enabled anyway because this is an online upgrade approach in a container and you stay on the system and you get errors from the container runtime

> as part of the -pre-checks package installation write out our defult configuration as ...

I was thinking about this part but I don't wanted to modify the system by the pre-checks code. During my absence people added a --fix option which modifies the system at several places. This smells and by default the checks are also not called with this option. In addition it's also not documented so customers don't know about --fix anyway and I guess it was never used in reality. I saw some changes which I would have declined because incomplete or undocumented or simply wrong... anyway I was not here, thus not eligible to complain. But the overall idea for pre-checks was to check and inform and not to modify the system.

Thoughts ?

schaefi: and maybe it would be a good idea if the live migration package e.g. `SLES15-Migration` provides a proper `/etc/sle-migration-service.yml`. In this case the target is finally an explicit setup and not defined by the OS version the live migration image was built for. But that would be subject to another PR and more work

schaefi: oh, and I was wrong. The --fix option is documented, sorry about that one

schaefi:
```
A `-f/--fix` option that will remediate the following issues:

1. Set multiversion.kernels to the correct value and remove all
old (not currently running) kernels.
```

I would have not allowed our code to do this. Remember the DMS is used to upgrade production systems. Old not running kernels are not necessarily subject to be removed. If they should be removed I want the customer to do this by himself and not some code from us where we then become responsible for. To be honest I would drop all of that fix code. Yet another topic...

jesusbv: > my idea was if the target product cannot be found by a configuration file we check for the installed live migration package and depending on that check a default applies. Currently I check for `SLES15-Migration` and if found `SLES/15.3/x86_64` applies. Whereas while writing this I believe the arch value should be taken from the host arch

Agree on taking the arch from the host, 
maybe a silly question but shouldn't the base product be from the host too ?

> 
> For SLES16 the package would be named `SLES16-Migration` and another default can apply
>
but the pre-check still works, right ? for the different defaults of `SLES16-Migration` values/ differences
 

> I was thinking about this part but I don't wanted to modify the system by the pre-checks code. During my absence people added a --fix option which modifies the system at several places. This smells and by default the checks are also not called with this option. In addition it's also not documented so customers don't know about --fix anyway and I guess it was never used in reality. I saw some changes which I would have declined because incomplete or undocumented or simply wrong... anyway I was not here, thus not eligible to complain. But the overall idea for pre-checks was to check and inform and not to modify the system.
> 
> Thoughts ?

can we fix the `--fix` option now then ?

rjschwei: The fix option is a separate topic and I would like to park that discussion. The only thing I will say at this point is that it was intended as a convenience for the user. But again we should discuss this in a different context.

Thanks for the link were the defaults are applied. Now I understand your thought process. Which leads me really to the next level of that question and you had hinted at this. Should we not make the default configuration explicit and put the file on the users system so they can see it?

But maybe that is also a discussion we should have for a separate PR. There is some not insignificant logic we would have to consider between pre-checks and the migration package.

schaefi: >The fix option is a separate topic and I would like to park that discussion. 

yes

>Should we not make the default configuration explicit and put the file on the users system so they can see it? But maybe that is also a discussion we should have for a separate PR

yes we should and yes we need to do this carefully. That's because the package that can provide the config is a package produced by a wrapper from an image build. So we need to adapt the wrapper and I don't know how to add additional source information to be consumed by the wrapper script. So this needs a bit of research first.

schaefi: >maybe a silly question but shouldn't the base product be from the host too ?

No the base product of the host is e.g SLE12 and you want to migrate to "what" ? You don't know that by looking at the host. You only know it if a user tells you (/etc/sle-migration-service.yml) or if you make assumptions according to the installed migration packages. If SLES15-Migration is installed the assumption is we migrate to SLES15 and that we migrate to SP3 of SLES15 is a default that we set due to a management decision. The real target product if not explicitly specified can only be identified by looking inside the live migration image and check its product. So you would need to loop mount the live iso, and look inside. I believe this is too much of an effort at this point and we also want to change this such that we explicitly provide the /etc/sle-migration-service.yml telling us the migration target when we install the SLE15-Migration package or the SLE16-Migration package... or whatever migration package comes along

schaefi: > but the pre-check still works, right ? for the different defaults of SLES16-Migration values/ differences

This is actually a good question. The current pre-checks do not differentiate between the distributions. The one I wrote here will be working for all products because it just targets the SCC API and ask for an upgrade path. This is a general useful pre-check. However there are other pre-checks e.g. for multiversion kernels. Does this concept still exist on SLE16, does this check still makes sense ? I can't tell that. So yes the pre-checks as of today also need to be reviewed from a design perspective such that it would be possible to differentiate them per distro. Yet another topic to discuss

schaefi: >can we fix the --fix option now then ?

Changing that would be a behavior change and as I don't know if there is anybody out there using it, I have no idea if we can afford to drop it. I would like to really fundamentally change this pre-checks thing. I mean look at the example output I provided. Apart from the message from SCC, do you think there is anything really useful available for a customer ? I find it just confusing to be honest. In my thinking the pre-checks should be quiet unless there is something to report. If it reports something it should be useful to a user giving an information what he can change to fix the issue. And by no means we are changing a production system. That's what I think

suse_migration_services/prechecks/scc.py
```diff
@@ -0,0 +1,169 @@
...
+                uri = '{}/connect/systems/products/offline_migrations'.format(
+                    registration_server
+                )
+                user = registration_credentials['username']
+                secret = registration_credentials['password']
+                response = requests.post(
+                    uri, json=data, auth=(user, secret), headers=headers
+                )
+                offline_migrations = response.json()
+                result_error = offline_migrations.get('error')
```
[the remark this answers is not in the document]
schaefi: done, how about it now ?

jesusbv (approved):

PR #330, opened 2025-05-29
Add high availablity extension check

### LINKED ISSUE #329: [Feature Request] Call HA Migration Prechecks
High availability module has breaking changes in SLE 16 and we have implemented [pre-migration checks in crmsh](https://github.com/ClusterLabs/crmsh/pull/1629). We'd like to integrate this to system-wide migration prechecks. Here is the background and our proposal. Please review and if we reach a consensus I will send a pull request.

# Background

The precheck in crmsh is provided in command `crm cluster health sles16`. This command is able to check cluster wide configuations and requires network connections between cluster nodes. To fully utilize this command's capabilities, we advise users to manually run this check and fix the detected problems before initiating a system-wide migration.

We also provide a limited local version of these checks with `crm cluster health sles16 --local` and it runs without depending on other cluster node. It is suitable to be integrated to the system-wide prechecks.

# Proposal

We'd like to add the following code:

* Add a module `suse_migration_services.prechecks.ha`.

* In this module, check if both file `/usr/sbin/crm` and `/etc/corosync/corosync.conf` exist. If any one of them does not exist, this host is not a ha cluster node and we will return silently.

* Check if `crm` is new enough to have the pre-migration check ability. If not, return silently. We will check this by calling `crm help cluster health` and check if there is a `sles16` option in the usage.

* Otherwise we call `crm cluster health sles16 --local`. This command writes its check process and result to stdout.

# Unresolved Problems

We are not able to block the migration when problems needing manually fixing are detected.

---

CC: @zzhou1 @liangxin1300

nicholasyang2022:
Add a check to call ha checks provided by crmsh. If ha cluster config files exist and crmsh is installed and up to date, this module calls crmsh as a subprocess to run the checks. The result is printed to stdout directly by crmsh.

close #329

### DIFF
--- a/suse_migration_services/prechecks/ha.py
+++ b/suse_migration_services/prechecks/ha.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2025 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+"""Call prechecks for HA provided by crmsh"""
+import logging
+import os
+import subprocess
+
+from suse_migration_services.defaults import Defaults
+
+
+log = logging.getLogger(Defaults.get_migration_log_name())
+
+
+def check_ha(migration_system=False):
+    if migration_system:
+        log.info('Skipped checks for high availablity extension in migration system.')
+        return
+    if not _corosync_conf_exists():
+        log.info('corosync.conf not found. Skipped checks for high availablity extension.')
+        return
+    if not _is_crmsh_capable_to_check():
+        log.warning('crmsh version is too old. Skipped checks for high availablity extension.')
+        return
+    subprocess.run(['crm', 'cluster', 'health', 'sles16', '--local'])
+
+
+def _corosync_conf_exists():
+    return os.access('/etc/corosync/corosync.conf', os.F_OK)
+
+
+def _is_crmsh_capable_to_check():
+    try:
+        result = subprocess.run(
+            ['crm', 'help', 'cluster', 'health'],
+            stdout=subprocess.PIPE, stderr=subprocess.DEVNULL,
+        )
+    except FileNotFoundError:
+        return False
+    except Exception as e:
+        log.error('Failed to call crmsh: %s', e)
+        return False
+    return result.returncode == 0 and 'sles16' in result.stdout.decode(errors='replace')
--- a/suse_migration_services/prechecks/pre_checks.py
+++ b/suse_migration_services/prechecks/pre_checks.py
@@ -24,6 +24,7 @@ from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import Logger
 from suse_migration_services.migration_config import MigrationConfig
 
+import suse_migration_services.prechecks.ha as check_ha
 import suse_migration_services.prechecks.repos as check_repos
 import suse_migration_services.prechecks.fs as check_fs
 import suse_migration_services.prechecks.kernels as check_multi_kernels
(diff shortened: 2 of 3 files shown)


suse_migration_services/prechecks/ha.py
```diff
@@ -0,0 +1,57 @@
...
+import os
+import subprocess
+
+from suse_migration_services.defaults import Defaults
+
+
+log = logging.getLogger(Defaults.get_migration_log_name())
+
+
+def check_ha(migration_system=False):
```
zzhou1, line 29:
```suggestion
def cluster_migration(fix=False, migration_system=False):
```

suse_migration_services/prechecks/ha.py
```diff
@@ -0,0 +1,57 @@
...
+    if migration_system:
+        log.info('Skipped checks for high availablity extension in migration system.')
+        return
+    if not _corosync_conf_exists():
```
zzhou1, line 33:
```suggestion
    if not _corosync_conf_exists(migration_system):
```

suse_migration_services/prechecks/ha.py
```diff
@@ -0,0 +1,57 @@
...
+        log.info('corosync.conf not found. Skipped checks for high availablity extension.')
+        return
+    if not _is_crmsh_capable_to_check():
+        log.warning('crmsh version is too old. Skipped checks for high availablity extension.')
+        return
+    subprocess.run(['crm', 'cluster', 'health', 'sles16', '--local'])
+
+
+def _corosync_conf_exists():
+    return os.access('/etc/corosync/corosync.conf', os.F_OK)
```
zzhou1, line 43:
```suggestion
def _corosync_conf_exists(migration_system=False):
    corosync_conf_path='/etc/corosync/corosync.conf'
    if migration_system:
            corosync_conf_path = Defaults.get_system_root_path() + corosync_conf_path
    return os.access(corosync_conf_path, os.F_OK)
```

suse_migration_services/prechecks/pre_checks.py
```diff
@@ -113,3 +114,7 @@ def main():
         migration_system=migration_system_mode
     )
     log.info('Done')
+
+    log.info('--> Checking high availability extension...')
+    check_ha.check_ha()
```
zzhou1, line 119:
```suggestion
    check_ha.cluster_migration(
        fix=args.fix, 
        migration_system=migration_system_mode
    )
```

suse_migration_services/prechecks/ha.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def _is_crmsh_capable_to_check():
```
zzhou1, line 46:
```suggestion
def _is_crmsh_capable_to_check(migration_system=False):
```

Sounds that we have challenge to run crmsh in the migration system mode!

suse_migration_services/prechecks/ha.py
```diff
@@ -0,0 +1,57 @@
...
+
+from suse_migration_services.defaults import Defaults
+
+
+log = logging.getLogger(Defaults.get_migration_log_name())
+
+
+def check_ha(migration_system=False):
+    if migration_system:
+        log.info('Skipped checks for high availablity extension in migration system.')
```
zzhou1, line 31:
```suggestion
        log.info('Skipped checks for high availablity extension in the migration system mode.')
```

zzhou1 (commented):
The code is good in general for the stage in 15SP6/SP7.

However, I realize that the major observation and gap is `crm cluster health sles16 --fix` can not run in the migration system mode currently. That means, currently the cluster migration can not proceed automatically.

We need keep brainstorming to find the way out if possible ...

schaefi:
You could consider to only check and in case a fixing is required that cannot be applied by the DMS itself, then just write an information to the user what needs to be done.

I do the same for the case when no upgrade path can be calculated through zypper migrate

In general I would have objected a pre-check system to "fix" == "modify" the system because I think this is an action the DMS should not perform. This "--fix" concept was added while I wasn't around, so it exists now, but I do not recommend any subsequent checks to add more code that modifies the customer production system prior a migration.

schaefi:
I also see that the code is not covered by tests. Usually a pull request triggers calling the tests but due to the SUSE org settings and because the PR comes from a fork github does not run them. @nicholasyang2022 I'm happy to add you to the project with write permissions such that you can create PRs from branches which then also triggers all our github actions. Please let me know if this would make sense ?

Other than that in your fork please make sure that the following passes without errors

```
make check
make test
```

Thanks

nicholasyang2022:
No worry. The actions is simple and I just enabled it in my fork. You can see the results there. And I will add test cases later.

schaefi:
@nicholasyang2022 I see everything is passing in your fork. Can you confirm this one is ready for a final review ? Thanks much

nicholasyang2022:
Yes, the precheck is ready for review.

And I'd like to add a unit to run the actual migration for HA in the migration system. Do you want it to be add together in this pull request, or in another one?

PR #333, opened 2025-06-25
Product migration

schaefi:
This is an idea to prevent upgrading into unsupported service packs. As discussed SLES for SAP wants to upgrade into SP3 which is a supported target for the SAP product. However, for standard SLES we are at SP7 and (SP6 within the 6months transition period) anything older is not supported and SUSE does not offer package updates for older SPs except with LTSS. The hop concept from an unsupported service pack e.g SP3 into a supported one (SP6/SP7) has proven to not work properly and is also never tested by QA.

Therefore my idea:

- [x] Provide product specific migration environments, beginning with SLES-SAP
- [ ] Update documentation and add a chapter for SAP customers
- [x] bump the standard SLES case to SP7 (latest SP)
- [ ] add a migration pre-check to identify the product of the host to upgrade and match if the correct SLE[X]-[XX]-Migration.rpm package was installed and if not tell the user which one to install
- [x] continue providing product specific  migration environments when needed

### DIFF
--- a/image/pubcloud/product/sles_sap/config.kiwi
+++ b/image/pubcloud/product/sles_sap/config.kiwi
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<image schemaversion="6.8" name="SLES15-SAP-Migration">
+<image schemaversion="6.8" name="SLES15-SAP_Migration">
     <description type="system">
         <author>Public Cloud Team</author>
         <contact>public-cloud-dev@susecloud.net</contact>
--- a/image/pubcloud/product/sles_sap/config.kiwi
+++ b/image/pubcloud/product/sles_sap/config.kiwi
@@ -1,0 +1,107 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="6.8" name="SLES15-SAP-Migration">
+    <description type="system">
+        <author>Public Cloud Team</author>
+        <contact>public-cloud-dev@susecloud.net</contact>
+        <specification>
+            SLES15 Distribution Migration Live System to upgrade
+            SLES12-SP4+ SAP instances to SLES15 SP3
+        </specification>
+    </description>
+    <preferences>
+        <type image="iso" primary="true" flags="overlay" firmware="efi" boottimeout="1"/>
+        <version>2.1.11</version>
+        <packagemanager>zypper</packagemanager>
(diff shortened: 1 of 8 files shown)


image/pubcloud/product/sles_sap/config.kiwi
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="6.8" name="SLES15-SAP-Migration">
+    <description type="system">
+        <author>Public Cloud Team</author>
+        <contact>public-cloud-dev@susecloud.net</contact>
+        <specification>
+            SLES15 Distribution Migration Live System to upgrade
+            SLES12-SP4+ SAP instances to SLES15 SP3
```
rjschwei, line 9: We should move to 15 SP4 as we are withing 6 months of SP3 reaching EOL and it will take us probably a couple of months to get this change through the system.

image/pubcloud/product/sles_sap/config.kiwi
```diff
@@ -0,0 +1,107 @@
...
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+    </preferences>
+    <users>
+        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
+        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/home/migration" name="migration" groups="users"/>
+    </users>
+    <repository type="rpm-md" >
+        <source path='obsrepositories:/'/>
+    </repository>
```
rjschwei, line 28: 15 SP4 repos

rjschwei (changes requested):

schaefi:
For reference I have built the SAP migration live system here:

* https://build.opensuse.org/package/show/home:marcus.schaefer:dms:live/SLES12-SP4-SLES15-SAP-Migration

The package name is

* SLES15-SAP_Migration-2.1.11-3.x86_64.rpm

for standard SLES based migrations the package name will stay unchanged ```SLES15-Migration``` We need to use this package information to check against the installed product such that we can tell the customer the correct migration package in case there is a mismatch. We also need to update the documentation and add a section for SAP customers

rjschwei (approved):

PR #334, opened 2025-06-26
RPM wrapper fixes

fcrozat:
Adapt the rpm wrapper to work properly with SLE16 reproducible build

### DIFF
--- a/image/package/suse-migration-rpm.changes
+++ b/image/package/suse-migration-rpm.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jul  2 12:55:06 UTC 2025 - Frederic Crozat <fcrozat@suse.com>
+
+- Add build and perl(Date::Parse) to ensure we can append changelog
+  entries to generated package.
+
+-------------------------------------------------------------------
 Wed Oct 30 13:21:08 UTC 2024 - Marcus Schäfer <marcus.schaefer@gmail.com>
 
 - Install migration live image to /migration-image
--- a/image/package/suse-migration-rpm.spec
+++ b/image/package/suse-migration-rpm.spec
@@ -26,6 +26,8 @@ Url:            https://github.com/SUSE/suse-migration-services
 Source:         %{name}.tar.gz
 BuildRequires:  filesystem
 Requires:       suse-migration-services
+Requires:       build
+Requires:       perl(Date::Parse)
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 
 %description


schaefi:
@fcrozat funny I was about to start fixing the wrapper too because yesterday I found out that it only worked for the SLE15 case. Thanks much for digging into it and fixing it

schaefi:
without a closer look yet, have you tested that this change still works for the SLE15 builds ?

schaefi:
I will submit your changes to /home/ms/sbs/external/home:marcus.schaefer:dms/suse-migration-rpm to check if everything builds

schaefi:
I get the following error

```
[  289s] cp: cannot stat '/.build-changelog': No such file or directory
[  289s] /usr/lib/build/kiwi_post_run script failed!

```

but the build package is pulled in

```
osc buildinfo SLE_15_SP5 x86_64 | grep '\"build\"'
Using local file: config.kiwi
  <bdep name="build" notmeta="1" version="20250602" release="160000.1.1" arch="noarch" hdrmd5="d84a0b891097e76509b195c642a06ed3" project="SUSE:SLFO:Main:Build" repository="standard"/>
```

So something is broken, can you take a look please ?

It's all active and live at

https://build.opensuse.org/project/show/home:marcus.schaefer:dms:live

fcrozat:
> For me this breaks all builds

obs-build is probably too old there, I’ll adapt the script

schaefi:
> obs-build is probably too old there, I’ll adapt the script

It's taken from `project="SUSE:SLFO:Main:Build"` shouldn't this be the latest and greatest ?

schaefi:
@fcrozat Thanks for fixing, the SLE15 builds are now working again but the SLE16 build failed with the same changelog not found error. This means the version of the build script in SLFO:Main seems to be not suitable for the feature you are using ? I did a copypac of the build package from TW into the :dms devel project and currently rebuild to see if that would fix it.

If it does we have to create a versioned and distro selected Requires to the spec file of the suse-migration-rpm package. Something like

```
%if 0%{?suse_version} && 0%{?suse_version} >= 1600
Requires: build >= 20250602
%endif
```

or whatever fits the target

schaefi:
with the latest build script I get

```
[  254s] error: %source_date_epoch_from_changelog is set, but %changelog has no entries to take a date from
[  254s] 

```

fcrozat:
> with the latest build script I get
> 
> ```
> [  254s] error: %source_date_epoch_from_changelog is set, but %changelog has no entries to take a date from
> [  254s] 
> ```

you need to create at a .changes file in for SLES15-SP5-SLES16-Migration with one changelog entry

schaefi:
done ... building again. Brings me to the questions

1. where do we want to maintain the changes file data for the live image build ?
2. what type of information should be added to the changes file ? If you want to track real changes of the image description, this changes information could be generated in the same way as we build all the other changes data from the git history for the other DMS components. If you like to track changes of the image content in this changes file it will become a manual task

schaefi:
and the next problem with this build weirdness

```
[  257s] Can't locate Date/Parse.pm in @INC (you may need to install the Date::Parse module) (@INC entries checked: /usr/lib/build /usr/lib/perl5/site_perl/5.40.2/x86_64-linux-thread-multi /usr/lib/perl5/site_perl/5.40.2 /usr/lib/perl5/vendor_perl/5.40.2/x86_64-linux-thread-multi /usr/lib/perl5/vendor_perl/5.40.2 /usr/lib/perl5/5.40.2/x86_64-linux-thread-multi /usr/lib/perl5/5.40.2 /usr/lib/perl5/site_perl) at /usr/lib/build/changelog2spec line 31.
[  257s] BEGIN failed--compilation aborted at /usr/lib/build/changelog2spec line 31.
```

schaefi:
trying to add

Requires: perl-DateTime

to the suse-migration-rpm.spec

fcrozat:
> trying to add
> 
> Requires: perl-DateTime-Format-DateParse
> 
> to the suse-migration-rpm.spec

perl-DateTime should be enough. perl-DateTime-Format-DateParse is really old wrapper around Date::Parse

schaefi:
> perl-DateTime should be enough. perl-DateTime-Format-DateParse is really old wrapper around Date::Parse

trying with `Requires:       perl(Date::Parse)` interesting enough the build spec has this as a recommended package but fails when it is not installed

schaefi:
> > done ... building again. Brings me to the questions
> > ```
> > 1. where do we want to maintain the changes file data for the live image build ?
> > ```
> 
> I'm tempted to say on git, so we only edit things in one location
> 
> > ```
> > 2. what type of information should be added to the changes file ? If you want to track real changes of the image description, this changes information could be generated in the same way as we build all the other changes data from the git history for the other DMS components. If you like to track changes of the image content in this changes file it will become a manual task
> > ```
> 
> We might need a bit of both, simply because the image description might not change a lot or at all except versioning but its content might be a lot different. Difficult to give a perfect answer.

I agree. So let's add an initial .changes information to the git. I vote for adding this file in this PR as it belongs together.

fcrozat:
> I agree. So let's add an initial .changes information to the git. I vote for adding this file in this PR as it belongs together.

I'll add those

fcrozat:
Fixed and rebased

schaefi (approved):
Looks good :+1:

PR #336, opened 2025-07-04
cleanup specfile to follow python singlespec policy

fcrozat:


package/suse-migration-services-spec-template
```diff
@@ -57,13 +56,19 @@ Requires:         suse-migration-services = %{version}-%{release}
 Requires(preun):  systemd
 Requires(postun): systemd
 BuildArch:        noarch
+%define python_subpackage_only 1
+%python_subpackages
```
rjschwei, line 60: But this creates a "python$VERSION-" binary package, even if we build only 1 subpackage, at least as far as I know. We do not want the binary packages with the version prefix, i.e.  `python311-`, `python313-` ... because then we have to chase the ever changing name.

fcrozat: @dirkmueller said it should be done this way

schaefi: correct we don't want that. I don't see an issue with the packaging and this PR has no description nor does it provide any reason why this change has any benefit. It does not work that way sorry

rjschwei: So we should add the necessary `Obsoletes:` directive. Not switch to the subpackages build that has undesired side effects.

dirkmueller: >We do not want the binary packages with the version prefix, i.e. python311-, python313- ... because then we have to chase the ever changing name.

the name is not "ever changing", it is reflecting where the files are being installed and makes coinstallation possible. also all the macro logic is in place for obsoleting the old versions so everything would be just working if done properly.

fcrozat:
> This change needs clarification and the spec must build for both SLE15 and whatever comes after it.

anything missing ?

package/suse-migration-services-spec-template
rjschwei: > > We do not want the binary packages with the version prefix, i.e. python311-, python313- ... because then we have to chase the ever changing name.
> 
> the name is not "ever changing", it is reflecting where the files are being installed and makes coinstallation possible. also all the macro logic is in place for obsoleting the old versions so everything would be just working if done properly.

I am sorry maybe I am not totally familiar with the `subpackages` macro. But to my understanding when we use the `subpackages` macro the binary rpm that gets produced has the Python version encode as given in the initial comment. How is this not "ever changing" whenever the Python interpreter changes?

In the current implementation of the spec file the name is always "python-" using subpackages it is "python311-" or "python313-" or "pythonXYZ-" depending on the interpreter. That means we create a condition where the name changes of the binary have to be chased in other places. And the name is "ever changing", outside of our control, because it is really hard to argue that "python311-" is the same as "python313-".

rjschwei:
> > This change needs clarification and the spec must build for both SLE15 and whatever comes after it.
> 
> anything missing ?

Drop the use of the `subpackages` macro. Unless the use of `python_subpackage_only 1` produces `python-` as the binary instead of `python$INTERPRETERVERSION-`

package/suse-migration-services-spec-template
dirkmueller: > But to my understanding when we use the subpackages macro the binary rpm that gets produced has the Python version encode as given in the initial comment. How is this not "ever changing" whenever the Python interpreter changes?

Hows that different to what the spec file is doing today?

`%package          -n %{pythons}-migration`

the way it is done right now means that it is just completely broken when pythons is not resolving to a single interpreter. but %primary_python is moving. 


> In the current implementation of the spec file the name is always "python-" using subpackages it is "python311-" or "python313-" or "pythonXYZ-" depending on the interpreter.

no it's not. please look at the spec file. 

> That means we create a condition where the name changes of the binary have to be chased in other places. And the name is "ever changing", outside of our control, because it is really hard to argue that "python311-" is the same as "python313-".

that the name is changing is intentional, because the *content* will also be changing. it is installed into different paths and hence the content of python311-foo will not be found by python3.11 interpreter. 

Really, we have 12600 python packages packaged that way, I don't see why this one package would make any material issue here? the dependency here is internal, package lists are generated automatically based on transitive dependency resolution, and the macros will automatically do the correct provides/obsoletes/weakobsoletes. Just like for the other 12600 python modules.

package/suse-migration-services-spec-template
```diff
@@ -58,23 +57,25 @@ Requires:         suse-migration-services = %{version}-%{release}
...
```
dirkmueller, line 61: this doesn't do anything when the main package name does not start with `python-` which is the case here. remove again please.

fcrozat: grrr, I thought it was already removed

package/suse-migration-services-spec-template
```diff
@@ -58,23 +57,25 @@ Requires:         suse-migration-services = %{version}-%{release}
...
+%description -n python3-migration
 Python based implementation for suse major code stream upgrade system
 
 %package          -n suse-migration-pre-checks
 Summary:          The pre-checks code used with the python migration module
 Group:            System/Management
 Requires:         %{pythons}-lxml
 Requires:         %{pythons}-requests
-Requires:         %{pythons}-migration
+Requires:         python3-migration = %{version}
```
dirkmueller, line 78:
```suggestion
Requires:         %{pythons}-migration >= %{version}-%{release}
```

potentially (e.g. the release part might be important)

schaefi: > and remove the python3-* package alltogether.

that sounds very weird to me. Of course it's reusable. python3-migration provides the `suse_migration_services` python namespace for distribution migration implementations of some kind. At the moment the module is used

* for the `suse-migration-services` package which provides systemd services invoking python entry-points from the  `suse_migration_services` python namespace 
* for the `suse-migration-pre-checks` package which  provides the `suse-migration-pre-checks` tool and service

The first runs in the context of the migration system and the later runs in the context of the system to become migrated. A python module package reusable for the different targets and code endpoints is a must have for this project and should not be changed.

I'm still very confused what actual problem are we trying to solve here ? It's surely not needed to provide the module package for different python versions. If we stick with `python3-migration` everywhere that should be fine, and that's what it builds with the proposed changes

package/suse-migration-services-spec-template
```diff
@@ -58,23 +57,25 @@ Requires:         suse-migration-services = %{version}-%{release}
...
 Requires(postun): systemd
 BuildArch:        noarch
+%define python_subpackage_only 1
+%python_subpackages
 
 %description
 Systemd services to prepare and run a distribution migration process.
 
-%package          -n %{pythons}-migration
+%package          -n python3-migration
```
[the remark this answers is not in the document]
fcrozat: renamed to -suse-migration-services

package/suse-migration-services-spec-template
```diff
@@ -61,19 +60,20 @@ BuildArch:        noarch
 %description
 Systemd services to prepare and run a distribution migration process.
 
-%package          -n %{pythons}-migration
+%package          -n python3-suse-migration-services
```
schaefi, line 63: This change looks so easy right ... and opens a can of worms and I can already smell the nightmare it will produce the next time I try to submit the DMS to SLE15. I see the obsoletes was set but no provides as documented here:

* https://en.opensuse.org/openSUSE:Upgrade_dependencies_explanation#Renaming_a_package

And again the question; What issue is this rename solving ? Yes the package is named python3-migration and oh well yes the python namespace is `suse_migration_services` and not just `migration` and by the way the package uses dashes and not underscores 

Really we are creating problems and trying to fix something that really nobody will care for. Whom does this help ?

schaefi: I really object here. The package names as they exist in SLE15 doesn't have to be changed for SLE16 unless this creates any sort of real conflict. We have to maintain this not only for SLFO and a package rename is an expensive change. I'm not even sure if we need an ECO for SLE15 in order to apply this change.

schaefi (changes requested):
unnecessary and incomplete package rename

fcrozat:
can we just rename the python subpackage to suse-migration-services-common starting with SLFO and be done with this ?

package/suse-migration-services-spec-template
fcrozat: done

package/suse-migration-services-spec-template
fcrozat: rename done

dirkmueller (approved):

PR #339, opened 2025-07-22
Fix suse-migration-console-log service

schaefi:
The console logging was based on forcing a dialog program on the main console. If the main console also runs a getty which is by default the case we run into several signal conflicts. This battle is always won by the getty because it restarts always whereas the console log service only restarts on failure. This leads to very weird conditions such as no log displayed on the console at all up to only a fraction of the log shown leading to wrong assumptions on the current state of the migration. This commit now changes the following:

1. The console log is forced into its own tty (8) and the service actively switches to this terminal

2. If the dialog ends (e.g user hits Exit) the service switches back to the first virtual terminal and restarts after 10 seconds. This is enough time to stop the log service in case it's unwanted, e.g. for debugging on this console

3. The dialog logging does no longer run on serial terminals. dialog does not really support to work well on serial consoles. The output is often completely broken and input sequences requires its own terminal. If the migration system is accessed from a terminal that doesn't show progress information a message in etc/motd has been added which provides information how to see progress

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,15 @@ SLES16-Migration: clean check test
 	cp image/generic/sle16/config.sh dist/config.sh
 	cp image/generic/sle16/SLES16-Migration.changes dist/SLES16-Migration.changes
 
+SLES15-Migration: clean check test
+	mkdir -p dist
+	tar --sort=name --owner=0 --group=0 --numeric-owner \
+		--transform 's,^\./,,' \
+		-czf dist/root.tar.gz \
+		-C image/pubcloud/sle15/root .
+	cp image/pubcloud/sle15/config.kiwi dist/config.kiwi
+	cp image/pubcloud/sle15/config.sh dist/config.sh
+
 setup:
 	poetry install --all-extras
 
--- a/image/generic/sle16/SLES16-Migration.changes
+++ b/image/generic/sle16/SLES16-Migration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 22 13:34:52 UTC 2025 - Marcus Schäfer <marcus.schaefer@suse.com>
+
+- Add etc/motd overlay file
+
+  Print message how to show migration progress information
+
+-------------------------------------------------------------------
 Fri Jun 27 11:48:55 UTC 2025 - Frederic Crozat <fcrozat@suse.com>
 
 - Fix build on more architectures
--- a/image/generic/sle16/root/etc/motd
+++ b/image/generic/sle16/root/etc/motd
@@ -1,0 +1,7 @@
+Migration System:
+
+If no progress is shown on this terminal watch via
+
+    tail -f /system-root/var/log/distro_migration.log
+
+Have a lot of fun :-)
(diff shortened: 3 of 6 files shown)


rjschwei (approved):

fcrozat:
Shouldn't we make sure systemd won't steal tty8 ? or maybe we dedicate the dialog to tty2 (more discoverable than 8) and add to motd the dialog is available on tty2 ?

schaefi:
> Shouldn't we make sure systemd won't steal tty8 ? or maybe we dedicate the dialog to tty2 (more discoverable than 8) and add to motd the dialog is available on tty2 ?

tty1-6 are reserved for consoles by our distribution. tty7 is reserved for graphics. All following ttys are not restricted to a use case. I don't see an issue with occupying that one. Also please remember this is the Migration live system, fully under our control and only for the purpose of running a migration process. We know exactly what tty is used for what as there is nothing else running other than a migration process.

>more discoverable than 8

As I wrote the service automatically switches to the tty if the condition is met. There is no reason that you need to discover anything.

>and add to motd the dialog is available on tty2 

If you are on a terminal and you don't see progress than the vt switch either did not work or this console cannot display it. Therefore the motd message tells you that you can run `tail -f ...` and watch something on the current console, whatever it might be. The message you want to display here would be pointless

schaefi:
For the record I have done the following tests with these changes

1. Public Cloud Workflow
    The progress dialog was displayed broken, lines were incomplete and cut off. As I said `dialog` does not cope well with serial terminals. The instance screenshot was empty because the progress was targeted to the serial console. With the patch in place, the login on the serial console displayed information how to watch progress and the instance screenshot showed a correct dialog window at the time of the screenshot.

2. QA/qemu Workflow
     Like with public clouds, broken data on the serial terminal when exposed via --serial stdout. With the patch in place the progress is visible on the qemu console. On the serial terminal information how to watch progress is displayed when you login

3. Real hardware (Raspberry PI) Workflow
     On arm I used `minicom` to attach to the ttyUSB0 serial console. Output was completely broken also affecting the serial terminal itself. Only a tty reset brought the connection back. With the patch in place progress is displayed if a display is connected to the device e.g HDMI display. On the serial console information how to watch progress is displayed when you login


Overall I find the behavior better than what we had before

schaefi:
>Since Migration is under our complete control, we can reuse tty as much as we want, we don't need to stick to the "installed "system convention.

right and there is no reason to continue sticking to it. Really what's your point in tty2, 8, 10, 42 ... it doesn't matter

>If the end-user switches to another tty and want to come back to the dialog.

The unit file has restart set to always after 10sec. You will automatically be back at this progress dialog in any case. The 10second delay exists to stop/interact the service in case you don't want to be back at the progress dialog and you are on an interactive terminal. If you remote login to the live system you will not land on the progress dialog because it runs on a tty different from 1-6

Makefile
fcrozat, line 1: This should be a different PR (or it is a git merge artefact ?)

fcrozat (changes requested):
please drop the Makefile change

fcrozat (approved):

PR #340, opened 2025-07-25
Install patterns-base-selinux for Apparmor migration

YoukouTenhouin:
Install patterns-base-selinux for Apparmor migration. This is necessary to make SELinux enabled after migration.

Fix bsc#1246219.

### DIFF
--- a/suse_migration_services/units/apparmor_migration.py
+++ b/suse_migration_services/units/apparmor_migration.py
@@ -22,10 +22,28 @@ import re
 # project
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import Logger
+from suse_migration_services.zypper import Zypper
 
 from suse_migration_services.exceptions import (
     DistMigrationAppArmorMigrationException
 )
+
+
+def install_patterns_base_selinux(root_path):
+    """Install patterns-base-selinux"""
+    Zypper.run([
+        '--no-cd',
+        '--non-interactive',
+        '--gpg-auto-import-keys',
+        '--root', root_path,
+        'install',
+        '--auto-agree-with-licenses',
+        '--allow-vendor-change',
+        '--download', 'in-advance',
+        '--replacefiles',
+        '--allow-downgrade',
+        'patterns-base-selinux'
+    ])
 
 
 def main():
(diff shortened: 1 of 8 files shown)


YoukouTenhouin:
Thanks for the idea! A generic service might worth considering, but for the purpose of this PR, I think since the installation of `patterns-base-selinux` is a part of the migration from Apparmor to SELinux, it makes more sense to add it to the existing Apparmor migration services (as how it is done right now with `update-bootloader` service). I will keep that idea in mind though.
What do you think?

schaefi:
That's fine for me.

> I will keep that idea in mind though.

yes please. I believe more and more services can make it more confusing and if we see that in the transition from SLE15 to SLE16 more packages/patterns needs to be replaced/added and this cannot be handled by the package manager, it would be good to have one configurable service to manage this part

With regards to the security system changing from apparmor to selinux, more changes than just the package replacement where needed e.g the update of the grub setup. So it's fine to have a dedicated service for this part

Thanks

suse_migration_services/units/apparmor_migration.py
```diff
@@ -22,12 +22,31 @@
...
 from suse_migration_services.exceptions import (
     DistMigrationAppArmorMigrationException
 )
 
 
+def install_patterns_base_selinux(root_path):
+    """Install patterns-base-selinux"""
+    Zypper.run([
+        '--no-cd',
+        '--no-cd',
```
schaefi, line 36: duplicate option `--no-cd` ?

suse_migration_services/units/apparmor_migration.py
```diff
@@ -22,12 +22,31 @@
...
+        '--non-interactive',
+        '--gpg-auto-import-keys',
+        '--root', root_path,
+        'in',
```
schaefi, line 40: Can we keep it explicit for better readability: `install` instead of `in` ?

schaefi (changes requested):
I like the new Zypper class :+1: 

the rest is just nitpick, thanks

suse_migration_services/units/apparmor_migration.py
YoukouTenhouin: Fixed, thanks for pointing it out!

suse_migration_services/units/apparmor_migration.py
YoukouTenhouin: Fixed. I also update the command in `update_bootloader` as well; it's a bit out of scope for this PR, but since we are already refactoring them to use the Zypper class I think it's a good chance to update it together.

schaefi (approved):
Looks good, thanks for this effort :+1:

PR #343, opened 2025-07-30
Fix requires on subpackage

dirkmueller:
I'm GUESSING what this is supposed to be. the line either needs to be removed, or moved to the suse-migration subpackage (to require -services) or it needs to be changed like this. no idea what is correct.


schaefi:
suse-migration-services requires itself, that is a mistake. This line should be removed

The package suse-migration is completely optional. It provides the migration tool to run an online upgrade container based

@dirkmueller can you delete that `Requires:         suse-migration-services = %{version}-%{release}` line and make the PR not a draft ? I think you found a real bug in the spec

Thanks for catching it

Issue #341, opened 2025-07-29
Migration sle15sp7 -> sle16 only succeeds after running twice


greygoo:
On a sle15sp7 system:
```
zypper ar -p 90 'http://download.suse.de/ibs/home:/fcrozat:/SLES16/SLE_$releasever' home_fcrozat_SLES16
zypper ar -p 90 http://download.suse.de/ibs/home:/fcrozat:/SLES16/images/ home:fcrozat:SLES16_images
zypper in suse-migration-sle16-activation
reboot
```
=> reboots into migration system
=> starts migration (ncurses output)
=> runs about 10 sec
=> reboots back to SLE15SP7
Log is very short, but does not contain any obvious errors:
[distro_migration-1st.log](https://github.com/user-attachments/files/21491863/distro_migration-1st.log)

However, when running again
```
zypper in suse-migration-sle16-activation
reboot
```
=> reboots into migration system
=> starts migration (ncurses output)
=> runs about 10 sec
=> jumps to login, no more nvurses output. cpu indicates guest is busy
=> runs abut 10min
=> reboots into SLE16
Log contains full update log and indicates not errors: [distro_migration-2nd.log](https://github.com/user-attachments/files/21491885/distro_migration-2nd.log)

Attached files:
[distro_migration-1st.log](https://github.com/user-attachments/files/21491863/distro_migration-1st.log)
[distro_migration-2nd.log](https://github.com/user-attachments/files/21491885/distro_migration-2nd.log)

schaefi:
One problem I found is that the pre-checks service which does some fixes on your system runs late.

However, that does not explain that other services does not run at all.

@greygoo can you please repeat the first step with the following setting in `/etc/sle-migration-service.yml`

```
debug: true
```

This will keep you in the migration live system and does not reboot. If you have console access you should be able to
login as either root or migration. The ssh keys from your system to upgrade should have been moved into the live migration system.

Once you are logged in, can you please check the status of the systemd services that has `suse-migration*` in its name. For example

```
systemctl status suse-migration-mount-system.service
systemctl status suse-migration-prepare.service
```

From your first log only the services:

```
cat distro_migration-1st.log  | grep "Running "
Running ssh keys service
Running grub setup service
Running suse-migration-pre-checks with options: fix: True
reboot
```

which is weird. Thus I expect some of the other services to have an issue that was not reported in the log file

Thanks

greygoo:
Running migration with debugging enabled, expect results asap

[...]

PR #348, opened 2025-08-06
Split /etc/resolv.conf setup into its own service

YoukouTenhouin:
The creation of /etc/resolv.conf should wait for network-online.target, but setup_host_network is supposed to be run before network is setup. Thus we split the setup of name resolvers into its own service to fix this issue.

Fix bsc#1246513


schaefi:
Makes a lot sense and is probably also related to #341

PR #350, opened 2025-08-06
Fix migration user home dir setup for SLE16

schaefi:
Added .ssh to migration user for SLE16 live image


schaefi:
Without the .ssh directory for the migration user the key setup will fail and no ssh login is possible during migration

schaefi:
tested by chuller and proven to work

Issue #349, opened 2025-08-06
Migration SLES4SAP 15SP7 to 16 - problem with product definition


angelabriel:
Starting with a SLES4SAP 15SP7 installation and following the description in https://confluence.suse.com/display/SLES16/SLES+16+Migration section 'Testing'.
`zypper in suse-migration-sle16-activation`
failed with

```
Problem: 1: nichts stellt 'product(SLES) >= 15.6' bereit, das vom zu installierenden suse-migration-sle16-activation-2.1.13-150700.1.1.noarch benötigt wird
 Lösung 1: suse-migration-sle16-activation-2.1.13-150700.1.1.noarch nicht installieren
 Lösung 2: suse-migration-sle16-activation-2.1.13-150700.1.1.noarch durch Ignorieren einiger Abhängigkeiten brechen
```

It seems the problem is that on SLES4SAP the **SLES_SAP-release** package is installed but on plain SLES  **sles-release**.
And the SLES_SAP-release package does not provide product() = SLES but product() = SLES_SAP
Whereas e.g. the provide product(SUSE_SLE) = 15.7 is available in both packages

```
# rpm -q --provides SLES_SAP-release | more
SLES_SAP-release = 15.7-150700.28.1
SLES_SAP-release(x86-64) = 15.7-150700.28.1
SLES_SAP-release-15.7
config(SLES_SAP-release) = 15.7-150700.28.1
displayorder(300)
distribution-release
product() = SLES_SAP
product(SLES_SAP) = 15.7-0
product(SLES_SAP-SP7) = 15.7-150700.28.1
product(SUSE_SLE) = 15.7
product(SUSE_SLE) = 15.7-150700.28.1
product(SUSE_SLE-SP7) = 15.7
product(SUSE_SLE-SP7) = 15.7-150700.28.1
product-cpeid() = cpe%3A%2Fo%3Asuse%3Asles_sap%3A15%3Asp7
product-endoflife()
product-label() = SLE%2D15%2DSP7%2DSAP
product-register-target() = sle%2D15%2Dx86_64
product-updates-repoid() = obsrepository%3A%2F%2Fbuild.suse.de%2FSUSE%3AUpdates%
3ASLE%2DINSTALLER%3A15%2DSP7%3Ax86_64%2Fupdate
product-updates-repoid() = obsrepository%3A%2F%2Fbuild.suse.de%2FSUSE%3AUpdates%
[...]

PR #351, opened 2025-08-07
Add proper release package for SLE16 migration

schaefi:
Make sure to install sles-release otherwise the ALP-dummy-release is installed by default which leads to a broken zypper migration because the target product is wrong. This is partially related to Issue #349

### DIFF
--- a/image/generic/sle16/config.kiwi
+++ b/image/generic/sle16/config.kiwi
@@ -99,5 +99,6 @@
         <package name="glibc-locale"/>
         <package name="cracklib-dict-full"/>
         <package name="ca-certificates"/>
+        <package name="SLES-release"/>
     </packages>
 </image>


greygoo:
I checked with frederics build, ans here, after booting the live system, /etc/product.d/baseproduct has sles16, it comes from a SLES-16-release package:

```
migration@localhost:/etc/products.d> grep -R 16 *
baseproduct:  <version>16.0</version>
baseproduct:  <cpeid>cpe:/o:suse:sles:16.0:server</cpeid>
baseproduct:    <name>SUSE Linux Enterprise Server 16.0</name>
...
migration@localhost:/etc/products.d> rpm -qf /etc/products.d/baseproduct 
SLES-release-16.0-160000.29.7.x86_64
```

schaefi:
> That said I am wondering why we would need the release package in the migration image itself. The origin distribution is determined from the system to be migrated. And the target distribution is set by the config value.

because if you don't set it, it currently includes

ALP-dummy-release

I found out the release package for SLE16 is named

SLES-release

and added it as such. Now it's building fine

schaefi:
> we'd have to have a profile dependency such that the SAP profile include the SLES_SAP-release

This falls into the playground of the #333 initiative

PR #365, opened 2025-08-20
Workaround for enable booting on s390x

YoukouTenhouin:
This is a workaround for booting on s390x.

For now we haven't decided how to move forward with booting on s390x, so I'm creating this as a draft; basically this workaround is to extract the kernel and the initrd directly to the local filesystem and let grub boot from there. This fixes bsc#1245225.

Let me know your opinions about the workaround, and if there's some better method to solve this...

### DIFF
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -59,8 +59,13 @@ done
 ARCH=$(uname -m)
 
 if grub_file_is_not_garbage "${migration_iso}"; then
-    kernel="(loop)/boot/${ARCH}/loader/linux"
-    initrd="(loop)/boot/${ARCH}/loader/initrd"
+    if [ "$ARCH" != "s390x" ]; then # Workaround for s390x grub loopback issues
+	kernel="(loop)/boot/${ARCH}/loader/linux"
+	initrd="(loop)/boot/${ARCH}/loader/initrd"
+    else
+	kernel="/migration-image/linux"
+	initrd="/migration-image/initrd"
+    fi
     boot_device_id="$(grub_get_device_id "${GRUB_DEVICE_BOOT}")"
     printf "menuentry '%s' %s \${menuentry_id_option} '%s' {\n" \
         "${OS}" "${CLASS}" "Migration-${boot_device_id}"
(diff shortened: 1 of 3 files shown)


schaefi:
Hi, so if I understood this correctly the following happens

1. the patch to the suse-migration-rpm wrapper causes the extraction of the kernel and initrd as separate files from the image and packages them
2. the grub.d/99_migration hook script will now create a grub entry that just loads the extracted initrd and kernel from the rootfs. At SUSE we use grub on s390 as a user space call. So there is zipl, loading some initrd and inside of that initrd a grub userspace call is issued which reads the grub config and boots the selected entry.

Ok I believe this can work

But I was questioning myself why we could not say that the loopback boot on s390 is not supported and customers need to use the kexec based method with `run_migration`. You can find this code here:

* https://github.com/SUSE/suse-migration-services/blob/master/tools/run_migration

It basically does the same what you did but boots via kexec which I think should also work on s390

Thoughts ?

YoukouTenhouin:
Thanks the reviewing this. I put this workaround together because the bugzilla report exists and I was trying to find a solution for it in case it is needed. If we decide to not support grub loopback boot on s390x then I agree this is not needed and I'll close the bugzilla ticket. How do you think? Thanks!

fcrozat:
For now, kexec was out of the picture for SLE15 => SLE16 migration, in the various hints I gave to QE and everybody else, since I was focusing on the regular way of booting system (YaST also stopped using kexec a while ago in the installer).

I'm a bit concerned of using kexec call directly in run_migration, since it will cause a immediate reboot of the system to be migrated, interrupting all running services like a power down. It would be better to use systemctl kexec to ensure the system is properly shut down before doing kexec call.

fcrozat:
Adding @mmaslano for her opinion too

mmaslano:
The _grub2zipl_: 1) I don't see as a big issue to do it SUSE way. Booting is already different on different distributions for s390x. 2) I don't like special sauce for s390x. It usually breaks at some point and it's hard to debug.

Add _systemctl kexec_ I and maybe Nikolay could test those upgrades on LPAR, but I intendeded to ask Mike anyway.

PR #366, opened 2025-09-01
Use systemctl kexec

schaefi:
Instead of direct kexec --exec let the system shutdown in a controlled way and run kexec through systemd. This is related to Issue #365


fcrozat:
tested, it works fine

rjschwei (approved):

PR #367, opened 2025-09-05
do not build sles4sap migration on s390x

fcrozat:
SLES for SAP is only available on x86_64 and ppc64le, do not build it for other architectures.

### DIFF
--- a/image/product/sle16/sles_sap/config.kiwi
+++ b/image/product/sle16/sles_sap/config.kiwi
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- OBS-ExclusiveArch: ppc64le x86_64 s390x -->
+<!-- OBS-ExclusiveArch: ppc64le x86_64 -->
 
 <image schemaversion="7.4" name="SLES16-SAP_Migration">
     <description type="system">


rjschwei (approved):

schaefi:
Hmm, that is unexpected. We switch it off completely for s390 ? why ? and this is not a SAP scope it's a general arch scope, no migration image on s390

@fcrozat As the PR has no description can you please explain the motivation, did I miss something

fcrozat:
> no description provided

fixed, this is only to disable building SLES_SAP on s390x, since this doesn't build there at all.

schaefi:
Hmm, but that's not true, it builds on all archs. See

https://build.opensuse.org/package/show/home:marcus.schaefer:dms/SLES16-Migration
https://build.opensuse.org/package/show/home:marcus.schaefer:dms/SLES16-SAP_Migration

and I also tested the upgrade of an s390 cloud instance in https://cicmgmt.prg2.suse.org/icic/login.html
using the kexec based startup

Confused ?

schaefi:
> SLES_SAP-release doesn't exist on s390x (we don't built it there). And it is not a product we (SUSE) create for this platform. SLES_SAP is only available on x86_64 and ppc64le. 

ok that makes sense to me, thanks for the clarification. But why do you disable the build for the generic image ? As I said the migration live image for SLES SAP is provided here:

* https://github.com/SUSE/suse-migration-services/tree/master/image/product/sle16/sles_sap

I'm fine if you add the OBS exclusive arch to this build.

Please also note that the wrapper RPM package for the live migration images has a product requirement set. The product requirement for the generic image is `product(SLES)` and will therefore be not installable on SLES_SAP. The product requirement
for the SLES_SAP image is `product(SLES_SAP)` 

I really doubt that plain SLES is not supported on s390 and that's the target for the generic image

> Are we talking about SLES for SAP ?

I tested plain SLES as this was the image you modified. A standard SLE15 to SLE16 migration (not SAP) worked on s390 with the generic image build

Again I'm fine if you disable s390 but please on the right image

Thanks

schaefi (approved):

PR #368, opened 2025-09-09
Add wicked2nm network migration

jcronenberg:
This PR adds a network migration from wicked to NetworkManager using https://github.com/openSUSE/wicked2nm.
It relies on some files to be present in `/system-root/var/cache` because wicked isn't present in the migration image, but `wicked2nm` relies on the xml generated by `wicked show-config`. So the approach is to have those generated and placed at the correct locations by rpm scriptlets. However I took care that it should not outright fail if not present and instead just skip the `wicked2nm` part.
There also is some special handling for migration from persistent to predictable names to keep the names persistent for existing interfaces. This again relies on some generated rules from rpm scriptlets.
The general approach is to mirror the network setup as close as possible to the host to allow reliable internet connection while the migration is happening and then copy that setup to the migrated system.


fcrozat:
we also need to integrate the various specfiles changes in this PR

suse_migration_services/units/setup_host_network.py
```diff
@@ -127,3 +130,65 @@ def log_network_details():
...
+    s390_rules = glob.glob(root_path + '/etc/udev/rules.d/*qeth*.rules') + glob.glob(root_path + '/etc/udev/rules.d/*-cio-ignore*.rules')
+    if s390_rules:
+        Command.run(['cp'] + s390_rules + ['/etc/udev/rules.d/'])
+
+    log.info('Running udevadm to apply network udev rules')
+    Command.run(['udevadm', 'control', '--reload'])
+    Command.run(['udevadm', 'trigger', '--type=subsystems', '--action=add'])
+    Command.run(['udevadm', 'trigger', '--type=devices', '--action=add'])
+    Command.run(['udevadm', 'trigger', '--subsystem-match=net', '--action=add'])
+    Command.run(['udevadm', 'settle'])
```
schaefi, line 194: I believe this implementation could be done more generic. How about the following:

Take a look at `suse_migration_services/units/post_mount_system.py` , extend the code block that handles the `for _, preserve_files in preserve_info.items():` in a way that glob patterns or entire directories are supported.

If this is done remove the code you added here with `if s390_rules` and the udevadm calls because that's already done in the post_mount service.

As a last action edit the image descriptions for the live image here:

* image/generic/sle16/config.sh

and place something like this

```bash
if [ "$(arch)" = "s390x" ];then
    cat >/etc/migration-config.yml <<-EOF
    preserve:
      rules:
      - /etc/udev/rules.d/cio-ignore*.rules
      - /etc/udev/rules.d/*qeth*.rules
    EOF
fi
```

schaefi: It would also be possible to create a package that provides the above `/etc/migration-config.yml` file. In this case no config.sh script is needed but just a `<package name="some" arch="s390x"/>`

I leave it up to you which way to go

jcronenberg: Well for the s390 handling I tend to agree. However why I initially wrote it this way and put it into the `setup-host-network` is because of the `70-migration-persistent-net.rules`. This is a special rule that only applies to the migration system and is not supposed to be applied to the migrated system. So I would keep this part, but move the s390 part to `migration-config.yml`.

schaefi: sounds good to me

schaefi: oh good point. Yes I overlooked that there is already an overlay file. So yes if we go with a package that overlay file and my script code needs to go away and the package provides some data arch specific.

If we stick with as it is now, my script suggestion needs to change to `cat >>/etc/migration-config.yml <<-EOF` to prevent overwriting.

Or you find another solution that is more appealing :) But I guess you see the point where we want to go right ?

jcronenberg: I've added the globbing to preserve_files and added the s390 specific rules to the `config.sh` like you suggested.

jcronenberg:
I added the relevant parts to the spec file templates and kiwi files for SLE16. I am however not sure how to handle the others like container and SLES_SAP. Do you think those should receive the same changes or leave it as is?

image/generic/sle16/root/etc/migration-config.yml
```diff
@@ -1 +1,3 @@
 soft_reboot: false
+network:
+  wicked2nm-continue-migration: true
```
jcronenberg, line 3: In my opinion this should be temporary and I added it for easier testing. I would keep the default to false, however until at least https://github.com/openSUSE/wicked2nm/issues/40 is fixed it is pretty much necessary to set this to true as otherwise even a simple dhcp connection will issue a warning.

PR #376, opened 2025-09-15
Include image changelog to version bump

schaefi:
Make sure the image changelog file also contains an entry to match with the version of the DMS

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -10,10 +10,13 @@ tag = True
 [bumpversion:file:image/pubcloud/sle15/config.kiwi]
 
 [bumpversion:file:image/pubcloud/product/sles_sap/config.kiwi]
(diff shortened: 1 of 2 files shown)


schaefi:
This commit adds three new Makefile targets to manage the version at the time of the submission of packages. As a packager when the time comes to submit DMS packages as a whole to the different targets in the distribution you usually called `bumpversion patch|minor|major` followed by `make some_package_target` followed by the SUSE submission process with the package data from `dist`

With this commit the process slightly changes in a way that you don't call `bumpversion` yourself but through one of the new Makefile targets

```
make patch|minor|major
```

The Makefile targets add a bit of sugar such that the .changes files of the image builds also contains a changelog entry matching the version bump

fcrozat (approved):

PR #377, opened 2025-09-16
architecture check before allowing migration to be activated

YoukouTenhouin:
SLE16 will panic on CPUs older than x86_64_v1 (i.e. x86_64). To avoid panicing the system we check for the architecture before activating the migration.

Fix bsc#1249593.

### DIFF
--- a/tools/run_migration
+++ b/tools/run_migration
@@ -139,6 +139,14 @@ if (( EUID != 0 )); then
     exit 1
 fi
 
+# Do not start migration for CPUs that are too old
+# SLE16 image panics on CPUs older than x86_64_v1 (i.e. x86_64) (bsc#1249593), so we need to block
+# migration before getting into the image.
+if [ "$(zypper system-architecture)" == "x86_64" ]; then
+   echo "CPU Architecture too old. Aborting migration."
+   exit 1
+fi
+
 # Set signal handler on EXIT
 trap cleanup EXIT
 
--- a/package/suse-migration-sle16-activation-spec-template
+++ b/package/suse-migration-sle16-activation-spec-template
@@ -31,6 +31,7 @@ Requires:         grub2
 Requires(post):   util-linux-systemd
 Requires(postun): util-linux-systemd
 Requires(pre):	  sed
+Requires(pre):	  zypper
 
 %description -n suse-migration-sle16-activation
 Script code to activate the SLE16 migration image to be booted at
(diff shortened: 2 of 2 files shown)


YoukouTenhouin:
For now I implemented this as a separated check instead of an unit of pre-check scripts. I would appreciate a second opinion here, but for now these are my considerations:

1. Mainly, in our current implementation some pre-checks will fail in the building containers created by OBS. Since failure to install the RPM is considered a build failure by OBS, simply letting pre-check failures preventing the installation of activation package will prevent the build of activation package.

2. Before this PR the pre-checks are basically performed in two situations: once in the installation of SLE*-Migration package, and once after booting into the migration image. Since in the first situation this doesn't prevent the installation of the activation package, and in the second case the system will panic before pre-check scripts even have a change to run, implementing this check as a pre-check will not benefic what we are trying to achieve, i.e. preventing system panic. 

So I think for now it makes sense to have this as a separate check instead of implementing it as a pre-check script. Is there any way better to do this? Thanks!

fcrozat (approved):

tools/run_migration
```diff
@@ -139,6 +139,12 @@ if (( EUID != 0 )); then
     exit 1
 fi
 
+# Prevent installation for CPUs that are too old
```
schaefi, line 142: same here, more information please

schaefi (changes requested):
code wise perfect, but really we need to add more describing information. Thanks much

schaefi:
for some reason one of my review comments was not shown.

schaefi (approved):
Thanks, this makes it clear :+1:

YoukouTenhouin:
> That said the idea of the pre-checks was that a user can run those to see if the system can be migrated in the first place or what they potentially need to fix before the attempt a migration. 

That's a good point. It might be beneficial to add this check into pre-check scripts as well so the user can have a change of catching it when running the pre-checks. I'll consider adding it and submit it as another PR. But I would suggest still keeping all the changes in this PR because:

> Now if the user insists to shoot themselves in the foot and run package install without scriptlet execution we do the check again when migration starts and exit before anything bad happens.

> However, in this particular case the attempt of a migration on an unsupported architecture will panic the system. Meaning we will not get beyond the kernel stage unless I misunderstood the initial comment. In such a case a migration should not even been tried or allowed because it will lead to some support cases of users who see their system to panic and unless it's not clear on which architecture the attempt was made, the debugging of such an issue could be time consuming and will cost the company money and headaches

Yes and even worse, because the migration image itself will also panic. In the kexec situation it is a bit better since the user can just reboot the system and not `run_migration`, but in the activation package case it is far worse, since you will have a grub menu with only 1s timeout and defaults to booting into an image that's doom to panic before it even have the chance to fail the pre-checks and roll back. So you get a system that stucks in a infinite loop of `booting into an image that panics -> reboot -> grub with only 1s timeout -> booting into an image that panics...` IMO "punishing" the user for not carefully examining the output of pre-checks is too harsh and should be avoided as much as possible.

PR #387, opened 2025-09-30
Implement x86_64 architecture version check as a pre-check

YoukouTenhouin:
This is a follow-up of #377.

It makes sense to implement x86_64 architecture version check as a pre-check as well, so user can catch the issue when running pre-checks before actually attempt to activate the migration.

### DIFF
--- a/suse_migration_services/migration_target.py
+++ b/suse_migration_services/migration_target.py
@@ -1,0 +1,78 @@
+
+# Copyright (c) 2025 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+import logging
+import yaml
+import os
+import platform
+
+from suse_migration_services.defaults import Defaults
+from suse_migration_services.command import Command
+
+log = logging.getLogger(Defaults.get_migration_log_name())
+
+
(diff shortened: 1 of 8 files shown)


suse_migration_services/migration_target.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+        if sles15_migration.returncode == 0:
+            # return default migration target
+            return {
+                'identifier': 'SLES',
+                'version': '15.3',
+                'arch': platform.machine()
+            }
+        sles16_migration = Command.run(
+            ['rpm', '-q', 'SLES16-Migration'], raise_on_error=False
```
fcrozat, line 69: it won't work with SLES_SAP16-Migration

schaefi (changes requested):
Let's use this and introduce a general method to detect the target.

suse_migration_services/migration_target.py
rjschwei: Shouldn't we also check for the existence of the configuration file and check the migration target if it exists?

tools/run_migration
```diff
@@ -167,6 +159,14 @@ if [[ "${migration_iso}" == *SLES16*-*Migration*.iso ]]; then
 fi
 
 if ${is_sles16_migration}; then
+    # Do not start migration for CPUs that are too old
+    # SLE16 image panics on CPUs older than x86_64_v2 (i.e. x86_64) (bsc#1249593), so we need to
+    # block migration before getting into the image.
+    if [ "$(zypper system-architecture)" == "x86_64" ]; then
+	echo "CPU Architecture too old. Aborting migration."
+	exit 1
+    fi
```
schaefi, line 168: nitpick, can you fix the indentation. Thanks

suse_migration_services/migration_target.py
```diff
@@ -0,0 +1,77 @@
...
+            return {
+                'identifier': 'SLES',
+                'version': '16.0',
+                'arch': platform.machine()
+            }
+        else:
+            # return default migration target
+            return {
+                'identifier': 'SLES',
+                'version': '15.3',
```
schaefi, line 75: Let's change this to ```15.4```, 15 SP3 is outside the transition period

schaefi (changes requested):
Thanks, just a few small hints. Nice work :+1:

suse_migration_services/migration_target.py
YoukouTenhouin: Done👌

schaefi (approved):

PR #389, opened 2025-10-01
bind mount only required subdirectories under /run into chroot

YoukouTenhouin:
bind mounting the entire /run directory in chroot confuses some packages (e.g. systemd-coredump) and may cause issues. This commit change the behaviour to only mounting required subdirectories.

Fix bsc#1250078

### DIFF
--- a/suse_migration_services/units/mount_system.py
+++ b/suse_migration_services/units/mount_system.py
@@ -179,9 +179,20 @@ def mount_system(root_path, fstab):
             system_mount.add_entry(
                 mount_type, mount_point
             )
-        log.info('Bind mount /run inside chroot {0}'.format(root_path))
+        log.info('Bind mount subdirectories from /run inside chroot {0}'.format(root_path))
+        os.makedirs(os.sep.join([root_path, 'run', 'NetworkManager']), exist_ok=True)
         Command.run(
-            ['mount', '-o', 'bind', '/run', os.sep.join([root_path, 'run'])]
+            [
+                'mount', '-o', 'bind', '/run/NetworkManager',
+                os.sep.join([root_path, 'run', 'NetworkManager'])
+            ]
+        )
+        os.makedirs(os.sep.join([root_path, 'run', 'netconfig']), exist_ok=True)
+        Command.run(
+            [
+                'mount', '-o', 'bind', '/run/netconfig',
+                os.sep.join([root_path, 'run', 'netconfig'])
+            ]
         )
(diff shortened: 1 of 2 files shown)


fcrozat (approved):

suse_migration_services/units/mount_system.py
```diff
@@ -179,9 +179,24 @@ def mount_system(root_path, fstab):
             system_mount.add_entry(
                 mount_type, mount_point
             )
-        log.info('Bind mount /run inside chroot {0}'.format(root_path))
+        log.info('Bind mount subdirectories from /run inside chroot {0}'.format(root_path))
         Command.run(
-            ['mount', '-o', 'bind', '/run', os.sep.join([root_path, 'run'])]
+            ['mkdir', '-p', os.sep.join([root_path, 'run', 'NetworkManager'])]
```
schaefi, line 184: Since there is a tendency to use `os.makedirs` it can be applied here too

suse_migration_services/units/mount_system.py
```diff
@@ -179,9 +179,24 @@ def mount_system(root_path, fstab):
...
+        )
+        Command.run(
+            [
+                'mount', '-o', 'bind', '/run/NetworkManager',
+                os.sep.join([root_path, 'run', 'NetworkManager'])
+            ]
+        )
+        Command.run(
+            ['mkdir', '-p', os.sep.join([root_path, 'run', 'netconfig'])]
```
schaefi, line 193: same here `os.makedirs`

suse_migration_services/units/mount_system.py
```diff
@@ -179,9 +179,24 @@ def mount_system(root_path, fstab):
...
+        )
+        Command.run(
+            [
+                'mount', '-o', 'bind', '/run/netconfig',
+                os.sep.join([root_path, 'run', 'netconfig'])
+            ]
```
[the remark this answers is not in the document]
schaefi: @fcrozat yes you are right, this was introduced from your side in #311 and was never a part of the 12-to-15 migration. Sorry I thought it existed for longer. So we are fine here as testing 15-to-16 migration is something I test pretty often these days :)

schaefi:
@rjschwei

> We probably also need to mount /run/cloud-init if it exists

Can you explain why ? We didn't had that for 12-to-15 migrations and they also had cloud init. Is this different on 15-to-16 ? Thanks

rjschwei (approved):

PR #394, opened 2025-10-03
shrink migration image

fcrozat:
remove a lot of things in the image, based on the work done on Agama Live ISO and Tumbleweed Live ISO

The two python scripts are a port of the Agama ruby script by Gemini Code Assist

### DIFF
--- a/image/generic/sle16/config.kiwi
+++ b/image/generic/sle16/config.kiwi
@@ -7,7 +7,7 @@
         <contact>public-cloud-dev@susecloud.net</contact>
         <specification>
             SLES16 Distribution Migration Live System to upgrade
-            SLES15-SP6+ instances to SLES16+
+            SLES15-SP7+ instances to SLES16+
         </specification>
     </description>
     <preferences>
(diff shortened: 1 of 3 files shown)


schaefi (changes requested):
see my comments. If at all this should be done in an extra profile. All that overlay code needs to be provided in a proper package and I strongly recommend test covered for code

fcrozat:
The migration RPM needs to be made available on maintenance update channel (rfight now, only 15 SP7 Basesystem update channel, but in the future, to other SP). 

It is a huge file (currently 974MB) and we need it twice (SLES and SLES_SAP, the latter going to SAP module).

This is a concern which was raised by MLM team, because they already have customers complaining about the size of SLE repositories (without DMS being published yet). And each release of DMS as maintenance update will increase the size of the repositories a lot (even if we don't plan to release it too often, it is a problem).

On the hardware support, by reusing agama cleanup list, we can garantee what was booting with Agama will boot with minified DMS. 

Agama people are willing to move their "minification" script and list to a dedicated package we could shared with them, in the future.

schaefi:
[smaller.patch](https://github.com/user-attachments/files/22730325/smaller.patch)

with this patch the image gets down to

```
-rw-r--r-- 1 ms users 341M Oct  6 22:16 SLES16-Migration-2.1.24-5.1.x86_64.rpm
-rw-r--r-- 1 ms users 366M Oct  6 22:16 SLES16-Migration.x86_64-2.1.24-Build5.1.iso
```

I must admit I have not tested it.

But I'm convinced the size of the image can be reduced by just reducing the set of packages.

schaefi:
> That is not sufficient. I would suspect that the Agama image is not boot tested in the Public Cloud ...

I fully agree. The assumption that agama and its tests forms a proper hardware base for system migrations is most likely not correct. PublicCloud is one example and as Agama does not even implement a migration workflow but only an installation workflow it is pretty dangerous to put both in the same bucket for requirements. I will go that far that the Agama people care for agama but surely not for the DMS. As such even with the scripts properly packaged there is no connection to our use cases. If you find out there is an issue with the scripts and we require a change, how likely is it that they accept that when they don't have this problem ?  I would assume they don't care. That would lead us branching the package or carrying custom patches ... a nightmare

This shrinking will cause us only trouble and I know  that because it's not the first time people thought it would be a good idea. It's always the same story. You shrink and something doesn't work, you add a bit back, and something else doesn't work, you add another bit back ... and in the end and after endless unhappy customer stories you are back at the full definition. It really hurts and I do not understand why we have space trouble on repos ? they grow always :) and not only because of our stuff. As such there must be a strategy for handling data. The strategy cannot be to force us shrinking an image to loose functionality

I simply having a hard time to accept broken migrations because we have not enough space to store a package.

Removing stuff that is not needed, beginning with that pattern package that pulls in stuff not needed is totally fine with me. 
We also surely never need sound firmware and most probably nothing from the bluetooth stack

Reducing the kernel by frankenstein surgery is a pita and will strike back 100% ... based on my imaging experience

schaefi:
>This will be added in the 16 migration ECO

ok great. I picked your version to my devel project such that on merge things still builds. I will merge and update my devel project with this change today. After that I'll repeat my integration testing for the public cloud target next week

PR #395, opened 2025-10-06
Fixed reading of migration config for target class

schaefi:
The MigrationTarget class implementation duplicates the reading of the config file for which we have the MigrationConfig class. This commit fixes the MigrationTarget by using an instance of the MigrationConfig class such that the latest changes in reading the configuration becomes effective


schaefi:
I'm in the process of testing the DMS for cloud use and stumbled across many issues for which I'm going to submit PRs.

rjschwei (approved):

Issue #393, opened 2025-10-02
Malicious user can trigger recursion overflow


rjschwei:
The configuration merge algorithm introduced in https://github.com/SUSE/suse-migration-services/pull/390 does not have a safeguard against a maliciously introduced recursion attack.

https://github.com/SUSE/suse-migration-services/blob/master/suse_migration_services/migration_config.py#L90

will recurs until recursion depth exception is reached when a specific config file is constructed, for example:

dict1 = { 'foo': { 'foo' :  'bar'}}

dict2 = { 'foo': {'foo': {'foo': {'foo' .........}}}}}

When dict to is an "endlessly" nested dict with the same key the algorithm will eventually reach the recursion depth limit and trigger a traceback.

We should have a recursion depth limit counter and when reached break out of the recursive loop.

It may also make sense to only accept known values in the configuration.

cfconrad:
> It may also make sense to only accept known values in the configuration.

Thanks for bring this up, I thought it isn't possible, but I will double check regarding the validation.

My assumption was:
 * Each config we read/merge will be loaded [with _parse_config_file()](https://github.com/SUSE/suse-migration-services//blob/72366fb68b7e20435f3b390724582c0c1f117f37/suse_migration_services/migration_config.py#L99)
 * this calls  [validator.validate(config_data)](https://github.com/SUSE/suse-migration-services//blob/72366fb68b7e20435f3b390724582c0c1f117f37/suse_migration_services/migration_config.py#L65)
 * which makes the read content valid to this [schema](https://github.com/SUSE/suse-migration-services//blob/72366fb68b7e20435f3b390724582c0c1f117f37/suse_migration_services/schema.py#L1)

Anyway, I think it makes sense to have a sanity check against max recursuions in [_merge_config_dicts()](https://github.com/SUSE/suse-migration-services//blob/72366fb68b7e20435f3b390724582c0c1f117f37/suse_migration_services/migration_config.py#L81) too!


I played with self referencing aliases like:
```
anchor_map: &self_ref
  name: "Top Level"
  parent: *self_ref

data:
  - item: "A"
  - item: "B"
  - item: *self_ref # Alias used again here
```

But the [yaml.safe_load()](https://github.com/SUSE/suse-migration-services//blob/72366fb68b7e20435f3b390724582c0c1f117f37/suse_migration_services/migration_config.py#L64) simply treat these as `strings`, so I didn't took extra checks for it, even without calling `validate()`, right?

[...]

PR #397, opened 2025-10-06
Add recursion guard to MigrationConfig::_merge_config_dicts

cfconrad:
Implements a recursion guard in `MigrationConfig::_merge_config_dicts()`
to prevent endless loops.

While currently only used with schema-validated dictionaries, 
future use in different contexts necessitates this safeguard.

The current limit of 2 reflects the maximum nesting depth
defined in the migration-config schema.

Fixes: https://github.com/SUSE/suse-migration-services/issues/393

### DIFF
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -78,7 +78,7 @@ class MigrationConfig:
         return config_data
 
     @staticmethod
-    def _merge_config_dicts(dict1, dict2):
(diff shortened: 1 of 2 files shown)


schaefi:
@cfconrad can you add a description please. Even though the commit title is good, some words on the initial description are helpful as they stay even if the commits gets rebased/reworded etc ... Thanks

PR #398, opened 2025-10-07
Warning user about sshd root login change before migration

### LINKED ISSUE #380: When migrating from SLE15 to SLE16, if sshd is enabled, warn user about deprecated root password authentication
When migrating to SLE16, we should warn users, if sshd is enabled (unless they configured SSHD to not allow root authentication) they might want to enable it back before starting the migration by creating a file 
`/etc/ssh/sshd_config.d/50-permit-root-login.conf` containing:

`PermitRootLogin yes`

Once the migration is complete, they can install openssh-server-config-rootlogin package and remove the previously added file.

YoukouTenhouin:
Root login by ssh is disabled by default on SLE16. We should warn user about this and give them a chance to enable it manually before migration actually starts.

Fixes #380


rjschwei (changes requested):

suse_migration_services/prechecks/sshd.py
```diff
@@ -0,0 +1,63 @@
...
+You might want to enable it back before starting the migration.
+To do so, create /etc/ssh/sshd_config.d/50-permit-root-login.conf
+with the following content:
+
+PermitRootLogin yes
+
+Once the migration is complete, you can install
+openssh-server-config-rootlogin package and remove the previously
+added file."""
+    )
```
schaefi, line 63: Can we please use 

```python3
from textwrap import dedent

message = dedent('''\n
    your long text
    and more long text
''')

log.warning(message)
```

It is so much better to read compared to the inline text messages. Thanks

suse_migration_services/prechecks/sshd.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    system_root = Defaults.get_system_root_path() if migration_system else '/'
+    sshd_enabled = Command.run(
+        [
+            'systemctl',
+            '--root', system_root,
+            '--quiet',
+            'is-enabled', 'sshd'
+        ]
+    ).returncode == 0
```
schaefi, line 40: If you use sshd_enabled in a bool condition check, as you do, then please turn it into a bool prior use

```
some = Command.run(...)
some_has_failed = bool(some.returncode)

if some_has_failed:
    # do something when failed
```
or the respective notation for "command has failed equals something is enabled or not"

The way it is written here is imho hard to read

suse_migration_services/prechecks/sshd.py
```diff
@@ -0,0 +1,63 @@
...
+log = logging.getLogger(Defaults.get_migration_log_name())
+
+
+def root_login(migration_system=False):
+    """
+    Check whether sshd is enabled and root login is allowed. If so,
+    warn user that they might want to enable it back before starting
+    the migration.
+    """
+    system_root = Defaults.get_system_root_path() if migration_system else '/'
```
schaefi, line 32: This code runs as pre-check on the host as well as during migration. During migration the DMS sets up the ssh capabilities itself and it is immaterial to check for them as they are all gone after migration. So correct me when wrong but the pre-check only makes sense on the host prior migration and then never again ?

If so, please differentiate this case and add

```python3
if migration_system:
    # This check should not run during migration
    return
```

schaefi (changes requested):
Nice pre-check :+1: 

Just a few comments that we need to address prior merge. Thanks much

suse_migration_services/prechecks/sshd.py
schaefi: Also you need to add

```
raise_on_error=False
```

to the ```Command.run``` invocation as we don't want to raise in case of failure

suse_migration_services/prechecks/sshd.py
YoukouTenhouin: Fixed. I was thinking about this as well, thanks for pointing out!

PR #399, opened 2025-10-07
Fixed behavior of wicked2m pre check

schaefi:
The output produced by the pre-check is in my opinion very misleading. It outputs information from wicked2m which says that warnings can be ignored by passing an option to the tool but for the actual DMS system migration this has no meaning. The additional information to place a config setting to a DMS config file is correct but completely confusing with regards to the first information provided. This commit refactors the code which also doesn't use the existing command interface and imho can be done simpler. In addition the precheck runs code when inside of the migration, but this code is useless because the real (no dry-run) call of wicked2nm happens in the setup_host_network service.

### DIFF
--- a/suse_migration_services/prechecks/wicked2nm.py
+++ b/suse_migration_services/prechecks/wicked2nm.py
@@ -15,61 +15,69 @@
 # You should have received a copy of the GNU General Public License
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
-"""Call prechecks for HA provided by crmsh"""
 import logging
 import os
-import subprocess
 import shutil
 
-from suse_migration_services.defaults import Defaults
+from tempfile import NamedTemporaryFile
 
+from suse_migration_services.defaults import Defaults
+from suse_migration_services.command import Command
 
 log = logging.getLogger(Defaults.get_migration_log_name())
 
 
 def check_wicked2nm(migration_system=False):
+    """
+    Run wicked2nm in dry run mode to check if a migration from
+    a wicked based network setup to NetworkManager is possible
+    """
     wicked_config_path = '/var/cache/wicked_config'
-    netconf_dir_path = '/etc/sysconfig/network'
+    wicked_config = os.sep.join(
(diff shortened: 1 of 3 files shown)


rjschwei (approved):

suse_migration_services/prechecks/wicked2nm.py
```diff
@@ -15,61 +15,69 @@
...
-            return False
-
-
-def wicked2nm_log_error(output):
-    msg = 'Wicked to NetworkManager migration pre-check failed:{0}{1}'.format(os.linesep, output)
-    if '--continue-migration' in output:
-        msg = msg + """If you want the migration to ignore the warnings add
-network:
-  wicked2nm-continue-migration: true
-to /etc/sle-migration-service.yml"""
```
jcronenberg, line 74: Personally I feel like the new message is a downgrade, but I guess that's personal taste.

suse_migration_services/prechecks/wicked2nm.py
```diff
@@ -15,61 +15,69 @@
...
+    if wicked2nm_result.returncode != 0:
+        log.warning('wicked2nm failed with:')
+        for relevant_error in wicked2nm_result.error.split(os.linesep):
+            if 'WARN' in relevant_error:
+                log.warning(relevant_error)
+        log.warning(
+            'To ignore the warning(s) set: {} in: {}'.format(
+                'wicked2nm-continue-migration: true',
+                Defaults.get_migration_host_config_file()
             )
```
jcronenberg, line 82: I had specifically looked for `--continue-migration` in the error because warnings may appear however when a different critical error occurs it won't work even with `--continue-migration` set. But then the error will look different and not contain `--continue-migration`.
Also if you define only warnings as relevant, first off I would heavily disagree, because the `INFO`s may still be relevant for users to know, and second there is also the `--log-level` option to restrict it to just warnings.

jcronenberg (commented):
Just for the future please at least involve me or @cfconrad in the discussion if changes to this part are made.

PR #406, opened 2025-10-15
Fix setup_host_network_test by mocking os.makedirs

jcronenberg:
When '/etc/udev/' doesn't exist the test fails. But the call should be mocked anyway and not actually executed on the system.


schaefi (approved):
I fully agree and great catch, thanks :+1:

PR #408, opened 2025-10-16
Simplify interface naming by disabling predictable names at boot

jcronenberg:
Currently our most difficult problem with regards to the network interface
naming scheme is that when the migration image boots the interfaces get
predictable names like 'enp1s0' which we then have to revert/write custom rules
around. To simplify this we can install a file at
/etc/systemd/network/99-default.link. This skips all naming schemes and the
kernel names are used. Then we can remove this file again and copy over all the
necessary udev rules and link files from the host and with udevadm trigger apply
these. This should imitate the behavior of the final migrated system most
closely.


jcronenberg:
#409 is a different approach. Both fix the issues we are currently facing with VMware/Hyper-V. Please let us know what you think is better.

jcronenberg:
> My question is, do we want to detect this situation in a pre-check and warn about it (in a new PR) or do you think the conditions when this can cause an issue are too difficult to be detected properly ?

Yes I think at least some sort of pre-check that warns of scenarios that we identified as not supported definitely makes sense. I will implement that as well.

rjschwei (approved):

jcronenberg:
I updated the PR with what we think is a better approach. Now we add a `/etc/systemd/network/99-default.link` file that disables the predictable naming scheme. Then we can remove this file and copy over udev rules and link files and run the `udevadm trigger`. This should mimick pretty much the exact behavior of the final migrated system. So e.g. when no `70-persistent-net.rules` is present, interfaces get the "new" predictable name. It should even work if the SLE15 system is using `net.ifnames=1` and link files.

Still need to update the unit tests though.

schaefi:
@jcronenberg looks good and matches what we tested today :+1: 

> Still need to update the unit tests though

yep, just drop the don't merge label when you see fit. With this also #409 becomes obsolete iirc ?

cfconrad (approved):
I also prefer this solution much more then https://github.com/SUSE/suse-migration-services/pull/409!

I have tested this with HyperV and VMWare guests. Also with `net.ifnames=1` it worked.

What I haven't tested, if the customer has a `99-default.link` file with some different naming-policies. But from code, I wouldn't expect a problem, right?

LGTM

jcronenberg:
> What I haven't tested, if the customer has a `99-default.link` file with some different naming-policies. But from code, I wouldn't expect a problem, right?

Yes that was my thought process for removing the file before any host rules and link files are copied.

jcronenberg:
CI looks good now https://github.com/jcronenberg/suse-migration-services/actions/runs/18681444700.
It's good to go from my side.

PR #413, opened 2025-10-23
Fix apparmor install procedure

schaefi:
The apparmor to selinux unit install the selinux pattern after the main migration has been performed. This zypper call has to happen inside of the new root system otherwise repo access is denied on e.g cloud systems which holds the registration information in the installed system.

### DIFF
--- a/suse_migration_services/zypper.py
+++ b/suse_migration_services/zypper.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 #
+import os
 from typing import List
 
 # project
@@ -29,7 +30,6 @@ class Zypper:
     """
     **Implements zypper invocation**
     """
-
     @staticmethod
     def run(args: List[str], raise_on_error=True, chroot=''):
         """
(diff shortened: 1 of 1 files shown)


schaefi:
In general after the main upgrade is done, any subsequent installation procedures should run in the upgraded system rather than in the live migration image running zypper --root. Same concept like in kiwi.

With this approach I could successfully migrate all the test VMs, the QA VM and a cloud instance (AWS only at the moment)

The installation of the selinux pattern however complained in some post script saying 
```
%posttrans(passt-selinux-20250415.2340bbf-160000.2.2.noarch) script output:
++ SELINUX=enforcing
++ SELINUXTYPE=targeted
%transfiletriggerin(systemd-257.7-160000.2.2.x86_64) script output:
Failed to connect to system scope bus via local transport: No data available
error: lua script failed: [string "%transfiletriggerin(systemd-257.7-160000.2.2...."]:5: exit code
Error from %transfiletriggerin(systemd-257.7-160000.2.2.x86_64)
```

Not sure what exactly it is doing here but I ignored the error for now because to activate selinux a reboot of the system with the new kernel options and the relabeling is required anyway. Within the system after reboot I see selinux functioning properly and the labels were applied and runs against the installed policy

Thanks

schaefi:
directly after the migration 

```
ec2-user@ip-172-31-14-183:~> sestatus 
SELinux status:                 disabled
```

expected ok, after reboot

```
ec2-user@ip-172-31-14-183:~> sestatus 
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      33
```

LGTM

rjschwei (approved):

PR #419, opened 2025-11-13
Omit cio_ignore kernel commandline on zkvm

cfconrad:
Based on bsc#1250003 make the migration
more robust, when removing `cio_ignore`
from commandline only for zKVM.


schaefi (approved):
good catch and makes sense to me. Actually only on LPARs this should be done

PR #422, opened 2025-11-21
Create systemd.link files for biosdevname (bsc#1253963)

cfconrad:
This pull request solve the problem of https://bugzilla.suse.com/show_bug.cgi?id=1253963

As we get the next udev-rules creation loop, I moved all these network pre requirement tasks
into a shell library called `network_pre.functions` installed with `suse-migration-sle16-scripts` package.

### DIFF
--- a/tools/network_pre.functions
+++ b/tools/network_pre.functions
@@ -16,6 +16,73 @@
 # along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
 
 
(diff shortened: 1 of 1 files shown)


cfconrad:
Set to not merge, because I will do some more testing. But maybe @schaefi you already will look at it. We discussed in advanced, but I used different names in the end, wdyt?

PR #423, opened 2025-11-26
Move script package to the main migration provider

schaefi:
In the effort to provide a common script code package, the package suse-migration-sle16-scripts was provided. During testing I found that not all places where script duplication exists was covered. This commit addresses the code that was missed and also suggests that we move the common script code as a sub package to the main suse-migration-services package. The reason here is that the common code is not only used in the activation package. In addition I also suggest that the package name changes to be more generic. The reason here is that this will go beyond sle16 and can also contain common script code for sle15 migrations as well as for past sle16 migrations as we are moving forward. As such my suggestion to also rename the package to just suse-migration-scripts.


cfconrad (approved):
Thanks good catch!

Issue #421, opened 2025-11-19
Add service migration API to allow for uninstalling no longer needed packages


schaefi:
# Description

when moving services like apparmor-to-selinux there might be packages in the system that are no longer required after the service migration completed successfully. We want the service to be able to uninstall this packages.

The idea covers the following concept:

1. Create an API class that implements the uninstall
2. Make sure every service migration class inherits from the API class mentioned in 1.
3. Update the existing service migrations and implement the service specific API methods, e.g. list of packages to uninstall

First services to apply the concept
* units/apparmor_migration.py
* units/wicked_migration.py

PR #425, opened 2025-11-26
Add DropComponents class

schaefi:
Implement DropComponents class to allow for removal of packages, files and directories. Every service class that has the need to perform this task needs to inherit from DropComponents. Along with the new class also all service units are turned into classes for consistency and to allow the inheritance model for eventually further extensions to the service units. This is related to Issue #421


schaefi:
I updated the package on the devel build project:

As such testing the above can be done from the devel project. The expectation should be that nothing has changed ;)

PR #427, opened 2025-11-27
Fix mount_system

cfconrad:
Fixes: 15549f619


schaefi (approved):
Great catch thanks much :+1: 

I broke this, code and test :facepalm: really great you found and fixed it

PR #428, opened 2025-12-02
Fixed spec file regarding wicked2nm

schaefi:
wicked2nm was set as a requirement to suse-migration-pre-checks beginning with SLE15 onward. wicked2nm is needed as pre-check tool on a SLE15 host as well as as migration tool inside of a SLE16 migration image/container. The package suse-migration-pre-checks is therefore installed into the live migration image/container and also set as a required package to the containment package that contains the live migration image blob. The problem is that wicked2nm only exists partially in SLE15, actually only on SLE15-SP7 for the x86_64 architecture. This is causing a problem for the SLE12-to-SLE15 migration images which builds migration live images for SLE15 for SP4(SAP target) and SP7(standard target) for x86_64 and aarch64. wicked2nm is a tool not really used or needed for this migration but the live migration image is based on SLE15 and of course other pre-checks are needed which means suse-migration-pre-checks has to be installed into those migration live image builds as well and that causes a problem when wicked2nm is a requirement not provided for this target. The solution proposed in this commit is to make wicked2nm a recommended package for suse-migration-pre-checks and to make sure that all SLE16 image builds explicitly installs wicked2nm which is already the case for all SLE16 based migration image descriptions in this git repo.

### DIFF
--- a/package/suse-migration-services-spec-template
+++ b/package/suse-migration-services-spec-template
@@ -93,8 +93,8 @@ Requires:         %{pythonpkgname} = %{version}-%{release}
 Requires:         %{pythons}-Cerberus
 Requires:         %{pythons}-PyYAML
 Requires:         %{pythons}-setuptools
-%if 0%{?suse_version} >= 1500
-Recommends:       wicked2nm
+%if %{sle_version} >= 150700 || %{suse_version} >= 1600
+Requires:         wicked2nm
 %endif
 Conflicts:        suse-migration-sle15-activation < 2.0.33
 Conflicts:        suse-migration-sle16-activation < 2.0.33


fcrozat:
you could ask maintenance team to have wicked2nm built on SP4 too, so the specfile change is not needed.

fcrozat:
Maintenance will have to ship it in older service pack when we support migrating SLES 15 SP6 or earlier anyway, with SLES 16.1
We can also unbranch wicked2nm from 15-SP7 and inherit it from 15-SP4, once it is released there.

I've not in favor to drop the requires, because some customers are disabling recommends (or they are disabled by default on all MinimalVM). This will break upgrade on those systems, since wicked2nm won't be installed with pre-checks.

Intead, I would suggest to replace %if 0%{?suse_version} >= 1500 with %if %{sle_version} >= 150700 || %{suse_version} >= 1600

This way, suse-migration-services built in SP4 won't have wicked2nm

rjschwei (approved):

fcrozat:
I'm fine with this plan. I would suggest to add Libor as project manager

PR #433, opened 2025-12-08
Run wicked_migration only if wicked is the active online network target

rjschwei:
While wicked was the default network configuration code in SLE 15 NetworkManager was also available either via the Desktop Module or PackageHub. As such it is conceivable that a user does not have wicked on the system to be migrated. Check if wicked is installed before adding it to the list of packages to be remove. An error condition would be triggered when an attempt is made to remove a package that is is not installed.

### DIFF
--- a/suse_migration_services/units/wicked_migration.py
+++ b/suse_migration_services/units/wicked_migration.py
@@ -41,8 +41,18 @@ class WickedToNetworkManager(DropComponents):
         super().__init__()
 
     def perform(self):
-        try:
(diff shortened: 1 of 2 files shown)


schaefi:
> works for me, but I will double check again

ok I was wrong, this does not work

```
ip-172-31-11-55:~ # systemctl --root /system-root show -p Id network.service
Verb 'show' cannot be used with --root= or --image=.

```

```
ip-172-31-11-55:~ # chroot /system-root systemctl show -p Id network.service
Failed to connect to bus: No data available
```

PR #434, opened 2025-12-09
Drop and backup /etc/sysconfig/network

schaefi:
On successful wicked to NetworkManager migration the data in /etc/sysconfig/network is no longer needed. There will be a backup of the data available in /var/migration


cfconrad:
How should we document this? Is there some output in `/var/log/distro_migration.log`?

schaefi:
> How should we document this? Is there some output in `/var/log/distro_migration.log`?

Good point. I added log information about the action. Those will appear in the log file

PR #439, opened 2025-12-11
Fixed migration live image package requires

schaefi:
When the containment rpm package, packages the live migration ISO image into an rpm package it also adds the run_migration tool to the package. The latest changes to run_migration code however also requires the new suse-migration-scripts package which was not yet required. This commit fixes it


cfconrad (approved):
THX

PR #440, opened 2025-12-12
Only generate wicked xml if service is the active network

jcronenberg:
Currently it only checks if wicked is installed and not if it is the active network service.


cfconrad (approved):
LGTM

PR #443, opened 2025-12-17
fix: ensure NetworkManager is installed on the target system

YoukouTenhouin:
NetworkManager might not be automatically installed when some services are enabled on the migrated system (e.g. PackageHub), which will cause the migration from wicked to NM to fail. Install NetworkManager packages explicitly to fix this issue.

Ref bsc#1254388.

### DIFF
--- a/suse_migration_services/units/wicked_migration.py
+++ b/suse_migration_services/units/wicked_migration.py
@@ -24,6 +24,7 @@ from suse_migration_services.command import Command
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.logger import Logger
 from suse_migration_services.drop_components import DropComponents
+from suse_migration_services.zypper import Zypper
 
 from suse_migration_services.exceptions import DistMigrationWickedMigrationException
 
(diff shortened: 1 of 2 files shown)


YoukouTenhouin:
Made some attempt to merge package installing with DropComponents (and renamed it to ComponentsHandler). Somehow this breaks the logger test so I mocked getLogger as well for now, please let me know whether this is the correct way to fix the issue.

schaefi (changes requested):
see my comment

cfconrad:
Update to the need of `NetworkManager` installation, with "PackageHub" extension is enabled.

The migration takes place, because the network in the DMS image is setup properly with "wicked2nm". But the final os is still using `wicked`. The `distro_migration.log` contains:
```
Running wicked to NetworkManager migration
Enabling NetworkManager in migrated system
Calling: ['systemctl', '--root', '/system-root', 'enable', '--force', 'NetworkManager.service']
EXEC: Failed with stderr: Failed to enable unit: Unit NetworkManager.service does not exist
, stdout: (no output on stdout)
wicked to NetworkManager migration failed with systemctl: stderr: Failed to enable unit: Unit NetworkManager.service does not exist
, stdout: (no output on stdout)
```

So I think it is necessary to enforce the installation, even if we do not have the `use_zypper_migration: true` case.

PR #449, opened 2026-02-02
Update documentation for 12-to-15 in pubclouds

schaefi:
Fix information about default service pack target. Add information about product specific migrations
Avoid repetitive and confusing information about SLES and the supported upgrade path


tomschr (commented):
@schaefi Looks good! 👍 One small addition, see below.

> I can't remember the details what needs to be done to update it. Thanks much

If you have merged this PR and needs a rebuild, just ping me via Slack. :slightly_smiling_face:

PR #453, opened 2026-02-08
Fix SLES SAP migration 12 - 15 in public clouds

schaefi:
Workaround an issue in the SLE15 SP4 based live migration system used for SAP migrations only. The zypper version on SLE15 SP4 is too old and does not support --gpg-auto-import-keys. This workaround patches the DMS code to not use this option. As soon as the SAP migration live image can switch to a newer SP, this workaround can hopefully be removed again.

### DIFF
--- a/image/pubcloud/product/sles_sap/config.kiwi
+++ b/image/pubcloud/product/sles_sap/config.kiwi
@@ -29,22 +29,22 @@
         <source path='obsrepositories:/'/>
     </repository>
     <repository type="rpm-md">
-     <source path="obs://SUSE:SLE-15-SP4:Update/standard"/>
+        <source path="obs://SUSE:SLE-15-SP4:Update/standard"/>
     </repository>
     <repository type="rpm-md">
-     <source path="obs://SUSE:SLE-15-SP4:GA/standard"/>
+        <source path="obs://SUSE:SLE-15-SP4:GA/standard"/>
     </repository>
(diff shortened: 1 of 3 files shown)


schaefi:
I'm actually not very happy with this change but it is the simplest change that we can keep under our control and that we can also easily throw away when the time comes. The alternatives are:

* A proper zypper update in SLE15 SP4, I doubt it will ever come
* A code implementation in the DMS to check for command capabilities, possible but over engineered for questionable reasons
* Do not use --gpg-auto-import-keys in general, unforeseeable side effects possible

As such, the proposed solution is easy to maintain and only temporary

schaefi:
> Why are we patching the installed code and not the sources during package build with an `%if` in the spec file?

because this only applies to the SAP migration where we are going with SLE15 SP4 which is an SP version that has this old zypper. Changing the spec of the suse-migration-services rpm package and carrying a custom patch in the package is a bigger maintenance nightmare and also opens the risk to affect other targets. Managing this in the image description surely doesn't harm other targets and the live migration images can also be submitted individually as a maintenance update without harming other components of the DMS as a whole.

To be honest I suffer from maintenance a lot with the DMS, they break my submissions, they ignore the special rules of this project and all that will not change. The DMS maintains a common code base for migrations and I don't want to run again into a painful update especially not for a target for which I simply drop the patch from the image description when SP4 becomes EOL for SAP.

PR #455, opened 2026-02-13
Fixed disk device name passed to azuremetadata

schaefi:
When setting the disk device name in the regionserverclnt.cfg for azure we used the unix device node, e.g /dev/sda. This is not a persistent device name and may change. This commit uses the by-id representation for the disk device from udev which is not expected to change.


schaefi:
Doesn't work

```
EXEC: Failed with stderr: lsblk: unknown column: ID-LINK
```

lsblk on SLE15 does not support this :(((

PR #457, opened 2026-02-17
Renew infrastructure registration prior migration

schaefi:
In case of a migration inside of the SUSE public cloud infrastructure make sure to renew the registration such that access to the repository server is working during migration


schaefi:
According to Andreas Mach testing it does not work, still the same permission denied errors on Azure

PR #461, opened 2026-02-23
Switch reboot default

rjschwei:
In light of recent issues reported with soft reboot we are chaning the reboot method to a full reboot. This is intended to reduce the number of requests we receive to investigate boot issues.

### DIFF
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -171,20 +171,24 @@ tux > sudo zypper in SLES15-Migration
 ----
 +
 The `run_migration` uses `kexec` to boot into the kernel delivered with the
-upgrade image delivered by the SLES15-Migration package. Once this system
-is live after the `kexec` the distribution migration process is automatically
-started. However, `kexec` is not supported and does not function in certain
-conditions. For example, the `run_migration` utility does not work in Xen
-based environments or when kernel keys change.
+upgrade image delivered by the SLES{$VERSION}-Migration package.
 +
-If kexec causes a kernel panic this can cause the system to hang and the
-distribution migration to fail. In that case refer to this TID:
-https://www.suse.com/support/kb/doc/?id=000019733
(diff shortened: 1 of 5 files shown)


rjschwei:
Marking this as WIP for now as we need some more conversation about this change. Therefore this should be considered a proposal.

We've had a number of long standing issues where kexec was involved. The proposal is to change the default and use a full reboot, which is more reliable. Those that desire more speed during the upgrade process can switch to using kexec and a soft reboot via changing the configuration. Some systems take a very long time to boot and as such having a more speedy, potentially successful reboot method may be desirable to enable.

schaefi:
I'm fine with the change to set the reboot after the migration to be a full reboot instead of rebooting via kexec.

>kexec is currently only mandatory for s390x but only for starting the migration.

right and this is enforced by us not providing an `activation` package that modifies the bootloader config. The `run_migration` method is the only you have in this case

So once the checks are green we can merge this

doc/adoc/user_guide.adoc
```diff
@@ -177,14 +177,13 @@ started. However, `kexec` is not supported and does not function in certain
 conditions. For example, the `run_migration` utility does not work in Xen
 based environments or when kernel keys change.
 +
-If kexec causes a kernel panic this can cause the system to hang and the
-distribution migration to fail. In that case refer to this TID:
-https://www.suse.com/support/kb/doc/?id=000019733
-And set the "soft_reboot" customization option:
+By default this method is disabled. Enabling soft reboot behavior via kexec
+requires an explicit configuration change by setting the "soft_reboot"
+customization option:
```
schaefi: I overlooked that this information is for starting the migration. As such you should mention that on the s390x architecture kexec is the only way to start

schaefi (changes requested):
Looks good except that on s390x only kexec based `run_migration` exists

PR #476, opened 2026-04-08
Add time info to backup directory

schaefi:
So far only the date information was part of the backup directory. However, if multiple migration attempts happens on the same day, this would overwrite the data.


cfconrad (approved):
Thanks!

This is a update to https://github.com/SUSE/suse-migration-services/pull/431

rjschwei (approved):

PR #477, opened 2026-04-09
Fix validation of zypper migration result

schaefi:
Up to now the assumption was that any situation in which zypper migration cannot migrate the system returns with an error, meaning exit code != 0. However, this assumption is wrong. There are condition in which zypper migration only indicates a problem with a message saying 'No migration available' and the call return with a successful error code = 0. This causes big trouble for the DMS in a way that it continues running its services which all assumes the migration to the next major release was performed. It misleads readers of the log file into the wrong direction and the worst it causes modifications to the host system when it was not migrated. This commit makes sure that the migration stops and treats the above message as an error condition.

### DIFF
--- a/suse_migration_services/units/migrate.py
+++ b/suse_migration_services/units/migrate.py
@@ -69,7 +69,7 @@ class MigrateSystem:
             os.environ['ZYPP_NO_USRMERGE_PROTECT'] = self.is_single_rpmtrans_requested()
 
             if migration_config.is_zypper_migration_plugin_requested():
-                Zypper.run(
+                zypper_call = Zypper.run(
                     [
                         'migration',
                         verbose_migration,
@@ -87,6 +87,10 @@ class MigrateSystem:
                         self.root_path,
                     ]
                 )
+                if 'No migration available' in zypper_call.output:
+                    raise DistMigrationZypperException(
+                        'The package management system did not receive any upgrade target information from the repository server'
+                    )
             else:
                 zypper_call = Zypper.run(
                     [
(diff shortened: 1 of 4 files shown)


suse_migration_services/units/migrate.py
```diff
@@ -87,6 +87,10 @@ def perform(self):
                         self.root_path,
                     ]
                 )
+                if 'No migration available' in zypper_call.output:
+                    raise DistMigrationZypperException(
+                        'Migration plugin did not find any migration to perform'
```
rjschwei: I would not "hide" the fact that this is really based on the update stack by indicating in the message that this is a migration system issue. I suggest:

"""The package management system did not receive any upgrade target information from the repository server"""

rjschwei (approved):

suse_migration_services/units/migrate.py
schaefi: ok, thanks I will change the message as you suggested

schaefi: done

suse_migration_services/units/migrate.py
```diff
@@ -87,6 +87,10 @@ def perform(self):
                         self.root_path,
                     ]
                 )
+                if 'No migration available' in zypper_call.output:
```
cfconrad, line 90: Just asking: Do we need to call zypper with `LANG=c` to not get in a translation trap? Actually I don't think as we are running on our iso. But maybe I oversee some installation path.

schaefi: good point. The DMS caller environment is a live system without locale support. The default locale is either "C" or en_US.utf-8. In both cases the message is the same. As we control the execution environment via the live system or container it is in my opinion not required to explicitly set the locale for the call.

schaefi:
Hrm, this needs to be done differently as all data gets redirected to the log file...

PR #478, opened 2026-04-13
precheck: XFSv4

YoukouTenhouin:
XFSv4 is no longer supported in SLES. Add a precheck to find XFSv4 filesystems and issue a warning for this.

### DIFF
--- a/suse_migration_services/prechecks/xfs.py
+++ b/suse_migration_services/prechecks/xfs.py
@@ -1,0 +1,56 @@
+# Copyright (c) 2026 SUSE Linux LLC.  All rights reserved.
+#
+# This file is part of suse-migration-services.
+#
+# suse-migration-services is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
+#
+"""Module for checking XFS filesystem version"""
+import logging
+import re
+
+from suse_migration_services.command import Command
+from suse_migration_services.defaults import Defaults
+
+
+def xfs_v4(migration_system=False):
+    """Check for XFS v4 root filesystem and warn about potential issues"""
+    if migration_system:
+        # This check must not be called inside of the
+        # migration system live iso or container image.
+        # It will only be useful on the system to migrate
+        # at install time of the Migration package or on
+        # manual user invocation of suse-migration-pre-checks
+        # prior the actual migration.
+        return
+
+    log = logging.getLogger(Defaults.get_migration_log_name())
+
+    stat_result = Command.run(['stat', '-f', '-c', '%T', '/'])
+    if stat_result:
+        target_fs = stat_result.output.strip()
+        if target_fs != 'xfs':
+            return
+
+    xfs_info_result = Command.run(['xfs_info', '/'])
+
+    # Check for v4 filesystem (absence of crc=1 indicates v4)
+    # v5 filesystems have: crc=1, finobt=1, sparse=1, rmapbt=0, reflink=1
+    # v4 filesystems typically show: crc=0 or no crc field
+    output = xfs_info_result.output
+    if re.search(r'crc=0\b', output) or 'crc=' not in output:
+        log.warning(
+            'XFS v4 root filesystem detected. XFS v4 is deprecated and support '
+            'will be dropped after September 2030. See the XFS formats chapter '
+            'in Storage Administration Guide for details on this issue.'
+        )
--- a/suse_migration_services/prechecks/pre_checks.py
+++ b/suse_migration_services/prechecks/pre_checks.py
@@ -33,6 +33,7 @@ import suse_migration_services.prechecks.scc as check_scc
 import suse_migration_services.prechecks.wicked2nm as check_wicked2nm
 import suse_migration_services.prechecks.cpu_arch as check_cpu_arch
 import suse_migration_services.prechecks.sshd as check_sshd
+import suse_migration_services.prechecks.xfs as check_xfs
 
 
 def main():
@@ -100,6 +101,10 @@ def main():
 
     log.info('--> Checking for encrypted rootfs...')
     check_fs.encryption(migration_system=migration_system_mode)
+    log.info('Done')
+
+    log.info('--> Checking for XFS v4 filesystems...')
+    check_xfs.xfs_v4(migration_system=migration_system_mode)
     log.info('Done')
 
     log.info('--> Checking for latest kernel in multiversion kernel system...')
--- a/test/unit/logger_test.py
+++ b/test/unit/logger_test.py
@@ -1,11 +1,14 @@
 import io
+import logging
 from unittest.mock import patch, MagicMock
 
 from suse_migration_services.logger import Logger
+from suse_migration_services.defaults import Defaults
 
 
 @patch('suse_migration_services.logger.Path.create')
 class TestLogger:
+    @patch.object(logging.getLogger(Defaults.get_migration_log_name()), 'handlers', [])
     def test_setup(self, mock_Path_create):
         with patch('builtins.open', create=True) as mock_open:
             mock_open.return_value = MagicMock(spec=io.IOBase)
@@ -16,6 +19,7 @@ class TestLogger:
                 '/system-root/var/log/distro_migration.log', 'a', encoding='locale', errors=None
             )
 
+    @patch.object(logging.getLogger(Defaults.get_migration_log_name()), 'handlers', [])
     def test_setup_no_system_root(self, mock_Path_create):
         with patch('builtins.open', create=True) as mock_open:
             mock_open.return_value = MagicMock(spec=io.IOBase)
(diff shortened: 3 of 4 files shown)


suse_migration_services/prechecks/xfs.py
```diff
@@ -0,0 +1,71 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# suse-migration-services is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with suse-migration-services. If not, see <http://www.gnu.org/licenses/>
```
schaefi: Please stick with the copyright format that we use in other source files

```
# Copyright (c) 2026 SUSE Linux LLC.  All rights reserved.
#
# This file is part of suse-migration-services.
#
...
#
code
```

Thanks

suse_migration_services/prechecks/xfs.py
```diff
@@ -0,0 +1,71 @@
...
+"""Module for checking XFS filesystem version"""
+import logging
+import re
+
+from suse_migration_services.command import Command
+from suse_migration_services.defaults import Defaults
+from suse_migration_services.fstab import Fstab
+
+
+def xfs_v4(migration_system=False):
```
schaefi, line 26: This is missing the scope check. I'm pretty sure we want to run this only on the host to migrate but not during migration. As such something like this needs to be first

```python3
if migration_system:
        # This check must not be called inside of the
        # migration system live iso or container image.
        # It will only be useful on the system to migrate
        # at install time of the Migration package or on
        # manual user invocation of suse-migration-pre-checks
        # prior the actual migration.
        return
```

suse_migration_services/prechecks/xfs.py
```diff
@@ -0,0 +1,71 @@
...
+    """Check for XFS v4 filesystems and warn about potential issues"""
```
schaefi: This is also missing a target check. I believe this check should only apply for a SLE15 to SLE16 migration. As such something like this is also needed

```python3
target = MigrationTarget.get_migration_target()
if not target.get('version').startswith('16'):
    # This check is only necessary for migration to SLE16
    # ... maybe a bit more information about the why ...
    return
```

suse_migration_services/prechecks/xfs.py
```diff
@@ -0,0 +1,71 @@
...
+
+    v4_filesystems = []
+
+    for fstab_entry in fstab_entries:
+        # Check if the filesystem type is XFS
+        result = Command.run(
+            ["blkid", "-s", "TYPE", "-o", "value", fstab_entry.device],
+            raise_on_error=False
+        )
+        if result.returncode == 0 and 'xfs' in result.output.lower():
```
schaefi: All this should be done without taking the filesystem information from fstab. Here is my suggestion

```python3
stat = Command.run(['stat', '-f', '-c', '%T', '/'])
if stat:
    target_fs = stat.output.strip()
    if target_fs == 'xfs':
        ...
```

suse_migration_services/prechecks/xfs.py
```diff
@@ -0,0 +1,71 @@
...
+                    'crc=' not in output and 'xfs' in output.lower()
+                ):
+                    v4_filesystems.append(fstab_entry.device)
+
+    if v4_filesystems:
+        log.warning(
+            'XFS v4 filesystems detected: {}. XFS v4 is deprecated and may '
+            'cause issues during or after migration. Consider upgrading to '
+            'XFS v5 before migration. See documentation for details on XFS '
+            'filesystem migration.'.format(', '.join(v4_filesystems))
```
schaefi: Is this really something we want to warn about ? It sounds like a severe issue to me when we allow an upgrade in this case ? So we can also think about adding a raise class and raise an error instead of a log.warning. 

```
raise DistMigrationXFSException(
    'The root filesystem is on XFS v4. This is no longer supported on SLE16. An upgrade to XFS v5 prior migration is required. For details see: some-doc-link'
)
```

This will cause suse-migration-pre-checks to fail with a raise message and an error code != 0. This has the consequence that the installation of the package will fail because the %post script which calls suse-migration-pre-checks will fail. In the end the user will not be able to run the migration until this condition is resolved.

To be honest I was not aware that SLE16 no longer supports XFS v4 and it doesn't sound like a smart move to be honest. All our public cloud images are XFS based and I did not spot a problem when testing migrations.

Anyway please let us clarify how severe this is. Thanks

schaefi (changes requested):
Thanks for your effort, please see my comments

suse_migration_services/prechecks/xfs.py
YoukouTenhouin: XFSv4 is deprecated at least since 12SP5, so I think migration from SLE12 to SLE15 should also benefit from this check? See https://documentation.suse.com/sles/12-SP5/html/SLES-all/cha-filesystems.html#sec-filesystems-major-xfs

suse_migration_services/prechecks/xfs.py
YoukouTenhouin: I'll discuss the severity of this; although looking at upstream documentation (https://docs.kernel.org/admin-guide/xfs.html) the v4 support in kernel will remain there until September 2030, so I guess officially it is still "supported" for now, so maybe we can issue a warning for now and only block the migration later?

schaefi: That's fine for me. So we know the SLE16 kernel will support it. As such I recommend to add the date when v4 will deprecate to the warning message but don't concern the user with `may cause issues during or after migration` which imho sounds vague and concerning :smile:

suse_migration_services/prechecks/xfs.py
```diff
@@ -0,0 +1,57 @@
...
+    # Check for v4 filesystem (absence of crc=1 indicates v4)
+    # v5 filesystems have: crc=1, finobt=1, sparse=1, rmapbt=0, reflink=1
+    # v4 filesystems typically show: crc=0 or no crc field
+    output = xfs_info_result.output
+    if re.search(r'crc=0\b', output) or 'crc=' not in output:
+        log.warning(
+            'XFS v4 root filesystem detected. XFS v4 is deprecated and support'
+            'will be dropped after September 2030. Consider migrating to '
+            'XFS v5 before migration. See documentation for details on XFS '
+            'filesystem migration.'
```
rjschwei: I think we should provide better information than what amounts to "go fishing and find some documentation, good luck". I suggest

"""
....See the XFS section of the SUSE Storage Administration Guide for details about XFS version migration.
"""

suse_migration_services/prechecks/xfs.py
```diff
@@ -0,0 +1,57 @@
...
+    xfs_info_result = Command.run(['xfs_info', '/'])
+
+    # Check for v4 filesystem (absence of crc=1 indicates v4)
+    # v5 filesystems have: crc=1, finobt=1, sparse=1, rmapbt=0, reflink=1
+    # v4 filesystems typically show: crc=0 or no crc field
+    output = xfs_info_result.output
+    if re.search(r'crc=0\b', output) or 'crc=' not in output:
+        log.warning(
+            'XFS v4 root filesystem detected. XFS v4 is deprecated and support'
+            'will be dropped after September 2030. Consider migrating to '
```
rjschwei: Hmm, per documentation [1] in 12 SP5 it does not look like a migration between v4 and v5 is possible without starting over. This is confirmed by a Google search

"""
...It is not possible to perform an in-place upgrade from XFS v4 to v5 on Linux. ...
"""
Since we are taking about the root filesystem here and not some data file system this is a bit more complicated, especially in the cloud. For someone to have a v4 setup that would mean they started with SLE 11, then migrated to 12 and now migrate to 15 or 16. In the cloud our SLE 11 images had ext3 and as such we should be safe since in SLE 12, when we started using XFS as the root file system we already used v5.

Anyway, long story short, since we are talking about the root filesystem and in place migration of XFS from v4 to v5 is not possible we should probably not call this a migration, it is more of a start over scenario.


[1] https://documentation.suse.com/sles/12-SP5/html/SLES-all/cha-filesystems.html#sec-filesystems-major-xfs

rjschwei (changes requested):

suse_migration_services/prechecks/xfs.py
schaefi: ok, thanks for clarifying

suse_migration_services/prechecks/xfs.py
schaefi: I agree, if it's not possible to run an in-place upgrade the check must fail with a raise condition and we should not allow to run the migration

> when we started using XFS as the root file system we already used v5.

That's good to know

YoukouTenhouin: Reworded the message a bit. XFSv4 is being deprecated due to 2038 issue, so we still have quite some time before it actually blow up. So I still consider it a bit extreme to block the migration due to this issue, and a head up should be enough.

suse_migration_services/prechecks/xfs.py
YoukouTenhouin: Reworded. Thanks for the suggestion!

suse_migration_services/prechecks/xfs.py
YoukouTenhouin: Reworded. Thanks for the suggestion!

schaefi:
@YoukouTenhouin Thanks for applying the changes. Can you now please rebase to master and after that call

```
make black
```

We have added the python black coding style for consistency reasons. After that it looks good to me

btw: I'm in the middle of a release process. So this change will only go to devel:DMS and will be subject to the next release round. I guess that's ok since it looks like a nice new check but not super urgent.

Would you agree ?

Thanks

YoukouTenhouin:
@schaefi Rebased and formatted. It's not urgent at all so we can perfectly wait until next round. Thanks for reviewing!

schaefi (approved):

PR #481, opened 2026-04-21
Use download-in-advance mode for zypper migration

schaefi:
This reduces the risk of download problems while inside of the upgrade process. This Fixes jira#PCT-1170


cfconrad (approved):
LGTM

PR #482, opened 2026-04-27
Fixed log message for already mounted system

### LINKED ISSUE #479: Bad logic in perform() ?
https://github.com/SUSE/suse-migration-services/blob/1a7e2644c1a83e0563997eff946b084e47846299/suse_migration_services/units/mount_system.py#L64

If it is not mounted what's the reason to log the it is mounted?

schaefi:
The log message was out of context and belongs to the block checking if the system was already mounted.
This Fixes #479


jesusbv (approved):
good catch

PR #483, opened 2026-04-27
Unified resolv.conf handling

cfconrad:
1) **New ResolvConf Management Class**
  * A dedicated class `suse_migration_services/resolv_conf.py` has been implemented to handle name resolution configuration. The main idea is to get a consistent resolv.conf handling during migration and in the target system.
This class should be used for migrations from sle12->sle15 and sle15->sle16.
  * When we moved from `sysconfig-netconfig` to `NetworkManager` it is nesessary to create a symlink (resolv.conf -> resolv.conf.static), to prevent NetworkManager of overwriting our custom resolv.conf.
  * To detect custom resolv.conf files,  the lagacy md5 checksum and `# Generated by NetworkManager` strings are considered. An empty resolv.conf file, is considered to be valid and custom. A dangling link will *not* be considered as custom.

2. **Workflow Integration**
  * Live System Setup: SetupNameResolver now utilizes ResolvConf.setup_live_system() to correctly apply name resolution to the migration host.
  * Target Root Setup: WickedToNetworkManager uses ResolvConf.setup_target_root() during the transition to ensure the migrated system has a valid resolver config and cleanup old links to `netconfig/resolv.conf`.

3. **Systemd Service Reordering**
  * To ensure connectivity is established correctly, the dependency order has been updated: `suse-migration-setup-name-resolver.service` now runs before `suse-migration-setup-host-network.service` and `suse-migration-prepare.service`.

4. **Example**
  * `/etc/resolv.conf` regular file. It was copied to live system and kept in the target system. It was prone to be overwritten by NetworkManger.
  * Now a "static link" will be created
    * `/etc/resolv.conf.static` (the original custom file)
    * `/etc/resolv.conf` -> `resolv.conf.static` (symlink)
    * Note: If `resolv.conf.static` already exists, a timestamp is appended (e.g., `resolv.conf.static-2026-04-29-1200`)

5. **netconfig workaround**
  * The current available `sysconfig-netconfig` version contains an issue on uninstalling ([bsc#1263889](https://bugzilla.suse.com/show_bug.cgi?id=1263889)). Once the fix is published the following workaround  baee54d is not needed anymore.

6. **resolv.conf** inside of chroot (`/system-root`)
  * The call `chroot $root_path SUSEConnect --list-extensions` in `suse_connect.py` need a working `resolv.conf` also within the chroot environment. 
  * this was introduces with https://github.com/SUSE/suse-migration-services/pull/311 and fixed in https://github.com/SUSE/suse-migration-services/pull/389 and https://github.com/SUSE/suse-migration-services/pull/442 . But this still lead to unexpected links in the chroot, which makes decisions in ResolvConf::setup_target_root() misleading. 
  * Do always mount /system-root/run as `tmpfs` [83e6f17](https://github.com/SUSE/suse-migration-services/pull/483/commits/83e6f17d78c4e725bc2f30393d9b88004e57612b)
  * Do only mount /etc/resolv.conf to /system-root/run/{NetworkManager,netconfig}/resolv.conf

7. **known issues**
  * The live iso do not contain local DNS resolver like `dnsmasq`, `systemd-resolvd` or `bind`. Thus if a corresponding configuration in `NETCONFIG_DNS_FORWARDER` or a `nameserver 127.0.0.1` will not work and should be replaced temporally.


suse_migration_services/resolv_conf.py
```diff
@@ -0,0 +1,167 @@
...
+                if self.is_custom_resolv_conf(resolv_realpath):
+                    self.copy(resolv_realpath, resolv_conf_dest)
+                    self.create_static_resolv()
+
+        else:
+            if self.is_custom_resolv_conf(resolv_conf):
+                self.copy(resolv_conf, resolv_conf_dest)
+                self.create_static_resolv()
+
+    def copy(self, src, dst):
```
cfconrad: Consider `cp --force`

suse_migration_services/resolv_conf.py
```diff
@@ -0,0 +1,167 @@
...
+
+        with open(file, 'r') as resolv:
+            for line in resolv:
+                # check if it contains "Generated by NetworkManager"
+                if line.startswith('# Generated by NetworkManager'):
+                    return False
+
+                # check if there are nameservers configured
+                elif line.lstrip().startswith('nameserver'):
+                    return True
```
cfconrad: No special handling for empty `/etc/resolv.conf`

suse_migration_services/resolv_conf.py
```diff
@@ -0,0 +1,167 @@
...
+    def is_custom_resolv_conf(self, file):
+        if not os.path.exists(file):
+            return False
+
+        # On sle12 netconfig check via md5sum if the resolv.conf is
+        # customized.
+        netconfig_md5, erx = self.read_netconfig_md5()
+        if netconfig_md5:
+            if netconfig_md5 == self.create_netconfig_md5(file, erx):
+                return False
```
cfconrad, line 140: Create a wrapper  function

suse_migration_services/resolv_conf.py
```diff
@@ -0,0 +1,167 @@
...
+                # check if it contains "Generated by NetworkManager"
+                if line.startswith('# Generated by NetworkManager'):
+                    return False
+
+                # check if there are nameservers configured
+                elif line.lstrip().startswith('nameserver'):
+                    return True
+        return False
+
+    def resolve_chroot_symlink(self, symlink_path, new_root):
```
cfconrad: something like: `Command.run('chroot $system-root test -f /etc/resolv.conf')`

cfconrad: The equivalent to the function is `chroot $system-root readlink -m src/etc/resolv.conf`

suse_migration_services/resolv_conf.py
cfconrad: it is `cp --remove-destination` as `cp --force` do not overwrite dangling symlinks.

suse_migration_services/resolv_conf.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    **Managing /etc/resolv.conf **
+    """
+
+    def __init__(self, target_root, src_dir, dest_dir=None):
+        self.target_root = target_root
+        self.src_dir = src_dir
+        self.dest_dir = dest_dir if dest_dir else src_dir
+
+    def prepare_resolv_conf(self):
```
jcronenberg: Personally I dislike this design of one magic function. And also I don't understand why it's named `prepare` if it does everything? To me `prepare` sounds like doing things necessary beforehand and then later doing the actual necessary steps.
But more importantly IMO it would be way more readable if it would be split into 2 functions. One that handles the live image and one that handles the target system. It may not perfectly follow DRY and be a bit more verbose, but the readability/'understandability' would be worth it I'd say.

cfconrad: I gave it a try https://github.com/SUSE/suse-migration-services/pull/483/changes/75570c506c09582744303e1977238c44d815c107 and I think you are right! I would be also for this way.

suse_migration_services/resolv_conf.py
cfconrad: Didn't do so. Only renamed `create_netconfig_md5()` into `generate_netconfig_md5()`

suse_migration_services/resolv_conf.py
cfconrad: done, but I'm not sure if the usage of `chroot` is rely a benefit. E.g. the integration tests do not work without mocking this part.

suse_migration_services/resolv_conf.py
```diff
@@ -0,0 +1,170 @@
...
+                if not re.match(r'^#|^\s*$', line):
+                    filtered_content += line.rstrip('\n') + '\n'
+        return hashlib.md5(filtered_content.encode()).hexdigest()
+
+    def is_custom_resolv_conf(self, file):
+        if not os.path.exists(file):
+            return False
+
+        # On sle12, netconfig checks via md5sum if the resolv.conf is
+        # customized.
```
schaefi: Can we please drop the SLE12 reference in this comment. Actually netconfig has this checksum feature and we make use of this information to look for custom modifications. Let's not connect this to some distro. Thanks

schaefi (approved):
Looks great 👍 Thanks much for your effort

Just nitpick on a comment

suse_migration_services/resolv_conf.py
schaefi: It has one benefit that this action will be part of the log file. As such it doesn't hurt to do it that way

cfconrad:
Ready from my side

PR #485, opened 2026-04-30
Avoid temporary file `test/unit/.coverage` in dist archive

cfconrad:
This temporary file(`test/unit/.coverage`), generated by pytest-cov, is created within each `make test` call containing different content, also if no source files has changed.

I didn't found any files which match `.coverage*` beside this temporary file. And for me, it makes no sense to distribute it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ packages = [
 
 include = [
    { path = ".bumpversion.cfg", format = "sdist" },
-   { path = ".coverage*", format = "sdist" },
    { path = "setup.cfg", format = "sdist" },
    { path = "grub.d", format = "sdist" },
    { path = "systemd", format = "sdist" },
@@ -31,7 +30,8 @@ include = [
    { path = "package", format = "sdist" },
    { path = "setup.py", format = "sdist" },
    { path = "MANIFEST.in", format = "sdist" },
-   { path = "test", format = "sdist" },
+   { path = "test/unit/**/*.py", format = "sdist" },
+   { path = "test/data", format = "sdist" },
 ]
 
 classifiers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ packages = [
 
 include = [
    { path = ".bumpversion.cfg", format = "sdist" },
-   { path = ".coverage*", format = "sdist" },
    { path = "setup.cfg", format = "sdist" },
    { path = "grub.d", format = "sdist" },
    { path = "systemd", format = "sdist" },
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ version := $(shell \
 )
 
 tar: clean check test
+	# remove eventually existing test artifacts
+	rm -f test/unit/.coverage
 	# build the sdist source tarball
 	poetry build --format=sdist
 
@@ -146,8 +148,7 @@ test: setup
 		--cov=suse_migration_services \
 		--no-cov-on-fail \
 		--cov-report=term-missing \
-		--cov-fail-under=100 \
-		--cov-config .coveragerc'
+		--cov-fail-under=100'
 
 black: setup
 	poetry run black --skip-string-normalization --line-length 100 suse_migration_services test/unit/


pyproject.toml
```diff
@@ -31,7 +30,8 @@ include = [
    { path = "package", format = "sdist" },
    { path = "setup.py", format = "sdist" },
    { path = "MANIFEST.in", format = "sdist" },
-   { path = "test", format = "sdist" },
+   { path = "test/**/*.py", format = "sdist" },
+   { path = "test/data/**/*", format = "sdist" },
```
schaefi: I'm fine removing the .coverage temporary data. It was placed to the dist target to keep the last successful coverage meta in the dist target, but I have no strong opinion about it.

However, the other change not related to .coverage is wrong. The manifest recursively includes all test data by the simple specification as it existed. You can check yourself by calling "make build" and run "tar -tf dist/suse-migration-services.tar.gz | grep test", it's all there

As such I do not understand and also disagree with this change. It's also not related to the description of this PR. As such please revert it and only keep the drop of the .coverage line

Thanks

schaefi (changes requested):

pyproject.toml
cfconrad: The problem is this:
```
> make build >/dev/null && tar -tf dist/suse-migration-services.tar.gz | grep coverage
suse_migration_services-2.1.34/test/unit/.coverage
```
the the coverage file is in `test/` and if we include the full test, this file is in there as well.
What about the following, I think it will read better

```
  { path = "test/unit/**/*.py", format = "sdist" },
  { path = "test/data", format = "sdist" },
```

I'm checking with:
```sh
make build >/dev/null && tar -tf dist/suse-migration-services.tar.gz | sort > after
git checkout master
make build >/dev/null && tar -tf dist/suse-migration-services.tar.gz | sort > before
diff before after
```
```
154d153
< suse_migration_services-2.1.34/test/unit/.coverage
```

schaefi: > the the coverage file is in test/ and if we include the full test, this file is in there as well.

and what problem do you see here ? What problem are you trying to solve ?

To be really honest I have zero issues with the sdist target as it exists today and as we deliver it since quite some time. Messing with the contents of the sdist target is source for trouble and we already have enough issues to fix.

So if you can explain to me what real problem this change will solve I'm happy to apply the change. But if this falls into the category cosmetic change I'm not willing to apply it.

`./test/unit/.coverage` is an artifact of a pytest run and yes it is pointless to have it in the sdist tarball. Instead of a new rule in the pyproject include I suggest to fix this differently. The original intend for the .coverage* rule was to include the coveragerc file. In this project we don't have any. As such I'm proposing the following patch

```patch
diff --git a/Makefile b/Makefile
index 4154597..351caae 100644
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ version := $(shell \
 )
 
 tar: clean check test
+       # drop eventually existing test artifacts
+       rm -f test/unit/.coverage
        # build the sdist source tarball
        poetry build --format=sdist
 
@@ -146,8 +148,7 @@ test: setup
                --cov=suse_migration_services \
                --no-cov-on-fail \
                --cov-report=term-missing \
-               --cov-fail-under=100 \
-               --cov-config .coveragerc'
+               --cov-fail-under=100'
 
 black: setup
        poetry run black --skip-string-normalization --line-length 100 suse_migration_services test/unit/
diff --git a/pyproject.toml b/pyproject.toml
index 18d4706..1cbb46d 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ packages = [
 
 include = [
    { path = ".bumpversion.cfg", format = "sdist" },
-   { path = ".coverage*", format = "sdist" },
    { path = "setup.cfg", format = "sdist" },
    { path = "grub.d", format = "sdist" },
    { path = "systemd", format = "sdist" },

```

Thoughts ?

cfconrad: It is actually a nitty thing, but  for testing I use 
```
make build
cp dist/* ~/my_obs_devel_branch/suse-migration-services

cd ~/my_obs_devel_branch/suse-migration-services
obs ci
...
```
And this triggers always a rebuild, even if nothing changed.

cfconrad: I updated to your suggestion.

schaefi (approved):

PR #486, opened 2026-05-06
Make sure to add multipath to the initrd

schaefi:
Add multipath dracut code to the initrd. This is related to bsc#1260340

### DIFF
--- a/image/generic/sle16/config.kiwi
+++ b/image/generic/sle16/config.kiwi
@@ -57,6 +57,7 @@
         <package name="grub2-powerpc-ieee1275" arch="ppc64le"/>
         <package name="grub2-s390x-emu" arch="s390x" />
         <package name="lvm2"/>
+        <package name="multipath-tools"/>
         <package name="tar"/>
         <package name="parted"/>
         <package name="openssh"/>
--- a/image/generic/sle16/root/etc/dracut.conf.d/20-multipath.conf
+++ b/image/generic/sle16/root/etc/dracut.conf.d/20-multipath.conf
@@ -1,0 +1,1 @@
+add_dracutmodules+=" multipath "


rjschwei (approved):

rjschwei:
This might create problems for Public Cloud instances. For Public Cloud images we have `multipath=off` on the kernel command line. For the upgrade path we had left `multipath` out of the initrd so we would not have to set `multipath=off` when we boot into the live-iso.

Now that the migration mechanism is not just for the Public Cloud anymore we need to support multipath, that is understood. It is also not practical to test on every instance type in every CSP, that would be > 600 tests.

I do not remember the specific details, i.e. which instance types, triggered issues with boot when multipath was on. We need to be prepared that a follow up change may be required. Possibly in the activation package.

schaefi:
> I do not remember the specific details, i.e. which instance types, triggered issues with boot when multipath was on.

I share the same concern. It's been a long time when the simple presence of multipath in the initrd caused issues with other device nodes and I also cannot remember the exact details. I will run the basic tests to see if it is still an issue

schaefi:
@rjschwei Do you remember why `multipath=off` is explcitly passed for our cloud images ? It's obvious that it's not needed but do you remember an issue that was connected with the setting ?

rjschwei:
> @rjschwei Do you remember why `multipath=off` is explcitly passed for our cloud images ? It's obvious that it's not needed but do you remember an issue that was connected with the setting ?

I do not remember the details, all I remember is that we had boot issues in certain conditions.

schaefi:
I will zip a devel:DMS build including this change but I will keep the PR open until we know that things do not break in public clouds. If we see boot issues due to this change we have to add code in `grub.d/99_migration` as well as in `tools/run_migration` to check if the host runs with root on multipath and if not set `multipath=off` to the cmdline.

As of today I haven't investigated the best way to check for multipath usage. Right now I would just check for a running `multipathd` but that's not the best way to do it

schaefi:
@rjschwei I tested with multipath as part of the initrd and we have no issue on clouds. Actually all settings that we use in cloud instances regarding the kernel are inherited by the DMS when the live migration system boots up. There is this code in grub.d/99_migration

```
# add relevant boot options used on the host to migrate
for host_boot_option in $(xargs -n1 < /proc/cmdline);do
  ...
```

and it takes over what is already used. For SLE15 based instances in the cloud we have this cmdline in the live migration as an example 

```
migration@ip-172-31-10-215:~> cat /proc/cmdline 
BOOT_IMAGE=(loop)/boot/x86_64/loader/linux iso-scan/filename=/migration-image/SLES16-Migration.x86_64-2.1.34-Build9.20.iso rd.live.image root=live:CDLABEL=CDROM migration_target=99c9cb1b-db9d-4fd4-b8cb-991e90839163 BOOT_IMAGE=/boot/vmlinuz-6.4.0-150700.53.40-default console=ttyS0 net.ifnames=0 iommu.passthrough=1 8250.nr_uarts=4 dis_ucode_ldr multipath=off nvme_core.admin_timeout=4294967295 nvme_core.io_timeout=4294967295 systemd.unified_cgroup_hierarchy=1
```

and consequently

```
migration@ip-172-31-10-215:~> systemctl status multipathd
○ multipathd.service - Device-Mapper Multipath Device Controller
     Loaded: loaded (/usr/lib/systemd/system/multipathd.service; disabled; preset: disabled)
     Active: inactive (dead)
```

As you can see `multipath=off` stays effective

On a system that has an active multipath setup we would inherit all relevant settings too and I expect this to work in the live migration as well, but this part needs testing too.

rjschwei:
@schaefi great, thanks for following up on this.

PR #487, opened 2026-05-07
Fix loading SELinux policy of Salt Bundle with workaround

vzhestkov:
Salt Bundle (`venv-salt-minion` package) is shipping its own SELinux policy with the single package.
It loads the SELinux policy only in case if SELinux is enabled, but with the migration process we have the situation when the packages are updating first and then the migration from AppArmor to SELinux is triggered what is causing the situation when Salt Bundle is installed, but the SELinux policy is not loaded.

This change is introducing a workaround with AppArmor migration and just triggering force reinstall of `venv-salt-minion` in case if it is installed on the system.

Tracks: https://github.com/SUSE/suse-migration-services/pull/487


schaefi (approved):
Looks good to me 👍

PR #488, opened 2026-05-12
Fix install of NetworkManager

schaefi:
Check if NetworkManager is installed already and only install it when needed. In addition run the installation of packages inside of the upgraded system as a chroot operation. We identified zypper metadata read issues when driving a package install from the outside (live migration system) into the target and we also want this operation to be as close as possible to a standard package installation on the target system.

### DIFF
--- a/suse_migration_services/units/wicked_migration.py
+++ b/suse_migration_services/units/wicked_migration.py
@@ -59,9 +59,14 @@ class WickedToNetworkManager(DropComponents):
             return
         try:
             self.log.info('Ensuring NetworkManager is installed in migrated system')
-            Zypper.install(
-                'NetworkManager', 'NetworkManager-config-server', system_root=self.root_path
-            )
+            if not self.package_installed('NetworkManager'):
+                Zypper.install(
+                    'NetworkManager',
+                    'NetworkManager-config-server',
+                    extra_args=['--no-recommends'],
+                    raise_on_error=True,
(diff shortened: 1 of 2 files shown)


suse_migration_services/units/wicked_migration.py
```diff
@@ -59,9 +59,14 @@ def perform(self):
...
         try:
             self.log.info('Ensuring NetworkManager is installed in migrated system')
-            Zypper.install(
-                'NetworkManager', 'NetworkManager-config-server', system_root=self.root_path
-            )
+            if not self.package_installed('NetworkManager'):
+                Zypper.install(
+                    'NetworkManager',
+                    'NetworkManager-config-server',
+                    extra_args=['--no-recommends'],
```
cfconrad, line 66: Actually this `--no-recommends` ring a bell on my side, what happen if we migrate ovs-bridges and do we get `NetworkManager-ovs` installed.... But this is a other storry. :-) I will take care in an other PR.

cfconrad (approved):
Sounds like a workaround for `zypper in --root xxx`.

LGTM, thx!

rjschwei (approved):

suse_migration_services/units/wicked_migration.py
schaefi: ok, let's handle this in another PR. Thanks

PR #489, opened 2026-05-18
Add precheck for POWER architecture version on SLES16 migration

YoukouTenhouin:
SLES 16 only supports POWER10 (and probably above), so we need to add a precheck to notify user if they are trying to migrate on an older system. We don't need to block the migration as brutal as the x86_64_v1 check, as this does not cause a kernel panic (yet).

### DIFF
--- a/suse_migration_services/prechecks/cpu_arch.py
+++ b/suse_migration_services/prechecks/cpu_arch.py
@@ -19,21 +19,25 @@
 """Module for checking for CPU architecture support"""
 import logging
 import os
+import platform
+import re
 
 from suse_migration_services.defaults import Defaults
 from suse_migration_services.command import Command
 from suse_migration_services.migration_target import MigrationTarget
 
+log = logging.getLogger(Defaults.get_migration_log_name())
+
 
 def x86_64_version():
     """Function to check whether current CPU architecture is new enough to support SLE16"""
     target = MigrationTarget.get_migration_target()
+    version = target.get('version')
 
-    if target.get('version') != '16.0':
+    if not version or not version.startswith('16'):
         # This check is only necessary for migration to SLE16
         return
 
-    log = logging.getLogger(Defaults.get_migration_log_name())
     os.environ['ZYPP_READONLY_HACK'] = '1'
     zypper_call = Command.run(['zypper', 'system-architecture'])
     arch = zypper_call.output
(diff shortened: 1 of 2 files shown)


schaefi:
@YoukouTenhouin I do not understand the pytest and coverage artifacts, but the additional cpu check looks good to me. We are also missing a description of this PR. Can you take a look into this parts and then feel free to open for review

suse_migration_services/prechecks/cpu_arch.py
```diff
@@ -45,6 +49,43 @@ def x86_64_version():
...
+    version = target.get('version')
+
+    if not version or not version.startswith('16'):
+        # This check is only necessary for migration to SLE16
+        return
+
+    machine = platform.machine()
+
+    if machine not in ['ppc64', 'ppc64le']:
+        return
```
schaefi, line 64: nitpick, you could consider to reverse the logic

```python3
if machine in ['ppc64', 'ppc64le']:
    try:
       ...
```

it saves you some test code and it's imho more straight forward to read. But I leave this up to you

suse_migration_services/prechecks/cpu_arch.py
```diff
@@ -45,6 +49,43 @@ def x86_64_version():
...
+
+    try:
+        with open('/proc/cpuinfo', 'r') as cpuinfo_file:
```
schaefi, line 67: Have you considered 

```python3
import platform
platform.processor()
```

if it does provide the information you need, you can code this shorter

schaefi:
@YoukouTenhouin any news on this one ?

YoukouTenhouin:
@schaefi We are waiting for the test result on PPC64 to confirm whether there's any issue; I will update the PR once I got the confirmation. Thanks for reviewing!

suse_migration_services/prechecks/cpu_arch.py
YoukouTenhouin: Thanks for the suggestion! Although I think if we write it like this then we will wrap almost the entire function body into an `if` branch, so I think the current structure is actually cleaner.

suse_migration_services/prechecks/cpu_arch.py
YoukouTenhouin: Thanks for the suggestion. I investigated the result on a PPC system and this only returns 'ppc64le', so we probably still need to check `/proc/cpuinfo`. But thanks for letting me know that this exists!

YoukouTenhouin:
Tested on both POWER9 and POWER10 systems and precheck seems to be working. Marking this PR as ready for review.

suse_migration_services/prechecks/cpu_arch.py
schaefi: ok

schaefi (approved):
Looks good 👍 Thanks much

schaefi:
@YoukouTenhouin I'll merge this and zip a new devel build

PR #492, opened 2026-06-12
Use multipath device instead of block devices system with multipath

greygoo:
When attempting migration on a mutlipath enabled system. it is not possible to directly mount the block devices used in the multipath setup, but instead the multipath devices need to be mounted. Since the lsblk call used to determine the root device name returns the block device name before the multipath device name, currently the block device is attempted to be mounted, which fails,
This change introduces a is_multipath() function, that returns true if the system is using multipath devices and will skip the use of the block devices so that the multipath device is used instead.

### DIFF
--- a/suse_migration_services/units/mount_system.py
+++ b/suse_migration_services/units/mount_system.py
@@ -129,7 +129,10 @@ class MountSystem:
                             device = block_record[0]
                             uuid = self.get_uuid(device)
                             if uuid == migration_rootfs_uuid:
-                                return device
+                                if self.is_multipath() and not device.startswith('/dev/mapper/'):
+                                    self.log.info('Skipping {0} since system is multipath'.format(device))
+                                else:
+                                    return device
             # nothing was found
             raise DistMigrationSystemMountException('no match for migration_target= in cmdline')
         except Exception as issue:
@@ -205,6 +208,14 @@ class MountSystem:
         # that case is not handled
         return os.path.ismount(mount_point)
 
+    def is_multipath(self):
+        lsblk_call = Command.run(['lsblk', '-p', '-n', '-r', '-o', 'NAME,TYPE'])
+        for entry in lsblk_call.output.split(os.linesep):
+            block_record = entry.split()
+            if len(block_record) >= 2:
+                block_type = block_record[1]
+                if block_type == 'mpath':
+                    return True
 
 def main():
     mount_os = MountSystem()


suse_migration_services/units/mount_system.py
```diff
@@ -129,7 +129,10 @@ def get_target_root(self):
                             device = block_record[0]
                             uuid = self.get_uuid(device)
                             if uuid == migration_rootfs_uuid:
-                                return device
+                                if self.is_multipath() and not device.startswith('/dev/mapper/'):
```
schaefi, line 132: This is not a safe match. A device mapped to /dev/mapper must not necessarily be a multipath device. I think it would be better to change:

```python3
considered_block_types = ['part', 'raid', 'lvm']
```

to

```python3
considered_block_types = ['part', 'raid', 'lvm', 'mpath']
```

next you can change the condition to

```python3
if self.is_multipath() and block_type != 'mpath':
    # we are multipath based, only match with mpath devices
    continue

device = block_record[0]
uuid = self.get_uuid(device)
if uuid == migration_rootfs_uuid:
    return device
```

Thoughts ?

suse_migration_services/units/mount_system.py
```diff
@@ -205,6 +208,14 @@ def is_mounted(self, mount_point):
...
         return os.path.ismount(mount_point)
 
+    def is_multipath(self):
+        lsblk_call = Command.run(['lsblk', '-p', '-n', '-r', '-o', 'NAME,TYPE'])
+        for entry in lsblk_call.output.split(os.linesep):
+            block_record = entry.split()
+            if len(block_record) >= 2:
+                block_type = block_record[1]
+                if block_type == 'mpath':
+                    return True
```
schaefi, line 218: That I believe is not safe. Just because there is one device a multipath device doesn't necessarily mean that the system we want to migrate is multipath based.

I would solve this by passing a kernel cmdline parameter indicating that the system to migrate is multipath based. The place to add this is in:
* grub.d/99_migration
* tools/run_migration

schaefi (changes requested):
A good start, see my comments

suse_migration_services/units/mount_system.py
schaefi: So you need to detect the multipath condition on the host prior the migration and pass a flag via the cmdline such that the DMS live system later knows about it.

schaefi: same concept as with `migration_target`

suse_migration_services/units/mount_system.py
greygoo: That's not gonna work, since the device with type mpath is not the root partition:

```
localhost:~ # lsblk -p -n -r -o NAME,TYPE,ID,UUID
/dev/loop0 loop  2026-06-11-19-31-30-00
/dev/loop1 loop  
/dev/loop2 loop  eb61a85a-52c5-4723-ad9e-d85f73e18f54
/dev/sda disk SIBM_2145_00e0208092f5XX00 
/dev/sda1 part SIBM_2145_00e0208092f5XX00-part1 
/dev/sda2 part SIBM_2145_00e0208092f5XX00-part2 24438359-39ae-4046-806e-fdef0c3d4ec5
/dev/sda3 part SIBM_2145_00e0208092f5XX00-part3 d1a1151b-54e5-4084-9eac-931dbeb21c49
/dev/sda4 part SIBM_2145_00e0208092f5XX00-part4 2d98e492-a44a-4267-8f84-2630cc8386f9
/dev/mapper/3600507638082024bd40000000000002d mpath 3600507638082024bd40000000000002d 
/dev/mapper/3600507638082024bd40000000000002d-part1 part 3600507638082024bd40000000000002d-part1 
/dev/mapper/3600507638082024bd40000000000002d-part2 part 3600507638082024bd40000000000002d-part2 24438359-39ae-4046-806e-fdef0c3d4ec5
/dev/mapper/3600507638082024bd40000000000002d-part3 part 3600507638082024bd40000000000002d-part3 d1a1151b-54e5-4084-9eac-931dbeb21c49
/dev/mapper/3600507638082024bd40000000000002d-part4 part 3600507638082024bd40000000000002d-part4 2d98e492-a44a-4267-8f84-2630cc8386f9
```

The multipath partition still is type partition. I'll change the check, if the found device name does not include the multipath wwid, it will skip.

schaefi: ok fair point, but the wwn id is unique to the root disk and its partitions. So it still would be better to base the condition on the wwn and not just the start of /dev/mapper 

```
/dev/mapper/3600507638082024bd40000000000002d mpath 3600507638082024bd40000000000002d 
```

wwn = 3600507638082024bd40000000000002d

all devices on this disk are

/dev/.../$wwn-partN

suse_migration_services/units/mount_system.py
schaefi: @greygoo you could also consider to retrieve the `wwn` id and pass this as an information for the cmdline 

```
migration_target_wwn=3600507638082024bd40000000000002d
```

Then in the code change that you did above check for `migration_target_wwn` and if it is present only take devices into account with that wwn

suse_migration_services/units/mount_system.py
greygoo: If migration-target is multipathed, then lsblk will show more than one device for the migration-target uuid.

- I'll check that on the host prior to starting the migrtion, and if this condition is true, I'll pass the mulitpath wwid via kernel command line as multipath_wwid=3600507638082024bd40000000000002d
- In the migration system, I'll check if a wwid is passed via command line, if yes, then migration_root_device is only returned if its name contains the wwid.

suse_migration_services/units/mount_system.py
greygoo: yes, see suggestion in other thread, I think the safest approach is to use the root partition uuid for migration_target and only add an additional migration_target_wwid parameter that only get passed if the target is multipathed

suse_migration_services/units/mount_system.py
schaefi: yep sounds perfect :+1:

greygoo:
This PR is replaced by 

https://github.com/SUSE/suse-migration-services/pull/493

PR #493, opened 2026-06-15
Handle multipath root

greygoo:
This PR adds handling multipath systems by determining the multipath wwn of the migration target and passing it as kernel command line parameter:

- In the host system to be maigrated, check if a the multipath wwn is used for the root device.
- Pass the miltipath wnn as multipath_wwn kernel parameter
-  In the migration system pick up the multipath_wwn
- Use the device name containing the multipath wwn instead of the block device with the same uuid

### DIFF
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -29,8 +29,10 @@ if [[ ${image_fs_type} =~ ^ext[[:digit:]] ]] ; then
    image_fs_type="ext2"
 fi
 
-if type -p multipath; then
-    migration_rootfs_device=$(lsblk -p -n -r -o NAME,MOUNTPOINTS | grep -E "/$" | uniq | cut -f1 -d" ")
+if type -p multipath &>/dev/null; then
+    migration_rootfs_device=$(
+        lsblk -p -n -r -o NAME,MOUNTPOINTS | grep -E "/$" | uniq | cut -f1 -d" "
+    )
     for wwn in $(multipath -l -v1 2>/dev/null); do
         if [[ "${migration_rootfs_device}" == *"${wwn}"* ]];then
             migration_rootfs_wwn="${wwn}"
--- a/grub.d/99_migration
+++ b/grub.d/99_migration
@@ -32,9 +32,9 @@ fi
 if [ -e /sbin/multipath ]; then
     migration_rootfs_device=$(lsblk -p -n -r -o NAME,MOUNTPOINTS | grep -E "/$" | uniq | cut -f1 -d" ")
     for wwn in $(/sbin/multipath -l -v1 2>/dev/null); do
-	if [[ "${migration_rootfs_device}" == *"${wwn}"* ]];then
-	    migration_rootfs_wwn="${wwn}"
-	fi
+        if [[ "${migration_rootfs_device}" == *"${wwn}"* ]];then
+            migration_rootfs_wwn="${wwn}"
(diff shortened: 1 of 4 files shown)


grub.d/99_migration
```diff
@@ -29,6 +29,15 @@ if [[ ${image_fs_type} =~ ^ext[[:digit:]] ]] ; then
    image_fs_type="ext2"
 fi
 
+if [ -e /sbin/multipath ]; then
+    migration_rootfs_device=$(lsblk -p -n -r -o NAME,MOUNTPOINTS | grep -E "/$" | uniq | cut -f1 -d" ")
+    for wwn in $(/sbin/multipath -l -v1 2>/dev/null); do
+	if [[ "${migration_rootfs_device}" == *"${wwn}"* ]];then
```
schaefi: nitpick, can you indent the if block by 4 spaces to indicate it belonging to the loop

tools/run_migration
```diff
@@ -72,6 +85,9 @@ function get_boot_options {
     if mdadm --detail "$(get_system_root_device)" &>/dev/null; then
         boot_options="${boot_options} rd.auto"
     fi
+    if [ -n "${migration_rootfs_wwn}" ]; then
+	boot_options="${boot_options} multipath_wwn=${migration_rootfs_wwn}"
```
schaefi: nitpick, can you indent the statement by 4 spaces to indicate it belonging to the block

schaefi (approved):
Looks great 👍 

Just fix the indentation and we are good to merge. Thanks a lot

grub.d/99_migration
```diff
@@ -29,6 +29,15 @@ if [[ ${image_fs_type} =~ ^ext[[:digit:]] ]] ; then
    image_fs_type="ext2"
 fi
 
+if [ -e /sbin/multipath ]; then
```
schaefi: sorry thinking twice about this one can we change this to 

```bash
if type -p multipath &>/dev/null; then
    multipath -l -v1 ...
fi
```

The reason why I think this is better is because checking with `type -p` ensures the tool is in the caller path and after that can be called by just its name. We avoid static path assignments in this case

Thoughts ?

greygoo: Makes sense.

grub.d/99_migration
```diff
@@ -29,6 +29,15 @@ if [[ ${image_fs_type} =~ ^ext[[:digit:]] ]] ; then
    image_fs_type="ext2"
 fi
 
+if [ type -p multipath ]; then
```
schaefi: This is not correct. In bash '[' calls test. You only want 

```bash
if type -p multipath &>/dev/null; then
   ...
fi
```

tools/run_migration
```diff
@@ -49,7 +49,17 @@ function get_system_root_device {
     # """
     # Looking for running system root device
     # """
-    lsblk -p -n -r -o NAME,MOUNTPOINT | grep -E "/$" | uniq | cut -f1 -d" "
+    lsblk -p -n -r -o NAME,MOUNTPOINTS | grep -E "/$" | uniq | cut -f1 -d" "
+}
+
+function get_multipath_wwn {
+    if [ type -p multipath ]; then
```
schaefi: same here

greygoo: yes, noticed it after pushing, fixed it.

PR #495, opened 2026-06-15
Add production/staging profiles for SLE16

schaefi:
In contrast to SLE15, SLE16 is still actively developed and not in maintenance mode. Therefore the rolling build for devel:DMS should use the latest version of SLE16 but should also allow for maintenance releases into production. For production the repository setup should come from the project as it is setup in the SUSE namespace. For the staging project we set the repositories as they fit the current development. This is SLE16.1 at the moment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,7 @@ SLES16-Migration: clean check test
 	cp image/generic/sle16/config.kiwi dist/config.kiwi
 	cp image/generic/sle16/config.sh dist/config.sh
 	cp image/generic/sle16/SLES16-Migration.changes dist/SLES16-Migration.changes
+	cp image/generic/sle16/_multibuild dist/_multibuild
 
 SLES16-SAP_Migration: clean check test
 	mkdir -p dist
@@ -118,6 +119,7 @@ SLES16-SAP_Migration: clean check test
 	cp image/product/sle16/sles_sap/config.kiwi dist/config.kiwi
 	cp image/product/sle16/sles_sap/config.sh dist/config.sh
 	cp image/product/sle16/sles_sap/SLES16-SAP_Migration.changes dist/SLES16-SAP_Migration.changes
+	cp image/product/sle16/sles_sap/_multibuild dist/_multibuild
 
 SLES15-Migration: clean check test
 	mkdir -p dist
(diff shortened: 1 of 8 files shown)


YoukouTenhouin:
Thanks! I was trying to run some similar experiment but I'm glad to see that the changes are already there.

LGTM. Just a question (rather than a review): does this mean that we will need to somehow setup the profile in IBS when submitting the migration images to SLES15? Thanks!

YoukouTenhouin (approved):
LGTM

schaefi:
@YoukouTenhouin you are right one consequence of the change would be that when we submit the image build to the SUSE namespace we have to add a file named `_multibuild` with following content

```
<multibuild>
  <flavor>Production</flavor>
</multibuild>
```

in devel:DMS we will have the same file but with different content

```
<multibuild>
  <flavor>Staging</flavor>
</multibuild>
```

So assuming we merge this you would have to create a `_multibuild` file if you would submitting the live image to the SUSE namespace. Or me when I do it 

Makes sense ?

schaefi:
and as you are mentioning it, I could add the production _multibuild here in the git

schaefi:
@YoukouTenhouin ok done. If you call

```
make SLES16-Migration

or 

make SLES16-SAP_Migration
```

it will produce all submission data for the SUSE namespace including use of the production profile in dist/

YoukouTenhouin:
Thanks for the info, and the update! I'll take care of the submission to SLES15 once this is release then.

Again LGTM

schaefi:
@YoukouTenhouin 

ok I have enabled the build with the profiles for all images in devel:DMS:live and the respective aggregates in devel:DMS.

IMPORTANT:

Please note when you submit migration live images for SLE16 the next time you need to update the following package first:

* suse-migration-rpm

As you can see from the PR a small change in the packaging of live ISO images into an RPM was needed now that we have profiled builds. The reason is the different name of the image files which now includes the profile name and that required a change in the code.

After the above package submission you can submit live image builds as you are used too please use

```
make SLES16-Migration
make SLES16-SAP_Migration
```

To create the sources for the submissions, this automatically also creates you the `_multibuild` file which selects the Production profile.

schaefi:
I'm going to merge this PR now

schaefi:
and I will also bump the version of the DMS due to this and other changes

PR #499, opened 2026-08-18
Add use_cd_repos option for migration from offline media

YoukouTenhouin:
**DRAFT: Still reviewing AI changes, don't merge yet**

Migration from offline media requires us to setup a cd based repository; this currently does not work due to the `--no-cd` option we pass to zypper. This PR adds a configuration flag to allow turning off this option, so migration from offline media can be possible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -114,7 +114,25 @@ Upgrade Modes:
   .. code:: bash
 
       use_zypper_migration: false
-  
+
+Migration From an Installation Medium:
+  Every CD/DVD repository of the system to migrate is ignored by default,
+  which keeps the stale **cd:** repositories left behind by the original
+  installation out of the upgrade. Switching **use_cd_repos** on lets them
+  take part, which is what allows an installation medium such as a SLE Full
+  media to serve as the package source. Such a medium is not registered
+  against a repository server, so **dup** mode is required as well:
+
+  .. code:: bash
+
+      use_zypper_migration: false
+      use_cd_repos: true
+
+  The medium has to be added as a repository on the system to migrate before
+  the migration is started, and any other CD/DVD repository that is still
+  configured should be removed or disabled, as it takes part in the upgrade
+  as well.
+
 Migration Startup:
   There are three ways to perform the migration
 
--- a/doc/adoc/user_guide.adoc
+++ b/doc/adoc/user_guide.adoc
@@ -282,6 +282,26 @@ we use for the zypper migration/dup workflow and for eventual
 package installation after the migration. The default zypper options are
 not changed by this setting.
 
+Migrate From an Installation Medium::
+By default the upgrade process ignores every CD/DVD repository of the
+system to be migrated. Letting them take part is what allows an
+installation medium, such as a SLE Full media, to be used as the package
+source:
++
+[listing]
+----
+use_zypper_migration: false
+use_cd_repos: true
+----
++
+[NOTE]
+Since an installation medium is not registered against a repository
+server, `use_zypper_migration` has to be switched off as well. The medium
+must be added as a repository on the system to be migrated before the
+migration is started. Every other CD/DVD repository that is still
+configured takes part in the upgrade as well, so remove or disable the
+ones left over from the original installation.
+
 Specify Migration Product::
 The default migration target is dependent on the origin version and the
 product. This default target can be changed via the `migration_product`
--- a/suse_migration_services/drop_components.py
+++ b/suse_migration_services/drop_components.py
@@ -91,10 +91,8 @@ class DropComponents:
     def _uninstall_packages(self):
         if self.drop_packages and self.root_path:
             zypper_call = Zypper.run(
-                [
-                    '--no-cd',
-                    '--non-interactive',
-                    '--gpg-auto-import-keys',
+                Zypper.global_args()
+                + [
                     '--root',
                     self.root_path,
                     'remove',
--- a/suse_migration_services/migration_config.py
+++ b/suse_migration_services/migration_config.py
@@ -179,6 +179,17 @@ class MigrationConfig:
     def is_zypper_migration_plugin_requested(self):
         return self.config_data.get('use_zypper_migration', True)
 
+    def are_cd_repos_requested(self):
+        """
+        Return whether CD/DVD repositories may take part in the migration
+
+        By default they are ignored, which keeps the stale cd: repositories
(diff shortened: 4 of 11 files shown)


schaefi:
Hmm, I value very much your effort in adding this feature. Thanks

However, I was asking myself where the request for it comes from and what stakeholders will use this feature ?
Please remember the DMS was exclusively implemented to provide a non-interactive way to upgrade systems. Supporting a process that requires to insert a CD (or pre-mount a CD based repo) in some way shape or form feels very weird to me. In cloud environments such a feature would be never consumable and generally speaking who installs/upgrade an OS from a CD media these days ... I personally haven't done it for years and sometimes I wonder why SUSE still produces ISO media with rpm's on it. I might be mistaken of course but before we merge this we need to understand if that is providing value to customers.

Does this makes sense to you too ?

YoukouTenhouin:
Hi Marcus,

Yes, I very much understand your doubts; that's why I didn't make this a priority and only started working on this right now. This PR is coming from PED-13669 in Jira, which request us to provide a way to migrate from an offline media. Should we take this back to PM?

schaefi:
> Should we take this back to PM?

Yes I think we need to understand if things makes sense from a non technical perspective. I also would value the input from Lukas Lanski and if you can drive the conversation with PM that would be great. Thanks much

I see it this way:

* Fresh installation from media can be done via Offline image and Network image (Agama based), no DMS scope
* Upgrade within the same code stream, can be done with several tools, `zypper migration`, `opensuse-migration-tool`, `manually ..`, no DMS scope
* Upgrade to new major code stream or family from an older code stream or family. That's a DMS scope and I don't get why we cannot make the presence of a network connection a pre-condition to perform this action. The only story I know is from some customer who has a closed network behind a proxy. Access from CD or a CD-mount would be in the same way a pre-setup requirement which is in my opinion not much different from setting up an RMT server providing a local sync of the repos the customer has a subscription for. Even worse a CD setup is static and immutable with the day that ISO was produced but if you upgrade you want the latest upgrades for the major release and not some old state. As such for upgrade processes, and that's the only one we are talking about, I do not understand why a static and old ISO content would be a desired source.

Feel free to take this information as a base for the conversation with PM

libormiksik:
Hello, I've double checked with SLES16.1 PM / architects and got confirmed that there use-cases where exactly this scenario will be needed. So please go ahead with needed changes

schaefi:
> Hello, I've double checked with SLES16.1 PM / architects and got confirmed that there use-cases where exactly this scenario will be needed. So please go ahead with needed changes

Thanks, can you please be more specific what "exactly this scenario" means ? I'm interested in a workflow definition how exactly a customer works with a SLES16 media ISO. I would also be interested why the following procedure is not acceptable/preferred

```
zypper install rmt-server

add SCC credentials to /etc/rmt.conf

systemctl start rmt-server

rmt-cli products enable SLES/16.1/x86_64
rmt-cli mirror

Can be done once in the customer network, then closed down, or via proxy, plenty of options
```

I'm pretty sure a customer wants to stay connected with SUSE, as such there is a subscription, there is surely a network and there is surely interest in updates. Nothing of this is provided by an ISO file.

Thanks much

libormiksik:
as Jiri wrote here -  https://jira.suse.com/browse/PED-13669?focusedId=1541344&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1541344
this is needed for migration for a system relaying only on physical media with no network access.

jsrain:
Typical workflow: You have a 15-SPx box, which is standalone, without connection to any SMT / RMT. You want to update it to 16.1 and bring the 16.1 full media.
Since the box is not connected to any update service, you install - from this media - the SUSE migration services package (just from the media, not from the 16.1 repo) without any need to bring it by any other means - everything you need on a single off-line media.

schaefi:
> Typical workflow: You have a 15-SPx box, which is standalone, without connection to any SMT / RMT. You want to update it to 16.1 and bring the 16.1 full media. Since the box is not connected to any update service, you install - from this media - the SUSE migration services package (just from the media, not from the 16.1 repo) without any need to bring it by any other means - everything you need on a single off-line media.

ok, got that thank you. It also means there will be no subscription from the customer, no connection to SCC and no updates because we are talking about an offline box without any connection to SUSE. Is that correct ? It also means the former system "15-SPx box" is not registered and was installed in the same way (from CD). Is that also correct ?

And we really have customers who do this, really, no updates and they run that in production ?

May I ask the question if this from CD upgrade is so important, why is there no support from Agama from its own media (CD) available to perform this ? Why does this has to become a responsibility of the DMS which was designed for remote and unattended upgrade procedures (initially for clouds)

Something feels wrong 

Of course technically we can do everything here but I'm not really happy with this feature

schaefi:
@YoukouTenhouin why do we need the custom patching ?

* image/pubcloud/product/sles_sap/root/zypper.patch‎

YoukouTenhouin:
@schaefi I was wondering that as well. It seems that this patch was introduced in 6258aac, and then because it is here claude just went ahead and updated it. I kept it here because I was hoping that I might get some feedback whether this will actually be needed. Should we update that one as well or should we leave that out of this PR?

schaefi:
> @schaefi I was wondering that as well. It seems that this patch was introduced in [6258aac](https://github.com/SUSE/suse-migration-services/commit/6258aacae678adda90a38b936a649241ab1431c8), and then because it is here claude just went ahead and updated it. I kept it here because I was hoping that I might get some feedback whether this will actually be needed. Should we update that one as well or should we leave that out of this PR?

:) yeah claude is not always correct. Please drop this patch. Once your implementation is in code no such patching will be needed. Thanks

schaefi:
However, it reminded my that this temporary zypper.patch needs to be dropped. I forgot to get rid of it now that zypper on SLES_SAP was updated. I will drop it

schaefi:
my questions got answered, thanks for the invite to the sync meeting and your comments there @fcrozat 

@YoukouTenhouin if you can fix the conflict which was the consequence of getting rid of the obsolete patch file I'm fine merging and rebuilding devel:DMS such that this feature can also be tested. The code looks good to me.

As stated in the sync meeting a short info to QA (maybe Felix) would be good such that they can think about an air-gaped test case which I believe we should have

YoukouTenhouin:
Rebased. Thanks for reviewing!

schaefi (approved):
Looks good 👍

PR #501, opened 2026-09-11
Allow the migration packages to install on Leap 16

greygoo:
The SLES16-Migration package requires product(SLES), which causes an installation on Leap to fail. This extends the product requirement to Leap products.

### DIFF
--- a/image/package/suse-migration-rpm.changes
+++ b/image/package/suse-migration-rpm.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Fri Aug 14 17:34:04 UTC 2026 - Christian Hueller <chuller@suse.com>
+
+- Allow the migration packages to install on Leap 16
+
+  The SLES16-Migration package requires product(SLES), which causes
+  an installation on Leap to fail. This extends the product
+  requirement to Leap products.
+
+-------------------------------------------------------------------
 Tue Jun 16 16:19:40 UTC 2026 - Marcus Schäfer <marcus.schaefer@gmail.com>
 
 - Add handling for Staging builds
--- a/image/package/suse-migration-rpm/kiwi_post_run
+++ b/image/package/suse-migration-rpm/kiwi_post_run
@@ -31,7 +31,14 @@ ARCH="$(uname -m)"
 NAME=$(echo "$IMAGE" | cut -f1 -d.)
 VERSION=$(echo "$IMAGE" | cut -f3 -d-)
 RELEASE=$(echo "$IMAGE" | sed -e "s@-Staging@@" | sed -e "s@-Production@@" | cut -f4 -d- | sed -e "s@Build@@" | sed -e 's@.iso$@@')
-PRODUCT='product(SLES) >= %{MinSLEVersion}'
+case "$NAME" in
+    SLES16*)
+        PRODUCT='(product(SLES) >= %{MinSLEVersion} or product(Leap) >= 16.0)'
+        ;;
+    *)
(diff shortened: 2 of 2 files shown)


rjschwei:
Hmmm, I do not think we want to support this. The basic assumption from the get-go of the migration system was a registered SLE system, i.e. `SUSEConnect` is available and has been used at some point. While there is a chance that the brute force approch to migration might work in openSUSE, i.e. `zypper dup` should work the migration target by default is determined by the information that `SUSEConnect` produces and what is set as the target in the config file.

Just installing the package is not helpful, and opens us up for lots of bug reports that we probably really do not want to address.

IMHO this should be closed and not considered for inclusion.

schaefi:
Hi @rjschwei, like me I believe you have missed the decision from PM which says that the DMS (alongside with CD/ISO source media) has to support migrations from:

* Leap -> SLES

I got informed about this via the DMS meeting which I also didn't know that it exists :) Now being invited to this meeting was very helpful to get some background information I was not aware of.

So with the above in mind it's clear that installing SLES16-Migration should be also possible on Leap and the initial idea to strictly require **one** product family can no longer be kept.

That's the reason why in a first step @greygoo added a change to allow this

There will be a second change required which affects the code in the DMS that selects the target product triplet when no configuration exists.

Since Leap is basically the same as SLES, I'm open to the packaging change to allow installation on both product families.

Thoughts ?

schaefi (approved):
Changes looks good to me from a technical perspective.

If there are no objections regarding the behavior change and the design going forward I'm fine

rjschwei:
> Hi @rjschwei, like me I believe you have missed the decision from PM which says that the DMS (alongside with CD/ISO source media) has to support migrations from:
> 
> * Leap -> SLES
> 
> I got informed about this via the DMS meeting which I also didn't know that it exists :) Now being invited to this meeting was very helpful to get some background information I was not aware of.
> 
> So with the above in mind it's clear that installing SLES16-Migration should be also possible on Leap and the initial idea to strictly require **one** product family can no longer be kept.
> 
> That's the reason why in a first step @greygoo added a change to allow this
> 
> There will be a second change required which affects the code in the DMS that selects the target product triplet when no configuration exists.
> 
> Since Leap is basically the same as SLES, I'm open to the packaging change to allow installation on both product families.
> 
> Thoughts ?

Well, I do not think the current code works on Leap and definitely not for a Leap to SLE migration. As such I think this step, i.e. install on Leap, should be the last step in supporting the feature request, not the first.

schaefi:
> Well, I do not think the current code works on Leap and definitely not for a Leap to SLE migration. As such I think this step, i.e. install on Leap, should be the last step in supporting the feature request, not the first

I agree and this is also the reason why we collect the changes we know that has to happen. This includes this one as well as the product detection change Christian Hüller is about to prepare. Once we believe the needed changes has been formulated as PRs I'm going to create a new build of the DMS in devel:DMS such that testing becomes possible. From that stage I also think further changes for a successful Leap to SLES migration will be needed. I also think the documentation must include a chapter how to prepare a Leap system such that it can turn into SLES.

greygoo:
I can confirm, that migration from a Leap16.1 system registered with a SLE key to SLE16.1 works with DMS. The only issues right now are the package installation (this PR) and the product detection, which will be another PR.
Without the product selection fix it is necessary to add migration_product: SLES/16.1/x86_64 to the dms config, but with that it does migrates fine in my tests.

schaefi:
@greygoo thanks much for the feedback. I'll wait for the product triplet fix from Christian and then merge both changes.

PR #502, opened 2026-09-14
Fixed release package for SLES15-SAP_Migration

schaefi:
The product specific build installed sles-release instead of SLES_SAP-release


schaefi:
Build with this change running here

* https://build.opensuse.org/package/show/devel:DMS:live/SLES12-SP4-SLES15-SAP-Migration

YoukouTenhouin (approved):
Not familiar with 12 to 15 Migration, but I compared with the SLES16  SLES_SAP migration image and also an installation of SLES15SAP, and this looks good to me.

greygoo:
Since with the current code the product still is retrived from the hsot and not the migration system, this did not show any difference when migrating SLE12->SLES15. LGTM

rjschwei (approved):

schaefi:
> Since with the current code the product still is retrived from the hsot and not the migration system, this did not show any difference when migrating SLE12->SLES15. LGTM

This has no relation to the product triplet setup in the code. This is about the live migration system for SAP to identify itself as SLES_SAP and not SLES which impacts the package requirements from an installation perspective. In addition a SLES_SAP product will always migrate in the same product family which is SLES_SAP and nothing else. For that condition the code that creates the product triplet is working as designed.

PR #503, opened 2026-09-15
Read the product to migrate to from the migration image

greygoo:
get_product_name() built the target product from the base product of the system to migrate and the version of the migration image. The name therefore described the source, not the target: migrating a Leap 16.0 host with the SLES 16.1 image auto-detected Leap/16.1/x86_64, a product that exists in neither, and zypper migration aborted with "No migration available".

Read the name from the base product of the migration image instead, the same place the version already comes from. The image ships the release package of the product it migrates to, SLES-release for SLES16-Migration and SLES_SAP-release for SLES16-SAP_Migration. The architecture stays with the system to migrate, migrations do not change it.

MigrationImageProduct overrides the root SUSEBaseProduct reads the product metadata from, so the existing callers keep reading the system to migrate.


suse_migration_services/suse_product.py
```diff
@@ -59,6 +58,11 @@ def __init__(self, log):
 
         self.base_product = base_product_files[0]
 
+    @staticmethod
+    def get_root_path():
+        """Root of the system the product metadata is read from."""
+        return Defaults.get_system_root_path()
+
```
schaefi: This is very confusing to me and imho unneeded. The PR introduced the method get_root_path and the method implementation itself is calling a method from the defaults space. Can we please avoid that and just call the method from the defaults space as needed.

suse_migration_services/suse_product.py
```diff
@@ -28,8 +28,7 @@
 class SUSEBaseProduct:
     def __init__(self, log):
         self.log = log
-        root_path = Defaults.get_system_root_path()
-        self.products_metadata = os.sep.join([root_path, 'etc', 'products.d'])
+        self.products_metadata = os.sep.join([self.get_root_path(), 'etc', 'products.d'])
```
schaefi: Can this please use

```
Defaults.get_system_root_path()
```

schaefi: In addition with this change **all** product information for setting the target product is now taken from the host to migrate. That is not what we want.

The product name to make up the target is created in the method `get_product_name` a SUSE product triplet consists out of three parts

* family name (self.get_tag('name')[0]) -- Leap / SLES/ ...
* target version taken from etc/os-release --- VERSION_ID="16.0" ...
* architecture (self.get_tag('arch')[0]) -- x86_64 / aarch64 /...

So far all of this information was taken from the migration live image and we only agreed on that the family name now is allowed to be taken from the host to migrate instead of the migration live system.

As such I would have expected something like

* refactor SUSEBaseProduct constructor and move the code into its own. Something like this

```patch

diff --git a/suse_migration_services/suse_product.py b/suse_migration_services/suse_product.py
index e4b665f..55d736f 100644
--- a/suse_migration_services/suse_product.py
+++ b/suse_migration_services/suse_product.py
@@ -28,12 +28,16 @@ from suse_migration_services.exceptions import DistMigrationSUSEBaseProductExcep
 class SUSEBaseProduct:
     def __init__(self, log):
         self.log = log
-        root_path = Defaults.get_system_root_path()
-        self.products_metadata = os.sep.join([root_path, 'etc', 'products.d'])
-        self.prod_filenames = glob.glob(os.path.join(self.products_metadata, '*.prod'))
+        self.base_product = self.read_product(
+            log, Defaults.get_system_root_path()
+        )
+
+    def read_product(self, log, root_path):
+        products_metadata = os.sep.join([root_path, 'etc', 'products.d'])
+        prod_filenames = glob.glob(os.path.join(products_metadata, '*.prod'))
         base_product_files = []
         xml = ElementTree()
-        for prod_filename in self.prod_filenames:
+        for prod_filename in prod_filenames:
             try:
                 xml.parse(prod_filename)
                 register_sections = xml.findall('register')
@@ -57,7 +61,7 @@ class SUSEBaseProduct:
             log.error(message)
             raise DistMigrationSUSEBaseProductException(message)
 
-        self.base_product = base_product_files[0]
+        return base_product_files[0]
 
     def delete_target_registration(self):
         self.backup_products_metadata()
@@ -81,12 +85,15 @@ class SUSEBaseProduct:
         previous state. The rollback restores this info back in place.
         """
         self.log.info('Creating backup of Product data')
+        products_metadata = os.sep.join(
+            [Defaults.get_system_root_path(), 'etc', 'products.d']
+        )
         Command.run(
             [
                 'rsync',
                 '-zav',
                 '--delete',
-                self.products_metadata + os.sep,
+                products_metadata + os.sep,
                 '/tmp/products.d.backup/',
             ]
        )
```

* Next a change in the `get_product_name` name method as we discussed, take the family name from the system to migrate and don't change anything else. Something like this

```patch
-    def get_tag(self, tag):
+    def get_tag(self, tag, base_product):
         """Get the content of tag if any."""
         xml = ElementTree()
         try:
-            xml.parse(self.base_product)
+            xml.parse(base_product or self.base_product)
             tag_nodes = xml.findall(tag)
             return [tag_node.text for tag_node in tag_nodes]
         except Exception as issue:
             # if we are here, it means no potentially no migration
             # product is defined
             self.log.warning(
-                'Parsing XML file {0} failed with: {1}'.format(self.base_product, issue)
+                'Parsing XML file {0} failed with: {1}'.format(
+                    base_product or self.base_product, issue
+                )
             )
 
     def get_product_name(self):
         """Get the product name to be migrated to."""
         migration_product_name = None
         try:
-            name = self.get_tag('name')[0]
+            self.base_product_system_to_migrate = read_product(
+                self.log, Defaults.get_migration_image_root_path()
+            )
+            name = self.get_tag('name', self.base_product_system_to_migrate)[0]
             arch = self.get_tag('arch')[0]
             if name and arch:
                 migration_product_name = '/'.join([name, self.get_default_target_version(), arch])
```

schaefi (changes requested):
Hi, thanks for looking into this.

The change here moves the base product reading from the migration live system to the host to migrate. This is not what we wanted and imho comes with a number of unwanted side effects. What we agreed on is to take the family name from the product triplet from the host to migrate but nothing else

See my patch suggestions and I hope this makes sense

suse_migration_services/suse_product.py
greygoo: I'm a bit confused about "So far all of this information was taken from the migration live image" - it was taken from the host I think. That's the issue - when it picks Leap and then SCC tells it "no migration path for that".

schaefi (approved):
Looks good 👍

schaefi:
@greygoo can you rebase the changes into one commit, then I think we are ready to merge :) Thanks

PR #505, opened 2026-09-16
Do not use profiles Staging/Production

schaefi:
The maintenance team does not accept submissions with profiles


schaefi:
Changes have been applied for testing in devel:DMS

YoukouTenhouin (approved):
LGTM

PR #506, opened 2026-09-17
Fix sles version check for SLE16

schaefi:
The current check does not match for SLE16 as it requires is_opensuse to be true which is not the case for SLE16+


schaefi:
heads up: I did this change as it seems natural to me, I haven't done a test yet. As such your review is much appreciated. Thanks

